PR #2, opened 2018-06-29
Story/1087

ws-mps:


addons/udes_report/tests/test_stock_export.py
```diff
@@ -43,9 +55,9 @@ def test_stock_no_locations(self):
     def test_stock_all_excluded_locations(self):
         ''' Stock - should error in case all locations are excluded '''
         self.stock_export.included_locations = \
-            self.LocationObj.browse(self.stock_location)
+            self.LocationObj.browse(self.stock_location.ids)
```
ws-mps: Browse is not needed, lines 57 and 58 are like:
self.stock_export.included_locations = self.stock_location

This might apply to excluded_locations and other tests.

PR #8, opened 2018-07-20
Refactor unittests warehouse setup

nocturnalastro:


addons/udes_stock/tests/common.py
```diff
@@ -14,10 +15,74 @@ class BaseUDES(common.SavepointCase):
...
+    @classmethod
+    def setup_default_warehouse(cls):
+        ''' Sets a warehouse with:
+            * test locations
+                test_input_01, test_stock_01, test_stock_02, test_output_01
+            * all picking types as a copy of the default with TEST_ prepended
+            * a simple inbound route goods_in->putaway
+            * a simple outbound route pick->goods_out
+        '''
+        cls._setup_pick_types()
```
ws-mps: Better use cls._setup_picking_types() since pick is a specific picking type.

addons/udes_stock/tests/common.py
```diff
@@ -14,10 +15,74 @@ class BaseUDES(common.SavepointCase):
...
+        setattr(user_warehouse, '{}_type_id'.format(ptype), new_pt)
+
+    @classmethod
+    def _setup_locations(cls):
+
+        Location = cls.env['stock.location']
         cls.stock_location = cls.env.ref('stock.stock_location_stock')
+
+        input_zone = cls.stock_location.copy({'name': 'TEST_INPUT'})
+        cls.recived_location = Location.create({
```
ws-mps: typo: received_location

Same for name and barcode.

addons/udes_stock/tests/common.py
```diff
@@ -30,26 +95,73 @@ def setUpClass(cls):
...
+    def _setup_routes(cls):
+
+        cls.picking_type_in.write({
+            'default_location_src_id': cls.env.ref('stock.stock_location_suppliers').id,
+            'default_location_dest_id': cls.recived_location.id,
+        })
+
+        cls.picking_type_internal.write({
+            'default_location_src_id': cls.recived_location.id,
+            'default_location_dest_id': cls.test_location_01.id,
```
ws-mps: why test location and not stock?

addons/udes_stock/tests/common.py
```diff
@@ -30,26 +95,73 @@ def setUpClass(cls):
...
+    @classmethod
+    def _setup_routes(cls):
+
+        cls.picking_type_in.write({
+            'default_location_src_id': cls.env.ref('stock.stock_location_suppliers').id,
+            'default_location_dest_id': cls.recived_location.id,
+        })
+
+        cls.picking_type_internal.write({
+            'default_location_src_id': cls.recived_location.id,
```
ws-mps: Not big fan of this change, internal by default should be from stock to stock.
Would be better to create putaway picking type for the inbound route.

addons/udes_stock/tests/common.py
```diff
@@ -126,12 +238,12 @@ def create_picking(cls, picking_type, products_info=False,
                 product_info.update(picking=picking)
                 move = cls.create_move(**product_info)
 
-        if confirm:
-            picking.action_confirm()
-
         if assign:
             picking.action_assign()
 
+        if confirm:
```
ws-mps: Why?  A picking is first confirmed, then assigned.

addons/udes_stock/tests/common.py
```diff
@@ -246,10 +358,12 @@ def create_simple_inbound_route(cls, picking_type_in, picking_type_internal):
         }
         route = Route.create(route_vals)
 
+        #TODO make pretty
+
         # PUTAWAY
-        sequence_putaway = Sequence.create({"name": "TestPutaway", "prefix": "TESTPUT", "padding": 5}).id
+        sequence_putaway = Sequence.create({"name": "TestPutaway", "prefix": "TESTPUT", "padding": 5})
```
ws-mps: If we create the Putaway picking type this is not going to be needed :)

PR #9, opened 2018-07-23
[IMP] Preventing open redirect with get param.

crwr45:
Overriding functions within Odoo to prevent open redirects being executed.

### DIFF
--- a/addons/udes_security/LICENSE
+++ b/addons/udes_security/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
+any other work released this way by its authors.  You can apply it to
+your programs, too.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  To protect your rights, we need to prevent others from denying you
+these rights or asking you to surrender the rights.  Therefore, you have
+certain responsibilities if you distribute copies of the software, or if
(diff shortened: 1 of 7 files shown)


addons/udes_security/common.py
```diff
@@ -0,0 +1,14 @@
+from odoo.http import request
+import re
+
+
+# Replaces redirect param with None if it's going external
+def remove_external_redirect(redirect):
+    base_location = request.httprequest.url_root.rstrip('/')
```
crwr45: Python convention is that the docstring for a method/class should be in triple double-quotes, immediately below the def:
```
def remove_external_redirect(redirect):
    """Replaces redirect param with None if it's going external"""
```

addons/udes_security/controllers/authSignupHome.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
crwr45: Filenames should be all lowercase, with underscore separators e.g. `auth_signup_home.py`
Again this is a python style guidelines: https://www.python.org/dev/peps/pep-0008/#package-and-module-names

addons/udes_security/controllers/authSignupHome.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import http, exceptions
+import odoo.addons.web.controllers.main as main
+import odoo.addons.auth_signup.controllers.main as authmain
+from odoo.addons.web.controllers.main import ensure_db
+from ..common import remove_external_redirect
+
+
+class AuthSignupHome(authmain.AuthSignupHome):
+
+    # Override web_login function
```
crwr45: Again, change to a docstring. I would also like to see a one-sentence explanation of what this doing to the web_login function. e.g. Override web_login function to remove open redirect.

addons/udes_security/controllers/authSignupHome.py
```diff
@@ -0,0 +1,32 @@
...
+        response.qcontext.update(self.get_auth_signup_config())
+
+        # Validate redirect used here
+        redirect = request.params.get('redirect')
+        redirect = remove_external_redirect(redirect)
+
+        if request.httprequest.method == 'GET' and request.session.uid and redirect:
+            # Redirect if already logged in and redirect param is present
+            return http.redirect_with_hash(redirect)
+        return response
```
crwr45: I would be much happier with this if it was calling super() and then modifying the result. What is the reasoning behind not using a super() call?

addons/udes_security/controllers/mainHome.py
```diff
@@ -0,0 +1,39 @@
...
+        res = super(Home, self)._login_redirect(uid, redirect)
+        return res
+
+    # Override web_client function
+    @http.route('/web', type='http', auth="none")
+    def web_client(self, s_action=None, **kw):
+        ensure_db()
+        if not request.session.uid:
+            return werkzeug.utils.redirect('/web/login', 303)
+        # Begin UDES security modifications
```
crwr45: Again, I'd be much happier if this was using super.

addons/udes_security/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
crwr45: No reason to add a license as part of this work, and we've used the LGPL license for our other things

PR #13, opened 2018-08-01
[ADD] udes_stock: Add load test for picking

nocturnalastro:
User-story: 1335


addons/udes_stock/tests/test_load_testing.py
```diff
@@ -15,6 +15,17 @@
 except ImportError:
     _HAS_PARAMETERIZED = False
 
+    # Gets around dependacies not being there during eval
```
mcb30: "dependencies"

addons/udes_stock/tests/test_load_testing.py
```diff
@@ -215,11 +223,13 @@ def time_validate_lines(self, pick, packages):
     def time_validate_pick(self, pick):
         pick.update_picking(validate=True)
 
-    # Printed things dont come out when using parameterized
-    # hense test_report
+    # Printed things dont come out when using parameterized hense test_report
```
mcb30: "hence"

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+from collections import defaultdict
+from matplotlib import pyplot as plt
+import sys
+
+COEFFICIENT_THRESHOLD = 0.7
+PVAL_THRESHOLD = 0.001
+
+# Do we want to bootstrap this?
+def scaling_coefficient(x, y):
+    """Nuno's neat method"""
```
mcb30: Given that this is public code, this should probably have a more globally meaningful docstring

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+from scipy.stats import spearmanr
+from collections import defaultdict
+from matplotlib import pyplot as plt
+import sys
+
+COEFFICIENT_THRESHOLD = 0.7
+PVAL_THRESHOLD = 0.001
+
+# Do we want to bootstrap this?
```
mcb30: bootstrap?

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+    nums, times = zip(*sorted(data[func_name].items()))
+    times = np.array(times)
+    means = times.mean(axis=1)
+    minmax = (means - times.min(axis=1), times.max(axis=1) - means)
+    nums = np.array(nums)
+    plt.errorbar(nums, means, label=func_name, yerr=minmax)
+
+    area, coefficient, pval = scaling_coefficient(nums, means)
+
+    definitely_worse_template = '{name} is definatly worse than linear:\n'
```
mcb30: "definitely"

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+            area=area,
+            name=func_name))
+    elif coefficient < -COEFFICIENT_THRESHOLD:
+        print(possibly_worse_template.format(
+            coefficient=coefficient,
+            pval=pval,
+            area=area,
+            name=func_name))
+    else:
+        print('no corrilation found for {}'.format(func_name))
```
mcb30: "correlation"

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+# Do we want to bootstrap this?
+def scaling_coefficient(x, y):
+    """Nuno's neat method"""
+    dydx = np.diff(y)/np.diff(x)
+    coefficient, pval = spearmanr(x[1:], dydx)
+    area = np.sum(dydx - dydx[0])
+    return area, coefficient, pval
+
+
+def parse_time_file(filename):
```
mcb30: What file format are we reading here, and is there some pre-existing parser that would already handle it?

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+    lines = open(filename, 'r').read()
+    data = defaultdict(lambda: defaultdict(list))
+
+    for line in filter(lambda l: len(l) > 0 and '=' not in l,
+                       lines.split('\n')):
```
mcb30: .splitlines() is probably better than .split('\n'), since it will gracefully handle possible variation in line endings between platforms

addons/udes_stock/tests/plot_load_tests.py
```diff
@@ -0,0 +1,86 @@
...
+    coefficient, pval = spearmanr(x[1:], dydx)
+    area = np.sum(dydx - dydx[0])
+    return area, coefficient, pval
+
+
+def parse_time_file(filename):
+    lines = open(filename, 'r').read()
+    data = defaultdict(lambda: defaultdict(list))
+
+    for line in filter(lambda l: len(l) > 0 and '=' not in l,
```
mcb30: Given that a non-empty string will be True anyway, can this be just "lambda l: l and '=' not in l"?

addons/udes_stock/tests/plot_load_tests.py
nocturnalastro: Randomly selecting values from the list of times instead of taking the means then repeating the process.  This could be used to account for there being a range of values but it would obviously be slower. Probably not worth it.

addons/udes_stock/tests/plot_load_tests.py
nocturnalastro: It was just a text file with
   name of the function, the dependant variably and the time in a tab separated line 
   with a line of ='s to separate run for someone looking by eye 
Could probably just write it as a csv and use the csv reader instead. 
I'll change both the read and write of the file.

PR #16, opened 2018-08-15
Story/1385 Load test module new

nocturnalastro:


baluke:
You have ```[(10,), (20,), (30,), (40,), (50,)] * 3``` 8 times in the code. It may be worth putting a constant in common.py .

It's also worth having the biggest value tested be over 50. Empirically, non-linear behaviour was clear at about 200-500 (and 1200+ simply take unnecessarily long to run).

addons/udes_load_testing/tests/test_outbound.py
```diff
@@ -0,0 +1,230 @@
...
+class OutboundLines(LoadRunner):
+
+
+    @classmethod
+    def _setup_3step_outbound_route(cls, pt_pick, pt_check):
+        Route = cls.env['stock.location.route']
+        Path = cls.env['stock.location.path']
+        Sequence = cls.env['ir.sequence']
+
+        # Chekc
```
baluke: Typo.

addons/udes_load_testing/tests/common.py
```diff
@@ -134,6 +100,18 @@ def test_report(self):
             for line in lines:
                 print(line)
 
+
+
+class BackgroundDataRunner(LoadRunner):
+    _N = int(1e5)
+    _dummy_picking_type = None
```
baluke: This is an unusual style to use in Python. Just initialize them in the setUpClass method you have already.

baluke:
Overall, looks good. I've flagged up a couple of coding style nits. Any chance of a README that demonstrates how to run all the tests, as well as a subset of the tests?

addons/udes_load_testing/tests/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Provides load tests for UDES
+
+## Running tests
+
+Load tests can be run using the normal odoo unit tests framework:
+To run all tests
+
+``` bash
+odoo-bin -i udes_stock,udes_load_testing --test-enable
```
baluke: Doesn't run via odoo-bin, only via udes, according to some manual testing I just did.

baluke: This can be left as-is, as per discussion.

addons/udes_load_testing/tests/README.md
```diff
@@ -0,0 +1,55 @@
...
+``` bash
+odoo-bin -i udes_stock,udes_load_testing -f udes_load_testing/test/test_picking.py
+```
+
+### Config
+
+To change the steps or repeats you can change add a udes_load_test section to your rc file.
+You can specify default values for the parameterized tests and the default number of repeats.
+It is also possible to define the parameterize values to be used for individual classes.
+This is done by specifing the class's name (case insenitive).
```
baluke: Please spell-check this file.

addons/udes_load_testing/tests/common.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+
+class BackgroundDataRunner(LoadRunner):
+
+    @classmethod
+    def setUpClass(cls):
+        super(BackgroundDataRunner, cls).setUpClass()
+        cls._N = config.get_background_N(cls.__name__)
+        cls._dummy_picking_type = cls.picking_type_pick
```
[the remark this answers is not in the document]
nocturnalastro: We might no always want picking_type_pick

PR #31, opened 2018-09-26
[1524] New logic for suggesting / checking drop locations

parisiale:
Adding:
 - new configuration
 - new constraint for ensuring move lines' `location_dest_id` is a child of the picking dest location
 - new abstraction for suggesting locations based on different policies
 - new `stock-picking` endpoint for requesting suggested locations (also, removing the old `stock-quant-package` one)

TODO:

- [x] address PRs in code comments (TODOs, commented-out stuff)
- [x] frontend interface changes
- [x] implement policies for suggesting drop locations
- [x] ensure docs are updated
- [x] remove old `stock-quant-package` endpoint for suggesting drop locations


addons/udes_stock/README.md
```diff
@@ -95,62 +95,79 @@ The type of stock.picking can is defined by this type. It can represent a goods
...
-| u_skip_allowed            | boolean | Is the user allowed to skip to the next item to pick? |
-| u_split_on_drop_off       | boolean | |
-| u_suggest_qty             | boolean | Do we display the suggested quantity, or get the user to enter it without any vision of what is expected. When we suggest it, the risk is that users will be automatically confirming it without a thorough check. |
-| u_over_receive            | boolean | Is the system able to receive more than is expected. |
-| u_target_storage_format   | string  | This defines how the stock is stored at the end of the stock.picking. |
-| u_user_scans              | string  | This defines what the user will scan. |
-| u_validate_real_time      | boolean | Do we validate move lines in real time |
+
+
+// TODO: REMOVE VVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVVV
```
ws-mps: Remove them :)

parisiale: When I don't need that anymore.
BTW, don't worry, as there a TODO item in the main thread to avoid forgetting that.

addons/udes_stock/controllers/stock_picking.py
```diff
@@ -58,3 +58,32 @@ def is_compatible_package(self, ident, package_name):
...
+        Picking = request.env['stock.picking']
+
+        picking = Picking.browse(int(identifier))
+
+        kwargs = {}
+        if package_name:
+            package = Package.get_package(package_name)
+            kwargs['package'] = package
+        elif move_line_ids:
+            kwargs['move_line_ids'] = MoveLine.browse(move_line_ids)
```
parisiale: I would raise a `ValidationError` in an `else` right above.

parisiale: No need for that ^^.
After chatting, we agreed that we don't need to validate the request here as we want to keep the format loose, being policy-specific.

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -0,0 +1,187 @@
...
+                            confirm=True,
+                            assign=True)
+
+        mls = picking.move_line_ids
+        mls.write({'location_dest_id': self.test_location_01.id})
+        location = picking.get_suggested_locations(move_line_ids=mls)
+        self.assertEqual(location, self.test_location_01)
+
+    def test02_get_suggested_locations_by_product(self):
+        """ Suggested locations should be all where the product is at.
```
crwr45: This should be rephrased to something like: "All of the suggested locations should have current stock of the product"

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -0,0 +1,187 @@
...
+                            products_info=self.pack_4apples_info,
+                            confirm=True,
+                            assign=True)
+
+        locations = picking.get_suggested_locations(products=self.apple)
+        self.assertEqual(locations, self.test_locations)
+
+    def test03_get_suggested_locations_by_product_using_move_lines(self):
+        """ Suggested locations should be all where the product is at, this test
+            uses move_lin_ids as paramater instead of products.
```
crwr45: move_line_ids

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -0,0 +1,187 @@
...
+        self.assertEqual(locations, self.test_locations)
+
+    def test04_get_suggested_locations_by_package(self):
+        """ Suggested locations should be all where the products inside the
+            package are at.
+        """
+        self.picking_type_putaway.u_drop_location_policy = 'by_packages'
+        self.create_quant(self.apple.id, self.test_location_01.id, 4,
+                          package_id=self.package_two.id)
+        self.create_quant(self.banana.id, self.test_location_02.id, 4)
```
crwr45: Wouldn't this be a better test if this quant was in a package as well?

ws-mps: I will add a note as discussed:
Note that for packages with multiple products it is suggested any location where any of the products are at.

PR #34, opened 2018-09-27
Story/1517

ws-mps:


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -0,0 +1,41 @@
...
+    _inherit = "sale.order"
+
+    # Rename states
+    state = fields.Selection([
+        ('draft', 'Quotation'),
+        ('sent', 'Quotation Sent'),
+        ('sale', 'In Progress'),
+        ('done', 'Done'),
+        ('cancel', 'Cancelled'),
+    ])
```
mcb30: Could this use selection_add instead of redefining the existing selection values?

ws-mps: The idea is to keep the same values, but just renaming their names, I'll check if we can use selection_add to do it.

ws-mps: Yes, looks like selection_add will replace the ones we change the name. I will update it.
Thanks.

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -0,0 +1,38 @@
...
+    @api.depends('order_line.move_ids.picking_id')
+    def _compute_picking_ids_by_line(self):
+        for order in self:
+            order.picking_ids = order.mapped(
+                'order_line.move_ids.picking_id')
+
+    def check_delivered(self):
+        """ Update sale orders state based on the states of their related
+            pickings.
+            An order is considered done or cancel when all its pickings are done
```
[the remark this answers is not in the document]
ws-mps: Updated thanks.

PR #35, opened 2018-10-01
Story/1605

unirho:


addons/udes_stock/tests/test_location_pi.py
```diff
@@ -301,8 +301,8 @@ def test11_process_pi_count_moves_valid_request_with_package(self):
                         "location_id": self.test_location_01,
                         "location_dest_id": self.test_location_02.id}]
 
+        print(self.test_location_02)
```
mcb30: This looks like debug code that should be removed

PR #36, opened 2018-10-04
Add product category fields to products and locations

ws-mps:


addons/udes_stock/models/product_template.py
```diff
@@ -2,9 +2,27 @@
...
+
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    def _domain_product_category(self, category):
+        """Domain for product categories, not including category itself"""
+        Category = self.env['product.category']
+        subcategories = Category.search([('id', 'child_of', category.id)])
+        return [('id', 'in', subcategories.ids),
+                ('id', '!=', category.id)]
```
mcb30: Can we not just return a domain which uses child_of directly, rather than building an explicit list of ids?

ws-mps: yes definetly.

parisiale:
👍

PR #37, opened 2018-10-05
[IMP] (1855) Update picking::suggested_locations endpoint

parisiale:
The request schema will now require only the `move_line_ids` array entry (mandatory).


ws-mps:
Looks good 👍

PR #38, opened 2018-10-07
[IMP] udes_stock: Extend stock.move.line location_dest_id constraint

ws-mps:
The constraint is also now triggered by onchange and checks if the
location is one of the suggested locations when the drop off constraint
is set to enforce at the picking type.

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -661,6 +661,8 @@ class StockMoveLine(models.Model):
 
         # HERE(ale): iterating picking_id even if it's a many2one
         # because the constraint can be triggered anywhere
+
+        # TODO(ale): consider using `picking::groupby` once available
         for picking in self.mapped('picking_id'):
             if not picking.is_valid_location_dest_id(location=location):
                 raise ValidationError(
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -672,12 +672,12 @@ class StockMoveLine(models.Model):
             if picking.picking_type_id.u_drop_location_constraint == 'enforce':
                 mls = self.filtered(lambda ml: ml.picking_id == picking)
                 locations = picking.get_suggested_locations(mls)
-                # location should be one of the suggested locations if there
-                # are any
+
+                # location should be one of the suggested locations, if any
                 if locations and location not in locations:
                     raise ValidationError(
                         _("Drop off location must be one of the suggested "
-                          "locations."))
+                          "locations"))
 
     def any_destination_locations_default(self):
         """Checks if all location_dest_id's are default_location_dest_id of
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -652,17 +652,14 @@ class StockMoveLine(models.Model):
         """
         location = self.mapped('location_dest_id')
 
-
         # On create we need to allow views to pass.
         # On other actions if this will be caught by odoo's code as you are not
         # allowed to place stock in a view.
         if set(location.mapped('usage')) == set(('view',)):
             return True
 
-        # HERE(ale): iterating picking_id even if it's a many2one
+        # iterating picking_id even if it's a many2one
         # because the constraint can be triggered anywhere
-
-        # TODO(ale): consider using `picking::groupby` once available
         for picking in self.mapped('picking_id'):
             if not picking.is_valid_location_dest_id(location=location):
                 raise ValidationError(
@@ -672,7 +669,16 @@ class StockMoveLine(models.Model):
                 )
 
             if picking.picking_type_id.u_drop_location_constraint == 'enforce':
-                mls = self.filtered(lambda ml: ml.picking_id == picking)
+
+                # Don't use move lines that are not available for suggesting
+                mls = self.filtered(
+                    lambda ml: ml.picking_id == picking and \
+                               ml.state in ('assigned', 'partially_available')
+                )
+
+                if not mls:
+                    continue
+
                 locations = picking.get_suggested_locations(mls)
 
                 # location should be one of the suggested locations, if any
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -643,9 +643,12 @@ class StockMoveLine(models.Model):
     ## Drop Location Constraint
     #
     @api.constrains('location_dest_id')
+    @api.onchange('location_dest_id')
     def _validate_location_dest(self):
         """ Ensures that the location destination is a child of the
-            default_location_dest_id of the picking.
+            default_location_dest_id of the picking and that is
+            one of the suggested locations if the drop off policy
+            is enforce.
         """
         location = self.mapped('location_dest_id')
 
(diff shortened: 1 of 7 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -654,5 +657,15 @@ def _validate_location_dest(self):
...
+                      (location.name, picking.location_dest_id.name)))
+            if picking.picking_type_id.u_drop_location_constraint == 'enforce':
+                mls = self.filtered(lambda ml: ml.picking_id == picking)
+                locations = picking.get_suggested_locations(mls)
+                # location should be one of the suggested locations if there
+                # are any
+                if locations and location not in locations:
+                    raise ValidationError(
+                        _("Drop off location must be one of the suggested "
+                          "locations."))
```
parisiale: Nitpick: not using full stop elsewhere for error messages. Can be addressed later.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -654,5 +657,15 @@ def _validate_location_dest(self):
...
             if not picking.is_valid_location_dest_id(location=location):
                 raise ValidationError(
-                    _("The location '%s' is not a child of the picking destination "
-                      "location '%s'" % (location.name, picking.location_dest_id.name)))
+                    _("The location '%s' is not a child of the picking "
+                      "destination location '%s'" %
+                      (location.name, picking.location_dest_id.name)))
+            if picking.picking_type_id.u_drop_location_constraint == 'enforce':
+                mls = self.filtered(lambda ml: ml.picking_id == picking)
+                locations = picking.get_suggested_locations(mls)
```
parisiale: Due to the two decorators, `get_suggested_locations` can be invoked twice: form field selection and move line update. Three times if we decide to present the suggestions. I still assume that we don't want to store the suggested locations for the sake of such workflow.

parisiale: This will be addressed within a separate story.

parisiale:
👍

addons/udes_stock/tests/test_move_line.py
```diff
@@ -0,0 +1,108 @@
...
+class TestValidateLocationDest(common.BaseUDES):
+    ''' Tests the drop location constraint '''
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestValidateLocationDest, cls).setUpClass()
+        cls.picking_type_in.u_target_storage_format = 'product'
+        cls._pick_info = [{'product': cls.apple, 'qty': 4}]
+        cls._constraint = cls.picking_type_putaway.u_drop_location_constraint
+        cls._policy     = cls.picking_type_putaway.u_drop_location_policy
```
nocturnalastro: should only one space around an "="

addons/udes_stock/tests/test_move_line.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def tearDown(self):
+        super(TestValidateLocationDest, self).tearDown()
+        self.picking_type_putaway.u_drop_location_constraint = self._constraint
+        self.picking_type_putaway.u_drop_location_policy     = self._policy
```
nocturnalastro: see above

addons/udes_stock/tests/test_move_line.py
```diff
@@ -0,0 +1,108 @@
...
+                                      assign=True)
+        mls = picking.move_line_ids
+        locations = picking.get_suggested_locations(mls)
+
+        # We'll use loc 02 to drop off, so we check the assumption
+        self.assertEqual(locations, self.test_location_01)
+
+        for ml in mls:
+            ml.write({'location_dest_id': self.test_location_02.id})
+            self.assertTrue(True)
```
nocturnalastro: why do this? it literally does nothing

addons/udes_stock/tests/test_move_line.py
```diff
@@ -0,0 +1,108 @@
...
+                                      assign=True)
+        mls = picking.move_line_ids
+        locations = picking.get_suggested_locations(mls)
+
+        # We'll use loc 01 to drop off, so we check the assumption
+        self.assertEqual(locations, self.test_location_01)
+
+        for ml in mls:
+            ml.write({'location_dest_id': self.test_location_01.id})
+            self.assertTrue(True)
```
nocturnalastro: see above

addons/udes_stock/tests/test_move_line.py
nocturnalastro: also probably better doing a write instead of two assignments

addons/udes_stock/models/stock_move_line.py
```diff
@@ -654,5 +657,17 @@ def _validate_location_dest(self):
...
                 raise ValidationError(
-                    _("The location '%s' is not a child of the picking destination "
-                      "location '%s'" % (location.name, picking.location_dest_id.name)))
+                    _("The location '%s' is not a child of the picking "
+                      "destination location '%s'" %
+                      (location.name, picking.location_dest_id.name)))
+
+            if picking.picking_type_id.u_drop_location_constraint == 'enforce' \
+                    and picking.state == 'assigned':
+                mls = self.filtered(lambda ml: ml.picking_id == picking)
```
nocturnalastro: Also as this effectively doing a self.groupby('picking_id') 
We are going to be adding edi as a dependency of udes_stock till that functionality is factored out.
Might be worth looking at.

addons/udes_stock/tests/test_move_line.py
parisiale: It tells you're testing.

addons/udes_stock/tests/test_move_line.py
parisiale: Ah, you don't like vertical alignment, ok, my eyes will blame you.

addons/udes_stock/tests/test_move_line.py
parisiale: To be honest, the whole setup / teardown dance is superfluous.

addons/udes_stock/tests/test_move_line.py
nocturnalastro: pylint(pep8) doesn't like it. Sorry

addons/udes_stock/models/stock_move_line.py
parisiale: Not sure I understand what you want here.
Even if you rely on `groupby`, you still to select the group with the picking id you're interested in, no?
What's the benefit?
The `filtered` call is pretty clear to me.

addons/udes_stock/tests/test_move_line.py
nocturnalastro: It by very definition can never fail. I'm missing something here :confused:

addons/udes_stock/models/stock_move_line.py
nocturnalastro: you can do 
`for picking, mls in self.groupby('picking_id'):`
I personally think its a bit neater but I don't have strong feeling about this just thought I'd mention it.

addons/udes_stock/tests/test_move_line.py
parisiale: Code is meant to be read. Don't miss that.

addons/udes_stock/models/stock_move_line.py
parisiale: Ah, makes sense, I'll add a TODO.

addons/udes_stock/tests/test_move_line.py
parisiale: Ah, like we're pep8ing everywhere :D

nocturnalastro: We should be :laughing:

nocturnalastro: Also why I suggested using an auto-formatter.

PR #40, opened 2018-10-08
story/1688: Putaway suggestions preprocessing

nocturnalastro:
Add the ability to preprocess suggested locations and add them to a move line
this allows the location to effectively reserved (for policies which check for this).

Also implements a policy to suggest locations based on height/speed category and also doing the aforementioned check

### DIFF
--- a/addons/udes_stock/README.md
+++ b/addons/udes_stock/README.md
@@ -115,6 +115,7 @@ A lot of custom UDES functionality is specfied at the picking type level. This i
 | u_validate_real_time       | boolean | Do we validate move lines in real time. |
 | u_drop_location_constraint | string  | Whether drop location should be scanned, suggested and, then, enforced (enum: 'dont_scan', 'scan', 'enforce', 'suggest'); default: 'scan'. |
 | u_drop_location_policy     | string  | To indicate the policy for suggesting drop locations (enum: 'exactly_match_move_line', 'by_products', 'by_packages'); default: 'exactly_match_move_line'. |
+| u_drop_location_preprocess  | boolean | Selects if suggestions u_drop_location_policy should be added on assignment. If this is set will apply the u_drop_location_policy and the first location is set as location_dest_id for the set of move_lines. This can only be used with polcies decorated with allow_preprocess, usage on other polcies will result in an ValidationError.|
 | u_display_summary          | string  | How to display the Source Document and a summary of all Package Names associated with that Source Document number at Goods-Out (enum: 'none', 'list', 'list_contents'). |
 | u_handle_partials          | boolean | If the picking type is allowed to handle partially available pickings. If True, then pickings of this type will report their u_pending value. |
 | u_create_procurement_group | boolean | Indicate if a procurement group should be created on confirmation of the picking if one does not already exist. |
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -144,10 +144,19 @@ class StockMove(models.Model):
         return bk_move
 
     def _action_assign(self):
+        """Extend _acction_assgin to merge/split moves and preprocess
+           location suggestions
+        """
+
         res = super(StockMove, self)._action_assign()
-        for picking_type in self.mapped('picking_type_id'):
-            self.filtered(lambda m: m.picking_type_id == picking_type). \
-                post_reservation_split()
+        for picking_type_id in self.mapped('picking_type_id'):
+            moves = self.filtered(
+                lambda m: m.picking_type_id == picking_type_id
+            )
+            moves.post_reservation_split()
+            if picking_type_id.u_drop_location_preprocess:
+                moves.mapped('picking_id').apply_drop_location_policy()
+        return res
 
     def post_reservation_split(self):
         """
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1352,12 +1352,12 @@ class StockPicking(models.Model):
             raise ValidationError(
                 _('Products missing to suggest location for.'))
 
-        quants = Quant.search(
-            [('product_id', 'in', products.ids),
-             ('location_id', 'child_of', self.location_dest_id.ids)])
-
-        suggested_locations = quants.mapped('location_id') \
-            .filtered(lambda loc: not loc.u_blocked and loc.barcode)
+        suggested_locations = Quant.search([
+            ('product_id', 'in', products.ids),
+            ('location_id', 'child_of', self.location_dest_id.ids),
+            ('location_id.u_blocked', '=', False),
+            ('location_id.barcode', '!=', False),
+        ]).mapped('location_id')
 
         if not suggested_locations:
             # No drop locations currently used for this product;
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1292,9 +1292,18 @@ class StockPicking(models.Model):
 
         for picking in self:
             policy = picking.picking_type_id.u_drop_location_policy
-
             if policy:
-                func = getattr(self, '_get_suggested_location_' + policy, None)
+                if move_line_ids and \
+                        picking.picking_type_id.u_drop_location_preprocess and \
+                        not move_line_ids.any_destination_locations_default():
+                    # The policy has been preprocessed this assumes the
+                    # the polciy is able to provide a sensible value (this is
+                    # not the case for every policy)
+                    # Use the preselected value
+                    func = self._get_suggested_location_exactly_match_move_line
+                else:
+                    func = getattr(self, '_get_suggested_location_' + policy,
+                                   None)
 
                 if func:
                     result = func(move_line_ids)
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1385,3 +1385,54 @@ class StockPicking(models.Model):
         products = mls.mapped('product_id')
 
         return self._get_suggested_location_by_products(mls, products)
+
+    @allow_preprocess
+    def _get_suggested_location_by_height_speed(self, move_line_ids):
+        """Get location based on product height and turn over speed categories
+        """
+        Location = self.env['stock.location']
+
+        height_category = move_line_ids.mapped(
+            'product_id.u_height_category_id'
+        )
+        speed_category = move_line_ids.mapped('product_id.u_speed_category_id')
+        default_location = move_line_ids.mapped('location_dest_id')
+
+        if not len(height_category) == 1 or not len(speed_category) == 1:
+            raise UserError(
+                _('Move lines with more than category for height(%s) or '
+                  'speed(%s) provided') % (
+                      height_category.mapped('name'),
+                      speed_category.mapped('name'),
+                  )
+            )
+        default_location.ensure_one()
+
+        # Get empty locations where height and speed match product
+        candidate_locations = Location.search([
+            ('location_id', 'child_of', default_location.id),
+            ('u_blocked', '=', False),
+            ('barcode', '!=', False),
+            ('u_height_category_id', '=', height_category.id),
+            ('u_speed_category_id', '=', speed_category.id),
+            # TODO(MTC): This should probably be a bit more inteligent perhaps
+            # get them all then do a filter for checking if theres space
+            ('quant_ids', '=', False),
+        ])
+
+        return self._location_not_in_other_move_lines(candidate_locations)
+
+    def _location_not_in_other_move_lines(self, candidate_locations):
+        MoveLines = self.env['stock.move.line']
+
+        for _indexes, valid_locations in candidate_locations.batched(size=1000):
+            # remove locations which are already used in move lines
+            valid_locations -= MoveLines.search([
+                ('picking_id.state', '=', 'assigned'),
+                ('location_dest_id', 'in', valid_locations.ids),
+            ]).mapped('location_dest_id')
+
+            if valid_locations:
+                return valid_locations
+
+        return self.env['stock.location']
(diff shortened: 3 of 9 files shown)


addons/udes_stock/README.md
```diff
@@ -115,6 +115,7 @@ A lot of custom UDES functionality is specfied at the picking type level. This i
 | u_validate_real_time       | boolean | Do we validate move lines in real time. |
 | u_drop_location_constraint | string  | Whether drop location should be scanned, suggested and, then, enforced (enum: 'dont_scan', 'scan', 'enforce', 'suggest'); default: 'scan'. |
 | u_drop_location_policy     | string  | To indicate the policy for suggesting drop locations (enum: 'exactly_match_move_line', 'by_products', 'by_packages'); default: 'exactly_match_move_line'. |
+|u_preprocess_drop_location  | boolean | Selects if suggestions u_drop_location_policy should be added on assignment |
```
parisiale: Bikeshedding: I would call this `u_drop_lcoation_preprocess` so it starts like the 2 entries above. It would possibly help searching and make it clear that the 3 things are related.
No strong opinion though.

parisiale: I would add a few details about how you add the suggestions (i.e. use the first suggestion as the `location_dest_id` of the move lines).

addons/udes_stock/models/stock_move.py
```diff
@@ -146,8 +146,13 @@ def split_out_move_lines(self, move_lines):
...
         for picking_type in self.mapped('picking_type_id'):
-            self.filtered(lambda m: m.picking_type_id == picking_type). \
-                post_reservation_split()
+            moves = self.filtered(lambda m: m.picking_type_id == picking_type)
+            moves.post_reservation_split()
+            if picking_type.u_preprocess_drop_location:
+                moves.mapped('picking_id').apply_drop_location_policy()
+
+
+
```
parisiale: Style: extra empty lines, inconsistent with the rest.

addons/udes_stock/models/stock_move.py
```diff
@@ -146,8 +146,13 @@ def split_out_move_lines(self, move_lines):
     def _action_assign(self):
         res = super(StockMove, self)._action_assign()
         for picking_type in self.mapped('picking_type_id'):
-            self.filtered(lambda m: m.picking_type_id == picking_type). \
-                post_reservation_split()
+            moves = self.filtered(lambda m: m.picking_type_id == picking_type)
```
parisiale: Above, perhaps it's better to use `_` instead of defining `res` as it's not used.

parisiale: Nitpick: I would rename `picking_type` to `picking_type_id` to improve readability.

parisiale: Can we add a brief docstring to `_action_assing`?

addons/udes_stock/models/stock_move.py
```diff
@@ -146,8 +146,13 @@ def split_out_move_lines(self, move_lines):
...
+            moves.post_reservation_split()
+            if picking_type.u_preprocess_drop_location:
+                moves.mapped('picking_id').apply_drop_location_policy()
```
parisiale: Question: as we are iterating the picking types and, based on that, filtering the move lines, we can be sure that the `apply_drop_location_policy` logic (such method iterates all the recordset instances) won't be applied to the same picking multiple times, right?

addons/udes_stock/models/stock_move.py
parisiale: Looking at the `picking::action_assign` method: do we need to validate the `res` returned by the parent class method?

addons/udes_stock/models/stock_picking.py
```diff
@@ -983,6 +983,13 @@ def action_assign(self):
         """
             Override action_assign to reserve full packages if applicable
         """
+
+        # This is only called when manually clicking check_availitbily
+        # or called drectly.
+        # This isn't called when next picking is assigned which is done on the
+        # move level
+        # TODO: Check this!
```
parisiale: It's not clear to me what are the consequences of the above.
Is there anything we should to fix / clarify that?
If not, can you improve the message?

addons/udes_stock/models/stock_picking.py
```diff
@@ -1235,10 +1254,15 @@ def get_suggested_locations(self, move_line_ids):
...
-                func = getattr(self, '_get_suggested_location_' + policy, None)
-
+                default_dest = picking.picking_type_id.default_location_dest_id
+                if move_line_ids and all(ml.location_dest_id != default_dest
+                                         for ml in move_line_ids):
+                    # If we have pre selected a location use it
+                    func = self._get_suggested_location_exactly_match_move_line
+                else:
+                    func = getattr(self, '_get_suggested_location_' + policy,
+                                   None)
```
parisiale: Please add an empty line below.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1235,10 +1254,15 @@ def get_suggested_locations(self, move_line_ids):
...
             policy = picking.picking_type_id.u_drop_location_policy
-
             if policy:
-                func = getattr(self, '_get_suggested_location_' + policy, None)
-
+                default_dest = picking.picking_type_id.default_location_dest_id
+                if move_line_ids and all(ml.location_dest_id != default_dest
+                                         for ml in move_line_ids):
+                    # If we have pre selected a location use it
+                    func = self._get_suggested_location_exactly_match_move_line
```
parisiale: Not sure this will work, but I don't think I have the full picture of what you're trying to achieve.

The allocation during the assignment trivially uses the first suggested location of the configured policy. This does not prevent a given location being used for multiple move line assignments. So, at this point, using the `exactly_match_move_line` policy may (likely) result in a bad suggestion.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1278,13 +1302,13 @@ def _get_suggested_location_by_products(self, move_line_ids, products=None):
...
             raise ValidationError(
                 _('Products missing to suggest location for.'))
-
-        quants = Quant.search(
-            [('product_id', 'in', products.ids),
-             ('location_id', 'child_of', self.location_dest_id.ids)])
-
-        return quants.mapped('location_id') \
-            .filtered(lambda loc: not loc.u_blocked and loc.barcode)
+        quants = Quant.search([
```
parisiale: Pleas add empty lines above and below the assignment.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1302,3 +1326,49 @@ def _get_suggested_location_by_packages(self, move_line_ids):
...
+            ('u_height_category_id', '=', height_category.id),
+            ('u_speed_category_id', '=', speed_category.id),
+            # This should probably be a bit more inteligent perhaps get them
+            #  all then do a filter for checking if theres space
+            ('quant_ids', '=', False),
+        ])
+        return self._location_not_in_other_move_lines(candidate_locations)
+
+    def _location_not_in_other_move_lines(self, candidate_locations):
+        MoveLines = self.env['stock.move.line']
```
parisiale: Extra empty line below.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1302,3 +1326,49 @@ def _get_suggested_location_by_packages(self, move_line_ids):
...
+        candidate_locations = Location.search([
+            ('location_id', 'child_of', default_location.id),
+            ('u_blocked', '=', False),
+            ('barcode', '!=', False),
+            ('u_height_category_id', '=', height_category.id),
+            ('u_speed_category_id', '=', speed_category.id),
+            # This should probably be a bit more inteligent perhaps get them
+            #  all then do a filter for checking if theres space
+            ('quant_ids', '=', False),
+        ])
```
parisiale: Empty lines before `return` are sexy!

addons/udes_stock/models/stock_picking.py
```diff
@@ -1302,3 +1326,49 @@ def _get_suggested_location_by_packages(self, move_line_ids):
...
+
+        height_category = move_line_ids.mapped(
+            'product_id.u_height_category_id'
+        )
+        speed_category = move_line_ids.mapped('product_id.u_speed_category_id')
+        default_location = move_line_ids.mapped('location_dest_id')
+
+        # Check categories are constiant
+        # probably better doing a len(x) > 1 or len(y) > 1
+        # so we can return a meaningful error message
```
parisiale: Is the above a TODO msg? If so, please add the label. Otherwise, I suggest to make it more clear.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1302,3 +1326,49 @@ def _get_suggested_location_by_packages(self, move_line_ids):
...
+
+        # Get empty locations where height and speed match product
+        candidate_locations = Location.search([
+            ('location_id', 'child_of', default_location.id),
+            ('u_blocked', '=', False),
+            ('barcode', '!=', False),
+            ('u_height_category_id', '=', height_category.id),
+            ('u_speed_category_id', '=', speed_category.id),
+            # This should probably be a bit more inteligent perhaps get them
+            #  all then do a filter for checking if theres space
```
parisiale: Same as above: TODO?

addons/udes_stock/models/stock_picking.py
```diff
@@ -1302,3 +1326,49 @@ def _get_suggested_location_by_packages(self, move_line_ids):
...
+            ('quant_ids', '=', False),
+        ])
+        return self._location_not_in_other_move_lines(candidate_locations)
+
+    def _location_not_in_other_move_lines(self, candidate_locations):
+        MoveLines = self.env['stock.move.line']
+
+
+        for _range, valid_locations in candidate_locations.batched(size=1000):
```
parisiale: Perhaps, use a descriptive token for your magic value.

addons/udes_stock/README.md
parisiale: Also, bad alignment (missing space before field name). I'll stop here with the style comments, promised.

addons/udes_stock/models/stock_picking.py
parisiale: As agreed by voice, consider:
 - checking the pre-process flag as part of the if clause;
 - factor out the move line loc dest check to be a move line method and document the assumptions, just for the sake of readability.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1235,10 +1254,15 @@ def get_suggested_locations(self, move_line_ids):
...
         for picking in self:
             policy = picking.picking_type_id.u_drop_location_policy
-
             if policy:
-                func = getattr(self, '_get_suggested_location_' + policy, None)
-
+                default_dest = picking.picking_type_id.default_location_dest_id
+                if move_line_ids and all(ml.location_dest_id != default_dest
+                                         for ml in move_line_ids):
+                    # If we have pre selected a location use it
```
parisiale: Remove the 'If' as we're in the `if` :P

addons/udes_stock/models/stock_move.py
nocturnalastro: moves.mapped('picking_id') will de-duplicate the pickings and a picking only has one picking_type so yeah it will only be applied once per pick.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1227,6 +1239,36 @@ def _new_picking_for_group(self, group_key, moves, move_lines):
...
         return picking
 
+    def apply_drop_location_policy(self):
+        """Apply suggested locations to move lines
+
+           raise ValidationError: if the policy set does not have
+                                  _allow_preprocess set
+        """
+
+        def check_policy_is_valid(policy):
```
parisiale: As you're not bringing anything from the outer method scope into a closure, I think it would be better to make this a free function instead of a nested one, so you won't be instantiating a new function object at each method invocation.

parisiale:
Thanks @nocturnalastro ; I left a minor comment about a nested function.

parisiale:
👍 
You can squash now.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from collections import OrderedDict, defaultdict
+from functools import wraps
```
parisiale: This seems unused.

addons/udes_stock/README.md
```diff
@@ -115,6 +115,9 @@ A lot of custom UDES functionality is specfied at the picking type level. This i
 | u_validate_real_time       | boolean | Do we validate move lines in real time. |
 | u_drop_location_constraint | string  | Whether drop location should be scanned, suggested and, then, enforced (enum: 'dont_scan', 'scan', 'enforce', 'suggest'); default: 'scan'. |
 | u_drop_location_policy     | string  | To indicate the policy for suggesting drop locations (enum: 'exactly_match_move_line', 'by_products', 'by_packages'); default: 'exactly_match_move_line'. |
+| u_drop_location_preprocess  | boolean | Selects if suggestions u_drop_location_policy should be added on assignment.
+If this is set will apply the u_drop_location_policy and the first location is set as location_dest_id for the set of move_lines.
+This can only be used with polcies decorated with allow_preprocess, usage on other polcies will result in an ValidationError.|
```
parisiale: Description is in the 'Field Name' column.

parisiale:
@nocturnalastro 	two more comments.

addons/udes_stock/models/stock_picking.py
nocturnalastro: Its used in the decorator :)

nocturnalastro: Your right!

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -209,3 +229,278 @@ def test08_get_suggested_locations_by_products_missing_parameter(self):
 
         msg = 'Cannot determine the suggested location: missing move lines'
         self.assertEqual(err.exception.name, msg)
+
+    def test07_get_suggested_locations_by_products_locations_blocked(self):
+        """Locations which are blocked should not appear in suggestions"""
```
mcb30: This docstring appears to be copied from another test?

PR #42, opened 2018-10-09
[ADD] hds: Added privacy packaging and return that field in api call

unirho:
User-story: 1521


addons/udes_stock/models/stock_move.py
```diff
@@ -41,7 +41,6 @@ def _prepare_info(self):
                 "quantity_done": self.quantity_done,
                 "product_id": self.product_id.get_info()[0],
                 "moves_line_ids": self.move_line_ids.get_info(),
-                "product_packaging": self.product_packaging.name,
```
mcb30: This appears to undo a change in an earlier commit from this series: is this deliberate?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -544,7 +544,6 @@ def _prepare_info(self):
                 "product_uom_qty": self.product_uom_qty,
                 "qty_done": self.qty_done,
                 "write_date": self.write_date,
-                "product_packaging": self.product_packaging.name
```
mcb30: This appears to undo a change in an earlier commit from this series: is this deliberate?

unirho: yes, this was owing to the long conversation we all had about where this flag was desired in the backend.

PR #47, opened 2018-10-10
WiP: [IMP] udes_stock: Push stock from a drop location.

crwr45:
In order to allow stock to be diverted to a "Damaged" location in
Inbound if a damaged pallet is received, push rules can now be triggered
by dropping stock into a location from which a push rule originates.
If such a rule exists when dropping stock, a new set of moves are
created, linked back to the moves that dropped the stock into that
location, and reserved, which in turn creates the rest of the push
route.

This allows for routes to fork, or to merge (untested) by dropping into
a location.

This behaviour can be disabled with a flag on the push rule.

Story: 1775
Signed-off-by: Charlie Wheeler-Robinson <crwr@pigotts.org.uk>

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -209,7 +209,11 @@ class StockMove(models.Model):
             move_vals = base_vals.copy()
             move_vals.update({
                 'product_uom_qty': sum(mls.mapped('qty_done')),
-                'move_orig_ids': [(6, 0, [move.id,])]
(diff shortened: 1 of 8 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from odoo import api, models, _
+from odoo.exceptions import UserError
```
mcb30: Seems to undo a deletion in the previous commit?  If so, should be squashed

PR #54, opened 2018-10-24
Story/1942 Micky Group moves by key

nocturnalastro:


addons/udes_stock/models/stock_move.py
```diff
@@ -310,13 +321,64 @@ def _create_moves_for_push(self, push, move_lines):
             created_moves |= move.copy(move_vals)
         return created_moves
 
+    def group_by_key(self):
+        # force recompute on u_grouping_key so we have an up-to-date key:
+        self._fields['u_grouping_key'].compute_value(self)
+
+        if any(pt.u_move_key_format is False
+               for pt in self.mapped('picking_id.picking_type_id')):
+            raise UserError(_("Cannot group move lines when their picking type"
```
crwr45: move instead of move line

addons/udes_stock/models/stock_move.py
```diff
@@ -310,13 +321,64 @@ def _create_moves_for_push(self, push, move_lines):
...
+                              "has no grouping key set."))
+
+        return self.groupby(lambda ml: ml.u_grouping_key)
+
+    def refactor_action_group_by_move_key(self):
+        """
+        group the moves by the splitting criteria
+        for each resulting group of stock.moves:
+            create a new picking
+            split any stock.move records that are only partially covered by the
```
crwr45: Moves are not going to be split as part of this. Looks like you removed the one from the move line method

addons/udes_stock/models/stock_move.py
```diff
@@ -310,13 +321,64 @@ def _create_moves_for_push(self, push, move_lines):
...
+
+        empty_picks = pickings.filtered(lambda p: len(p.move_lines) == 0)
+        if empty_picks:
+            _logger.info(_("Cancelling empty picks after splitting."))
+            # action_cancel does not cancel a picking with no moves.
+            empty_picks.write({
+                'state': 'cancel',
+                'is_locked': True
+            })
+
```
crwr45: Need to return the union of all `move_group`s at the end

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -2,7 +2,8 @@
 from odoo import api, fields, models, _
 from odoo.exceptions import ValidationError
 
-core_lifecycle_actions = [
+CORE_LIFECYCLE_ACTIONS = [
+        ('none', 'None'),  # TODO: can we unset if no inactive?
```
crwr45: Can remove the `none` option here.

addons/udes_stock/models/stock_move.py
nocturnalastro: I thought I had already done that :confused:

addons/udes_stock/models/stock_move.py
nocturnalastro: wouldn't that just be self?

PR #57, opened 2018-10-25
[WIP] endpoint change putaway exception handling

nocturnalastro:

### DIFF
--- a/addons/udes_stock/controllers/stock_location.py
+++ b/addons/udes_stock/controllers/stock_location.py
@@ -71,3 +71,51 @@ class Location(UdesApi):
         location = Location.get_location(location_id)
 
         return location.process_perpetual_inventory_request(pi_request)
+
+    @http.route('/api/stock-location/block/',
+                type='json', methods=['POST'], auth='user')
+    def block(self, reason,
+              location_id=None, location_name=None, location_barcode=None):
+        """
+            Blocks the specified location and creates a Stock
+            Investigation for it.
+
+            Raises a ValidationError if the specified location is
+            unknown or already blocked.
+
+            Returns True in case of success.
+
+            @param <ident>  location_id of the location to block
+            @param reason   Reason for the location being blocked
+        """
+        Location = request.env['stock.location']
+        Picking = request.env['stock.picking']
+        ResUsers = request.env['res.users']
+
+        identifier = location_id or location_name or location_barcode
+
+        if not identifier:
+            raise ValidationError(
+                _('You need to provide an id, name or barcode for the location.'))
+
+        location = Location.get_location(identifier)
+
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -334,63 +340,64 @@ def unpickable_item(self, reason, product_id=None, location_id=None,
...
-            picking.unlink()
-
-        # If the batch does not contain any remaining picking to do,
-        # it can be set as done
-        remaining_pickings = self.picking_ids.filtered(
-            lambda x: x.state in ['assigned'])
-
-        if not remaining_pickings.exists():
-            self.state = 'done'
+        if raise_stock_investigation:
```
parisiale: Can you just `return True` otherwise instead of having a big clause, for the sake of readability?

addons/udes_stock/controllers/stock_location.py
```diff
@@ -68,6 +68,49 @@ def pi_count(self, pi_request):
             raise ValidationError(
                 _('You need to provide a valid id for the location.'))
 
-        location = Location.get_location(location_id)
+        location  = Location.get_location(location_id)
```
joe-gb: Accident, is removed in next commit. ( Will squash before we merge )

addons/udes_stock/controllers/stock_location.py
```diff
@@ -71,3 +71,46 @@ def pi_count(self, pi_request):
         location = Location.get_location(location_id)
 
         return location.process_perpetual_inventory_request(pi_request)
+
+    @http.route('/api/stock-location/<ident>/block/',
```
mcb30: Could use '/api/stock-location/<int:ident>/block': this would automatically perform the validation of ident as a non-empty integer.

The trailing slash on the URI seems incorrect: is there a reason for it?

addons/udes_stock/controllers/stock_location.py
```diff
@@ -71,3 +71,46 @@ def pi_count(self, pi_request):
...
+        """
+        Location = request.env['stock.location']
+        Picking = request.env['stock.picking']
+        ResUsers = request.env['res.users']
+
+        location_id = ident
+
+        if location_id is None:
+            raise ValidationError(
+                _('You need to provide a valid id for the location.'))
```
mcb30: Not needed: see above.

addons/udes_stock/controllers/stock_location.py
```diff
@@ -71,3 +71,46 @@ def pi_count(self, pi_request):
...
+
+        location = Location.get_location(int(location_id))
```
mcb30: int() not needed: see above

PR #63, opened 2018-10-27
Story/1934

ws-mps:


addons/udes_stock/models/stock_move.py
```diff
@@ -156,11 +156,15 @@ def split_out_move_lines(self, move_lines):
                 'move_orig_ids': [],
                 'ordered_qty': total_ordered_qty,
                 'product_uom_qty': total_initial_qty,
-                'state': 'assigned',
+                'state': self.state,
+                # TODO: ^^^ Not always true if we split a move which
+                # is partially availabe
```
crwr45: This needs to be resolved.
Take a look at either `stock.move._get_relevant_state_among_moves()` or `stock.move._recompute_state()` from Odoo core stock module to see which is the most appropriate.

addons/udes_stock/models/stock_move.py
```diff
@@ -156,11 +156,15 @@ def split_out_move_lines(self, move_lines):
...
             })
             move_lines.write({
                 'move_id': bk_move.id,
-                'state': 'assigned',
+                'state': self.state,
+                # TODO: ^^^ Not always true if we split a move which
+                # is partially availabe
```
crwr45: Same as previous comment.

addons/udes_stock/models/stock_move.py
```diff
@@ -443,9 +454,13 @@ def refactor_action_group_by_move_line_key(self):
...
                     group_moves |= move
 
-            Picking._new_picking_for_group(key, group_moves)
+            origin = None
+            if len(group_pickings) == 1:
+                origin = group_pickings.origin
+            Picking._new_picking_for_group(key, group_moves, origin=origin)
             result_moves |= group_moves
 
+        # NOTE: empty pick = no moves and no move_lines
```
crwr45: I don't think this comment is necessary, the filtered should be self explanatory.

addons/udes_stock/models/stock_move.py
```diff
@@ -330,7 +334,7 @@ def _create_moves_for_push(self, push, move_lines):
                 # TODO: reservation to the src location of a move if it has
                 # TODO: preceeding moves.
                 # TODO: Uncomment this, and unskip tests, when story/1934 is completed
-                # 'move_orig_ids': [(6, 0, [move.id,])]
+                'move_orig_ids': [(6, 0, [move.id,])]
```
crwr45: Doesn't this also then need `move_dest_ids` to be set on the previous move, as with the related change from https://taiga.udes.io/project/udescore/task/2030?kanban-status=87?

addons/udes_stock/models/stock_move.py
```diff
@@ -384,7 +388,12 @@ def refactor_action_group_by_move_key(self):
                       '') % (key, move_group.mapped('location_id'),
                              move_group.mapped('location_dest_id')))
 
-            Picking._new_picking_for_group(key, move_group)
+            group_pickings = move_group.mapped('picking_id')
+            origin = None
+            if len(group_pickings) == 1:
```
crwr45: This should set the origin if there are multiple group picks with the same origin. Something like:
```
origins = set(group_pickings.mapped('origin'))
origin = None
if len(origins) == 1:
    origin = origins[0]

```

addons/udes_stock/models/stock_move.py
```diff
@@ -443,9 +454,13 @@ def refactor_action_group_by_move_line_key(self):
                 else:
                     group_moves |= move
 
-            Picking._new_picking_for_group(key, group_moves)
+            origin = None
+            if len(group_pickings) == 1:
+                origin = group_pickings.origin
+            Picking._new_picking_for_group(key, group_moves, origin=origin)
```
crwr45: Again, keep the origin if all are the same.

addons/udes_stock/models/stock_move.py
ws-mps: The filtered is wrong, we should fix it.

ws-mps: Just ignore my comment misread move_lines by move_line_ids

addons/udes_stock/models/stock_move.py
nocturnalastro: I think we only need to set the one as they are related fields, when I checked it was linked already

addons/udes_stock/models/stock_move.py
```diff
@@ -326,11 +334,7 @@ def _create_moves_for_push(self, push, move_lines):
             move_vals = base_vals.copy()
             move_vals.update({
                 'product_uom_qty': sum(mls.mapped('qty_done')),
-                # FIXME: CWR: This is commented out because odoo does not limit
-                # TODO: reservation to the src location of a move if it has
-                # TODO: preceeding moves.
-                # TODO: Uncomment this, and unskip tests, when story/1934 is completed
-                # 'move_orig_ids': [(6, 0, [move.id,])]
+                'move_orig_ids': [(6, 0, move.ids)]
```
crwr45: Very pleased to see that TODO resolved.

PR #71, opened 2018-11-02
[IMP] udes_stock: Use same logic for unassigning batches in logout an…

joe-gb:
drop_off_picked and unassign_user_batch now use the same logic and function so that the functionalities now match.

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -262,8 +262,7 @@ class StockPickingBatch(models.Model):
         Validate the move lines of the batch (expects a singleton)
         by moving them to the specified location.
 
-        In case continue_batch is flagged, mark the batch as
-        'done' if appropriate.
+        In case continue_batch is flagged, unassign the batch
         """
         self.ensure_one()
 
@@ -294,11 +293,8 @@ class StockPickingBatch(models.Model):
                 # at this point pick_todo should contain only mls done
                 pick_todo.update_picking(validate=True)
 
-        incomplete_picks = self.picking_ids.filtered(
-            lambda x: x.state not in ['done', 'cancel'])
-
         if not continue_batch:
-            incomplete_picks.write({'batch_id': False})
+            self.unassign()
 
         return self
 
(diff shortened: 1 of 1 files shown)


parisiale:
Thanks for this!

Sorry for being pedantic, but I'm afraid your intent is not clear in your commit message; you're describing the changes, that are clear by simply looking at the diff, but not your reasons and objectives. I would assume you're looking for consistency, but what's the real benefit?. 
Describing the intent is fundamental for the reviewer, but most importantly for the future reader of the commit tree. And that should be focus of the description.

Also, you're not separating the commit title (gigantic) from the description. I suggest to have a look at https://chris.beams.io/posts/git-commit/.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -506,16 +503,19 @@ def get_user_batches(self):
...
     def unassign_user_batches(self):
-        """ Removes pickings from all batches assigned to the user and cancels the batch
+        """ Get batches for user and unassign them
         """
         # Get user batches
-        batches = self.get_user_batches()
+        self.get_user_batches().unassign()
 
+    def unassign(self):
+        """ Removes incomplete pickings and removes responsible user from batches
```
parisiale: Docstring looks good, but would it make sense to have two methods instead, one to unassign another to remove the incomplete pickings, so to improve naming and functionality clarity?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -506,16 +503,19 @@ def get_user_batches(self):
         return batches
 
     def unassign_user_batches(self):
-        """ Removes pickings from all batches assigned to the user and cancels the batch
+        """ Get batches for user and unassign them
```
parisiale: You're also removing the incomplete pickings.

addons/udes_stock/models/stock_picking_batch.py
joe-gb: Probably not, it's one method - the act of "unassigning" the batch is to follow the logic of the "unassign" function as defined: It's to unassign the user AND the incomplete pickings IF it's ephemeral

parisiale: 👍

addons/udes_stock/models/stock_picking_batch.py
joe-gb: yes! unassigning a batch is to do just that - if it's ephemeral

joe-gb:
> Thanks for this!
> 
> Sorry for being pedantic, but I'm afraid your intent is not clear in your commit message; you're describing the changes, that are clear by simply looking at the diff, but not your reasons and objectives. I would assume you're looking for consistency, but what's the real benefit?
> Describing the intent is fundamental for the reviewer and for the future reader of the commit tree. And that should be focus of the description.
> 
> Also, you're not separating the commit title (gigantic) from the description. I suggest to have a look at https://chris.beams.io/posts/git-commit/.

Tidied up the commit message

parisiale:
👍

PR #79, opened 2018-11-08
[IMP] Printer config check

parisiale:


parisiale:
We decided for a different approach - closing this.

PR #80, opened 2018-11-08
Task/2147

ws-mps:


addons/udes_stock/models/stock_move_line.py
```diff
@@ -775,3 +776,24 @@ def any_destination_locations_default(self):
         )
         default_dest.ensure_one()
         return any(ml.location_dest_id == default_dest for ml in self)
+
+    def package_name_policy(self):
```
mcb30: Please rename this method to "new_package_name", since it returns a name (rather than returning something representing a policy).

addons/udes_stock/models/stock_move_line.py
```diff
@@ -775,3 +776,24 @@ def any_destination_locations_default(self):
...
+        """ Given a move line compute the next package name according to
+            the policy assigned to the its picking picking type.
+            If no policy is assigned the default policy is applied.
+        """
+        self.ensure_one()
+        policy = self.picking_id.picking_type_id.u_new_package_policy
+        name = None
+        if policy:
+            func = getattr(self, 'new_package_' + policy, None)
```
mcb30: May be sensible to have the constructed methods called "new_package_name_" + policy, since they return a name (rather than returning a package).

addons/udes_stock/models/stock_move_line.py
```diff
@@ -775,3 +776,24 @@ def any_destination_locations_default(self):
...
+            if func:
+                name = func()
+        # Always return a name
+        if not name:
+            name = self.new_package_default()
+
+        return name
+
+    def new_package_default(self):
```
mcb30: new_package_name_default

addons/udes_stock/models/stock_move_line.py
```diff
@@ -39,6 +39,7 @@ def _prepare_result_packages(self, package, result_package, products_info):
         picking = self.mapped('picking_id')
         picking.ensure_one()
         target_storage_format = picking.picking_type_id.u_target_storage_format
+        Package = Package.with_context(move_line_id=self[0].id)
```
mcb30: The semi-arbitrary [0] here suggests that we should maybe pass move_line_ids in the context, rather than move_line_id.  The subclass's stock.move.line.new_package_name_xxx() method can then verify that the move lines are actually allowed to be placed into the same package.   For example: a policy that requires all move lines to belong to the same sale.order could use self.mapped('move_id.sale_line_id.order_id').ensure_one()

PR #84, opened 2018-11-09
[IMP] udes-open: automated printing for stock picking

spenced:
Trigger automated printing strategies for stock pickings
by picking type when the picking is done (completed).


addons/udes_stock/tests/test_picking_print_strategy.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    def default_picking_type(self):
+        """Return the default picking type"""
+        return self.picking_type_in
+
+    def action_context(self, obj):
+        return super().action_context(self.picking_in)
+
+    def benign_context(self, obj):
+        return super().action_context(self.picking_pick)
```
mcb30: Is it intentional that this method calls a superclass method with a different name?

spenced: It is. The base action_context() is the single place which knows how to build a dictionary for the with_context(context).run() calls.

PR #89, opened 2018-11-13
Automatic printing of shipping labels (task 2152)

richardwhitefoot:


addons/udes_stock/data/picking_print.xml
```diff
@@ -11,6 +11,15 @@
             <field name="strategy_id" ref="model_udes_stock_stock_picking_print_strategy"/>
         </record>
 
+        <!-- Print on Picking Update event -->
```
parisiale: Bare with me, as I'm not very familiar with this...
When a picking is finished, the last thing the frontend invokes is `pick::update` that ends up calling `StockPicking::update_picking` where the print operation is implemented.
Does it means that we would have two printings due to the final `picking_update` and the above `picking_done`?

addons/udes_stock/models/stock_picking.py
```diff
@@ -601,6 +601,11 @@ def update_picking(
             # validate stock.picking
             picking.action_done()  # old do_transfer
 
+        self.env.ref('udes_stock.picking_update').with_context(
```
parisiale: Big method with 14 optional args... This is not the only path of execution :(
Are we sure we don't want to consider also the case where the method gets `quant_ids` (creates "extra stock.moves to the picking" and return `True`)?
On the other hand, for the above `if` clause where an error is thrown, I think it's ok not to trigger the update.

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,11 @@ def strategies(self, picking):
         return self.search([
             ('picking_type_id', '=', picking.picking_type_id.id),
         ])
+
+    @api.multi
+    def records(self, picking, report):
+        '''Return records to print for `report`.'''
+        picking.ensure_one()
+        if report.model == 'sale.order':
```
parisiale: Maybe pedantic.
This requires implicitly a dependency on odoo's `sale_stock` module, where `sale_line_id` is defined.
Wouldn't be cleaner to override this method in some customer module to add this `if` block?

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,11 @@ def strategies(self, picking):
         return self.search([
             ('picking_type_id', '=', picking.picking_type_id.id),
         ])
+
+    @api.multi
+    def records(self, picking, report):
```
parisiale: This changes the signature of the parent class and causes some test failures.

addons/udes_stock/data/picking_print.xml
richardwhitefoot: This is the desired behaviour, eg. on the update_picking call we are printing package labels and the picking_done call we could (not for HDS) be printing a summary of that entire picking.

addons/udes_stock/models/stock_picking_print_strategy.py
richardwhitefoot: That was my original thought, but udes_stock has a dependancy on sale_stock in it's manifest so this should not be an issue.

addons/udes_stock/models/stock_picking_print_strategy.py
richardwhitefoot: Fixed

addons/udes_stock/models/stock_picking.py
richardwhitefoot: Due to the nature of the update_picking method I agree it's okay not to trigger the update in this case; as you say it's a separate execution path.

parisiale:
👍

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,14 @@ def strategies(self, picking):
         return self.search([
             ('picking_type_id', '=', picking.picking_type_id.id),
         ])
+
+    @api.multi
+    def records(self, picking, report=None):
```
spenced: The `report` arg was added on the assumption that this would help a derived implementation of `records` select which records to print. The corresponding change in odoo-print isn't on master, which means that arg can be deleted.

Further, providing an explicit target could be made common in odoo-print. If a pair for, say, 'print_records' is passed in the context, then this could override any further selection.

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,18 @@ def strategies(self, picking):
         return self.search([
             ('picking_type_id', '=', picking.picking_type_id.id),
         ])
+
+    @api.multi
+    def records(self, picking, context=None):
```
spenced: Given the way you've written the change in odoo-print, `context` will always be a valid dict. You can safely remove the default value from the signature.

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,18 @@ def strategies(self, picking):
...
+        '''Return records to print for `report`.'''
+        picking.ensure_one()
+        print_records = None
+
+        if context is not None:
```
spenced: As per previous comment, this branch isn't required.

addons/udes_stock/models/stock_picking_print_strategy.py
```diff
@@ -23,3 +23,16 @@ def strategies(self, picking):
         return self.search([
             ('picking_type_id', '=', picking.picking_type_id.id),
         ])
+
+    @api.multi
+    def records(self, picking, context):
+        '''Return records to print for `report`.'''
+        picking.ensure_one()
+        print_records = None
```
spenced: This line can be deleted.

PR #92, opened 2018-11-14
[FIX] udes_stock: Do not propagate move qty change

ws-mps:
When a move is split, the qty of the original move is updated and
incorrectly propagates the qty update to the following move.
Use do_not_propagate context variable to avoid it.


parisiale:
👍

PR #94, opened 2018-11-15
[FIX] udes_stock: Redifine batch::check-user-batches

parisiale:


addons/udes_stock/static/src/js/logout.js
```diff
@@ -1,33 +1,33 @@
...
-            }
-        });
-    }
-    core.action_registry.add('logout', logout_override);
+  error = function (msg) {
+    Dialog.alert(self, _(msg), {title: _("Error")});
+  };
 
+  logout_override = function () {
+    test = ajax.jsonpRpc("/api/stock-picking-batch/check-user-batches/", 'call').then(function (data) {
```
richardwhitefoot: Setting 'test' and then not using it

addons/udes_stock/static/src/js/logout.js
```diff
@@ -1,33 +1,33 @@
...
+      if (data.error) {
+        error(data.error);
+      } else {
+        if (data != false) {
```
richardwhitefoot: Data can't be false at this point, if it passed the test above

PR #95, opened 2018-11-15
[IMP] udes_stock: Add confirm expiry date flag to picking type

ws-mps:
Used to determine if expiry dates should be requested to the user
for a given picking type.

### DIFF
--- a/addons/udes_product_expiry/__manifest__.py
+++ b/addons/udes_product_expiry/__manifest__.py
@@ -13,6 +13,8 @@
     ],
     'data': [
         'report/report_deliveryslip.xml',
+        'views/stock_move_line_views.xml',
+        'views/stock_picking_views.xml',
     ],
     'qweb': [
     ],
--- a/addons/udes_product_expiry/__manifest__.py
+++ b/addons/udes_product_expiry/__manifest__.py
@@ -1,0 +1,24 @@
+{
+    'name': 'UDES Product Expiry Functionality',
+    'version': '11.0',
+    'summary': 'Inventory, Logistics, Warehousing',
+    'description': "Holds specific functionality for product expiry for use with UDES",
+    'depends': [
+        'udes_stock',
+        'product_expiry'
+    ],
+    'category': 'Warehouse',
+    'sequence': 11,
+    'demo': [
+    ],
+    'data': [
+        'report/report_deliveryslip.xml',
+    ],
+    'qweb': [
+    ],
+    'test': [
+    ],
+    'installable': True,
+    'application': False,
+    'auto_install': False,
+}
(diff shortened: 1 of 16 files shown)


addons/udes_product_expiry/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    'name': 'UDES Product Expiry Functionality',
+    'version': '11.0',
+    'summary': 'Inventory, Logistics, Warehousing',
+    'description': "Holds specific functionality for product expiry for use with UDES",
+    'depends': [
+        'sale_stock',
+        'stock',
+        'stock_picking_batch',
+        'udes_stock',
+        'product_expiry'
```
ws-mps: Only depends on udes_stock and product_expiry.

addons/udes_product_expiry/models/stock_move_line.py
```diff
@@ -0,0 +1,44 @@
+from odoo import api, fields, models,  _
+from odoo.exceptions import ValidationError
+
+
+class StockMoveLine(models.Model):
+    _inherit = 'stock.move.line'
+
+    u_expiry_date = fields.Date('Expiry date')
+
+    def _action_done(self):
```
ws-mps: Is this function needed?

addons/udes_product_expiry/models/stock_move_line.py
```diff
@@ -0,0 +1,44 @@
...
+        return super(StockMoveLine, self)._action_done()
+
+    def _prepare_line_product_info(self, ml_values, products_info):
```
ws-mps: I think this changes should be in _update_products_info()

addons/udes_stock/models/stock_move_line.py
```diff
@@ -326,6 +326,13 @@ def _update_products_info(self, product, products_info, info):
         """
         tracking = product.tracking
         if tracking != 'none':
+
+            # Auto generate lot name if name is missing and required
+            if len(info['lot_names']) == 0 and self.mapped('picking_id.picking_type_id').u_confirm_tracking == 'no':
```
ws-mps: Create function.

addons/udes_product_expiry/models/stock_move_line.py
ws-mps: As per discussion, yes they should be in both.

addons/udes_product_expiry/models/stock_picking.py
```diff
@@ -0,0 +1,24 @@
...
+
+        self.validate_expiry_dates()
+        return super(StockPicking, self).action_done()
+
+    def validate_expiry_dates(self):
+        """ Validate expiry dates for all move lines in self based on picking and product config """
+
+        mls = self.mapped('move_line_ids')
+
+        if self.picking_type_id.u_confirm_tracking == 'no':
```
ws-mps: Group move lines by picking type and loop over different picking types. It shouldn't happen but makes the code more robust. Also self can contain more than one picking, hence self.picking_typed_id might generate an error.

ws-mps: This flag should be u_confirm_expiry_date.

PR #98, opened 2018-11-21
[ADD] udes_stock: batch: new states, assign, unassign

nocturnalastro:
Add more granular states to batch using pick state and user.
Add new endpoints for assigning and unassigning batches.
Add new picking type options for goodsout workflow.

User-story: 2163

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -211,6 +211,9 @@ class StockPicking(models.Model):
     def write(self, vals):
         """ If writing picking, check if previous batch is now complete
         """
+        # This will be used to trigger recompute of the batch state
+        # we can't relate on the state after as batch_id might be
+        # removed in the write
         batches = self.mapped(lambda p: p.batch_id)
         context_vals = {'orig_batches': batches} if batches else {}
         return super(StockPicking, self.with_context(**context_vals)) \
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -193,7 +193,7 @@ class StockPicking(models.Model):
         mls = self.mapped('move_line_ids')
         res = super(StockPicking, self).action_done()
         picks = mls.mapped('picking_id')
-        picks.with_context(orig_batches=False)._trigger_batch_state_recompute()
+        self._trigger_batch_state_recompute(picks=picks)
 
         self.env.ref('udes_stock.picking_done').with_context(
             active_model=picks._name,
@@ -1481,14 +1481,16 @@ class StockPicking(models.Model):
         return super(StockPicking, self)._put_in_pack()
 
     @api.constrains('state', 'batch_id')
-    def _trigger_batch_state_recompute(self):
+    def _trigger_batch_state_recompute(self, picks=None):
         """Batch state is dependant on picking state and batch_id"""
-        batch = self.env.context.get('orig_batches')
 
-        if batch is False:
+        if picks is not None:
             # We only want to run this when told
             # not when there isnt a batch
-            batch = self.mapped('batch_id')
+            batch = picks.mapped('batch_id')
+        else:
+            batch = self.env.context.get('orig_batches')
+
 
         if batch:
             batch._compute_state()
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -188,7 +188,8 @@ class StockPicking(models.Model):
         """
         self.assert_not_pending()
         mls = self.mapped('move_line_ids')
-        res = super(StockPicking, self).action_done()
+        res = super(StockPicking,
+                    self.with_context(lock_batch_state=True)).action_done()
         picks = mls.mapped('picking_id')
         self._trigger_batch_state_recompute(picks=picks)
 
@@ -1496,3 +1497,28 @@ class StockPicking(models.Model):
         if batch:
             batch._compute_state()
         return True
+
+    def raise_stock_inv(self, reason, quants, location):
+        """Unreserve stock create stock investigation for reserve_stock and
+           attempt to reserve new stock
+        """
+        Picking = self.env['stock.picking']
+        Group = self.env['procurement.group']
+        wh = self.env.user.get_user_warehouse()
+        stock_inv_pick_type = wh.u_stock_investigation_picking_type or \
+            wh.int_type_id
+
+        self._refine_picking(reason)
+        group = Group.get_group(group_identifier=reason, create=True)
+
+        # create new "investigation pick"
+        Picking.create_picking(
+            quant_ids=quants.ids,
+            location_id=location.id,
+            picking_type_id=stock_inv_pick_type.id,
+            group_id=group.id,
+        )
+
+        # Try to re-assign the picking after, by creating the
+        # investigation, we've reserved the problematic stock
+        self.action_assign()
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -848,10 +848,12 @@ class StockPicking(models.Model):
         if PickingBatch.get_user_batches():
             raise ValidationError(_('User %s already has an in progress batch') % user.name)
 
-        if self.batch_id is False:
-            batch = PickingBatch.create({'user_id': user.id})
+        if not self.batch_id:
+            batch = PickingBatch.create({
+                'user_id': user.id,
+                'u_ephemeral': True,
+            })
             self.batch_id = batch.id
-            batch.write({'u_ephemeral': True})
             batch.confirm_picking()
 
     def _get_package_search_domain(self, package):
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -193,7 +193,9 @@ class StockPicking(models.Model):
         res = super(StockPicking,
                     self.with_context(lock_batch_state=True)).action_done()
         picks = mls.mapped('picking_id')
-        batches += picks.mapped('batch_id')
+        # batches of the following stage should also be recomputed
+        picks |= mls.mapped('picking_id.u_next_picking_ids')
+        batches |= picks.mapped('batch_id')
         self._trigger_batch_state_recompute(batches=batches)
 
         self.env.ref('udes_stock.picking_done').with_context(
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -53,13 +53,11 @@ class StockPicking(models.Model):
         compute='_compute_related_picking_ids',
         help='Next pickings',
     )
-
     u_first_picking_ids = fields.One2many(
         'stock.picking', string='First Pickings',
         compute='_compute_first_picking_ids',
         help='First pickings in the chain'
     )
-
     u_created_back_orders = fields.One2many(
         'stock.picking', string='Created Back Orders',
         compute='_compute_related_picking_ids',
@@ -93,8 +91,7 @@ class StockPicking(models.Model):
 
     u_has_discrepancies = fields.Boolean(
         string='Has discrepancies', compute='_compute_picking_quantities',
-        readonly=True, help='Flag to indicate if the picking still has discrepancies.',
-        )
+        readonly=True, help='Flag to indicate if the picking still has discrepancies.')
 
     u_num_pallets = fields.Integer(
         'Total Pallets count', compute='_compute_picking_packages', store=False,
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1483,14 +1483,13 @@ class StockPicking(models.Model):
     @api.constrains('state', 'batch_id')
     def _trigger_batch_state_recompute(self, picks=None):
         """Batch state is dependant on picking state and batch_id"""
-
         if picks is not None:
             # We only want to run this when told
             # not when there isnt a batch
             batch = picks.mapped('batch_id')
         else:
-            batch = self.env.context.get('orig_batches')
-
+            batch = self.env.context.get('orig_batches') or \
+                self.mapped('batch_id')
 
         if batch:
             batch._compute_state()
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -189,10 +189,12 @@ class StockPicking(models.Model):
         self.assert_not_pending()
         mls = self.mapped('move_line_ids')
         # Prevent recomputing the batch stat
+        batches = mls.mapped('picking_id.batch_id')
         res = super(StockPicking,
                     self.with_context(lock_batch_state=True)).action_done()
         picks = mls.mapped('picking_id')
-        self._trigger_batch_state_recompute(picks=picks)
+        batches += picks.mapped('batch_id')
+        self._trigger_batch_state_recompute(batches=batches)
 
         self.env.ref('udes_stock.picking_done').with_context(
             active_model=picks._name,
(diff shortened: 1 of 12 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,35 @@ class StockPickingBatch(models.Model):
...
+            ('waiting', 'Waiting'),
+            ('ready', 'Ready'),
+        ],
+        compute='_compute_state',
+        store=True,
+    )
+
+    @api.multi
+    def confirm_picking(self):
+        """Over write method to use our waiting state and recompute state
```
unirho: overwrite

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +84,87 @@ def _compute_priority(self):
...
+    @api.constrains('user_id')
+    def _compute_state(self):
+        """ Compute the state of a batch post confirm
+            waiting    : Aleast some picks are not ready
+            ready      : All picks are in ready state (assigned)
+            in_progess : All picks are ready and a user has been assigned
+            done       : All picks are complete (in state done or cancel)
+
+            the other two states are draft and cancel are manual
+            to tranistions from or to the respective state.
```
unirho: transitions

parisiale:
Don't forget to squash the `wip` commit :P

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +84,87 @@ def _compute_priority(self):
...
+        """
+        Picking = self.env['stock.picking']
+
+        for batch in self:
+
+            if batch.state in ['draft', 'done', 'cancel']:
+                # Can't do anything with them don't bother trying.
+                continue
+
+            if not batch.picking_ids:
```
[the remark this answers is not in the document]
parisiale: Forget about above... For some reason I couldn't read the `continue` check above.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +84,87 @@ def _compute_priority(self):
...
+                if pick.state == 'assigned':
+                    ready_picks |= pick
+
+                elif pick.state in ['done', 'cancel']:
+                    complete_picks |= pick
+
+                else:
+                    other_picks |= pick
+
+            # Based on groups compute batch state
```
parisiale: As discussed, I suggest to be explicit about the current state, transitions and assumptions (i.e. dispatch the computation based on the current state).

addons/udes_stock/models/stock_picking.py
```diff
@@ -205,23 +206,15 @@ def action_cancel(self):
         Check if picking batch is now complete
         """
         res = super(StockPicking, self).action_cancel()
-        self.check_batch()
         return res
 
-    @api.one
-    def check_batch(self):
```
parisiale: Definitely worth a taco.

addons/udes_stock/models/stock_picking.py
```diff
@@ -205,23 +206,15 @@ def action_cancel(self):
...
-            self.batch_id.check_batches()
-
     def write(self, vals):
         """ If writing picking, check if previous batch is now complete
         """
         batches = self.mapped(lambda p: p.batch_id)
-        res = super(StockPicking, self).write(vals)
-        batches.check_batches()
-        return res
+        context_vals = {'orig_batches': batches} if batches else {}
```
parisiale: I think a comment about `recompute` would be nice, as I think you're actually sending a message there, right?

addons/udes_stock/models/stock_picking.py
```diff
@@ -837,11 +831,11 @@ def batch_to_user(self, user):
         if PickingBatch.get_user_batches():
             raise ValidationError(_('User %s already has an in progress batch') % user.name)
 
-        if self.batch_id.id is False:
```
parisiale: Hmmm, looks like the the check did never fail otherwise it would have broke, right?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,35 @@ class StockPickingBatch(models.Model):
...
+
+    @api.multi
+    def confirm_picking(self):
+        """Over write method to use our waiting state and recompute state
+           incase this should be ready or in_progress
+        """
+        pickings_todo = self.mapped('picking_ids')
+        self.write({'state': 'waiting'})  # Get it out of draft
+        res = pickings_todo.action_assign()
+        self._compute_state()
```
parisiale: And this will trigger the automatic transition to `ready`, in case all pickings are `ready`, right?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,35 @@ class StockPickingBatch(models.Model):
...
+        store=True,
+    )
+
+    @api.multi
+    def confirm_picking(self):
+        """Over write method to use our waiting state and recompute state
+           incase this should be ready or in_progress
+        """
+        pickings_todo = self.mapped('picking_ids')
+        self.write({'state': 'waiting'})  # Get it out of draft
```
parisiale: Do we want to enforce that the we're doing a `draft` --> `waiting` transition? 

If so, is such condition expected (so we can assert here - i.e. it's up to the caller to ensure that)? Or do we want to raise an invalid state error?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,35 @@ class StockPickingBatch(models.Model):
...
+        res = pickings_todo.action_assign()
```
parisiale: This may throw.
Wouldn't be better to set the state after it?

addons/udes_stock/models/stock_picking_batch.py
nocturnalastro: The `draft` -> `waiting` transition is a forced one its just them saying I want to do this piece of work.
And setting it as `waiting` is purely a lets get started on our automatic transitions path kinda thing.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +84,159 @@ def _compute_priority(self):
...
+            elif batch.state == 'ready':
+                batch._from_ready_transitions(other_picks)
+
+            elif batch.state == 'in_progress':
+                batch._from_in_progress_transitions(ready_picks, other_picks)
+
+            else:
+                # How did we get here?
+                # Let do nothing to be safe!
+                continue
```
parisiale: The only possibility is having an unexpected state; I suggest to log a warning.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,38 @@ class StockPickingBatch(models.Model):
...
+            )
+
+        pickings_todo = self.mapped('picking_ids')
+        self.write({'state': 'waiting'})  # Get it out of draft
+        try:
+            res = pickings_todo.action_assign()
+        except:
+            # On an error rollback to draft
+            self.write({'state': 'draft'})
+            # Then reraise the error
```
parisiale: Style: the code is clear, I suggest to remove both comments; too much noise otherwise.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -29,6 +29,38 @@ class StockPickingBatch(models.Model):
...
+            raise
+        return res
```
parisiale: Style: I would return directly in the `try` block; referencing a block var is a code smell as it's error prone.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+
+            elif batch.state == 'in_progress':
+                batch._from_in_progress_transitions(ready_picks, other_picks)
+
+            else:
+                # How did we get here?
+                # Let do nothing to be safe!
+                continue
+
+    def _from_in_progress_transitions(self, ready_picks=None, other_picks=None):
```
parisiale: Are the optional args meant for dependency injection during testing? If not, I suggest to make them not optional as this method is only invoked once.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+        """ Transition from in_progress to another state"""
+        self.ensure_one()
+
+        # Make sure we have the pick groups
+        if ready_picks is None or other_picks is None:
+            ready_picks, other_picks = \
+                self._calculate_pick_groups(self.picking_ids)
```
parisiale: The caller computes `ready_picks` and `other_picks` by using the `picking_ids` of a single batch out of the recordset instance. Here you're using the entire recordset. Is this correct?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+
+        # This transitions should be handled already in _compute_state
+        # but as a guard against it being called else where by someone
+        # lets add it in here as well
+        if not ready_picks and not other_picks:
+            # All picks are complete so we're done
+            self.state = 'done'
+            return True
+
```
parisiale: Style pedantry: as we have one if block above separated by this by an empty line, the structure is a little confusing. I would suggest to remove all empty lines between `if` and related `elif`s.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+        elif ready_picks and not other_picks and not self.user_id:
+            # User is removed lets go back to being ready
+            self.state = 'ready'
+            return True
+
+        # None really a transition but a bit of batch pick managemnet
+        # which has to be delt with
```
parisiale: The comment should go inside the `elif` block. Please remove the above empty line, otherwise it's hard to identify the blocks and keep `if` & `elif`s associated.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+            ready_picks, other_picks = \
+                self._calculate_pick_groups(batch.picking_ids)
+
+            if not batch.picking_ids or not (ready_picks or other_picks):
+                # This accounts for all pickings being complete, canceled or
+                # all pickings get removed for some reason.
+                # The last two can happen in any state
+                batch.state = 'done'
+
+            # Attempt transitions based on state
```
parisiale: I see the point of this comment, but it just ruins the `if` / `elif` indentation. Please remove the empty lines; you can't tell whether they go together at a glance.

addons/udes_stock/models/stock_picking_batch.py
parisiale: Could you move the method after the `ready` one, so that their order matches the status flow?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+            return True
+
+        # None really a transition but a bit of batch pick managemnet
+        # which has to be delt with
+        elif other_picks:
+            # We shouldn't have these check if we can make them ready
+            # otherwise remove them
+            return self._remove_unready_picks()
+
+    def _from_waiting_transitions(self, other_picks=None):
```
parisiale: Same as above: not optional if possible.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+
+        # Make sure we have the pick groups
+        if ready_picks is None or other_picks is None:
+            ready_picks, other_picks = \
+                self._calculate_pick_groups(self.picking_ids)
+
+        # This transitions should be handled already in _compute_state
+        # but as a guard against it being called else where by someone
+        # lets add it in here as well
+        if not ready_picks and not other_picks:
```
parisiale: I'd love to get rid of the optional args, the above assignment, and this (surely duplicate) check.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+            self.state = 'in_progress'
+        else:
+            self.state = 'ready'
+        return True
+
+    def _from_ready_transitions(self, other_picks=None):
+        """ Transition from ready to another state"""
+        self.ensure_one()
+
+        # Make sure we have the pick groups
```
parisiale: Same...

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+            elif pick.state in ['done', 'cancel']:
+                continue
+
+            else:
+                other_picks |= pick
+
+        return ready_picks, other_picks
+
+    @api.multi
+    def _remove_unready_picks(self, picks=None):
```
parisiale: In general, optional args are confusing as they imply a default which may not be clear to the user.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -231,51 +416,24 @@ def _create_batch(self, user_id, picking_type_id, picking_priorities=None):
...
-        is incomplete.
-        """
-        Picking = self.env['stock.picking']
-        PickingBatch = self.env['stock.picking.batch']
-
-        if not batches:
-            batches = PickingBatch.search([('user_id', '=', user_id),
-                                           ('state', '=', 'in_progress')])
-
+    def _check_user_batch_in_progress(self, user_id=None):
```
parisiale: Sorry to repeat, but the optional arg requires reading `get_user_batches` to understand what's going on. I suggest to only use optional args when necessary.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -308,11 +466,12 @@ def drop_off_picked(self, continue_batch, location_barcode):
             for pick in pickings:
                 pick_todo = pick
                 pick_mls = completed_move_lines.filtered(lambda x: x.picking_id == pick)
+
                 if pick._requires_backorder(pick_mls):
                     pick_todo = pick._backorder_movelines(pick_mls)
-                # at this point pick_todo should contain only mls done
-                pick_todo.update_picking(validate=True)
 
+                # at this point pick_todo should contain only mls done
```
parisiale: Is this a fix or a required behaviour change?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -308,11 +466,12 @@ def drop_off_picked(self, continue_batch, location_barcode):
             for pick in pickings:
                 pick_todo = pick
```
parisiale: Not needed anymore ifchanges below are correct.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -308,11 +466,12 @@ def drop_off_picked(self, continue_batch, location_barcode):
...
                 pick_mls = completed_move_lines.filtered(lambda x: x.picking_id == pick)
+
                 if pick._requires_backorder(pick_mls):
                     pick_todo = pick._backorder_movelines(pick_mls)
```
parisiale: Perhaps `_pick_todo` to indicate that we don't need the return value but are only interested in the side effect (backorder creation). BTW, why `_backorder_movelines` with the underscore if we're calling it from `StockPickingBatch`?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+            ['draft', 'waiting', 'confirmed']
+        not_ready = picks.filtered(not_ready_lam)
+
+        if not_ready:
+            confirmed_not_ready = not_ready.filtered(
+                lambda pick: pick.state == 'confirmed')
+            if confirmed_not_ready:
+                confirmed_not_ready.action_assign()
+
+            not_ready.filtered(not_ready_lam).write({'batch_id': False})
```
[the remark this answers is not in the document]
parisiale: That is needed as the change of state may impact the filtering... Perhaps a comment.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,159 @@ def _compute_priority(self):
...
+        # first lets double check there not ready
+        not_ready_lam = lambda pick: pick.state in \
+            ['draft', 'waiting', 'confirmed']
+        not_ready = picks.filtered(not_ready_lam)
+
+        if not_ready:
+            confirmed_not_ready = not_ready.filtered(
+                lambda pick: pick.state == 'confirmed')
+            if confirmed_not_ready:
+                confirmed_not_ready.action_assign()
```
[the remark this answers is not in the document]
parisiale: Comment?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -55,6 +87,162 @@ def _compute_priority(self):
...
+                ready    : pick.state is assigned
+                other    : pick.state is in any other state than the ones above
+        """
+        Picking = self.env['stock.picking']
+
+        ready_picks = Picking.browse()
+        other_picks = Picking.browse()
+
+        for pick in picks:
+
```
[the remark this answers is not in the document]
parisiale: We'll do, thanks.

parisiale:
💓

PR #99, opened 2018-11-22
[IMP] udes_stock: Draft pickings displayed in Kanban view

richardwhitefoot:


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -289,3 +289,6 @@ def get_picking_type(self, picking_type_id):
                     _('Cannot find picking type with id %s') %
                     picking_type_id)
         return picking_type
+
+    def get_action_picking_tree_draft(self):
+        return self._get_action('udes_stock.action_picking_tree_draft')
```
unirho: I believe it's convention to put an empty newline at the end =)

PR #102, opened 2018-11-28
story/2163 Lock batch state when required

nocturnalastro:


addons/udes_stock/models/stock_picking.py
```diff
@@ -1482,3 +1483,28 @@ def _trigger_batch_state_recompute(self, picks=None):
         if batch:
             batch._compute_state()
         return True
+
+    def raise_stock_inv(self, reason, quants, location):
```
parisiale: Simple factoring out, way clearer... 👍

PR #108, opened 2018-12-07
udes_stock: new drop-off workflow

parisiale:


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -453,11 +453,112 @@ def drop_off_picked(self, continue_batch, location_barcode):
...
+        all_mls_to_drop = self._get_move_lines_to_drop_off()
+
+        if not len(all_mls_to_drop):
+            return {'last': True,
+                    'move_line_ids': [],
+                    'summary': ""}
+
+        picking_type = self.picking_ids[0].picking_type_id
+
+        if not all([p.picking_type_id == picking_type for p in self.picking_id]):
```
ws-mps: Batch has a computed field picking_type_ids, which gets the value from its pickings, so we can use `len(self.picking_type_ids) > 1`

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -453,11 +453,112 @@ def drop_off_picked(self, continue_batch, location_barcode):
...
+         - unknown or invalid (e.g. 'all') drop off criterion.
+        """
+        self.ensure_one()
+        assert self.state == 'in_progress', \
+               "Batch must be in progress to be dropped off"
+
+        all_mls_to_drop = self._get_move_lines_to_drop_off()
+
+        if not len(all_mls_to_drop):
+            return {'last': True,
```
ws-mps: Shouldn't this be an error? Does not make sense to have item_identity and no move lines to drop off.

parisiale: That's what the API says;l we tried to make the API robust - you never know what can happen to the system once a request is issued, so, if that happens, you have a specified behaviour.

PR #110, opened 2018-12-18
[FIX] udes_stock: Properly compute batch state at remove_unfinished_work

ws-mps:


parisiale:
👍 	
Added to #110 - will close this.

PR #116, opened 2018-12-20
[IMP] udes_stock: Keep more fields when refactoring moves/move lines

ws-mps:
Created new function to prepare all the extra fields needed for the
new pickings when at _new_picking_for_group().
The new function now returns date_done and partner_id in addition
to origin.

User-story: 2176

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -361,6 +361,28 @@ class StockMove(models.Model):
 
         return self.groupby(lambda ml: ml.u_grouping_key)
 
+    def _prepare_extra_info_for_new_picking_for_group(self, pickings, moves):
+        """ Given the group of moves to refactor and its related pickings,
+            prepare the extra info for the new pickings that might be created
+            for the group of moves.
+            Fields with more than one value are going to be ignored.
+        """
+        values = {}
+
+        origins = list(set(pickings.mapped('origin')))
+        if len(origins) == 1:
+            values['origin'] = origins[0]
+
+        partners = pickings.mapped('partner_id')
+        if len(partners) == 1:
+            values['partner_id'] = partners.id
+
+        dates_done = list(set(moves.mapped('date')))
+        if len(dates_done) == 1:
+            values['date_done'] = dates_done[0]
+
+        return values
+
     def refactor_action_group_by_move_key(self):
         """
         group the moves by the splitting criteria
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -361,6 +361,24 @@ def group_by_key(self):
...
+        if len(origins) == 1:
+            values['origin'] = origins[0]
+
+        partners = pickings.mapped('partner_id')
+        if len(partners) == 1:
+            values['partner_id'] = partners.id
+
+        dates_done = list(set(moves.mapped('date')))
+        if len(dates_done) == 1:
+            values['date_done'] = dates_done[0]
```
parisiale: The above can be factored out as you're repeating the same logic 3x.

addons/udes_stock/models/stock_move.py
```diff
@@ -361,6 +361,24 @@ def group_by_key(self):
 
         return self.groupby(lambda ml: ml.u_grouping_key)
 
+    def _prepare_extra_values_for_group(self, pickings, moves):
+
```
parisiale: One empty line before and one empty line after the signature just scatter my eyes.

addons/udes_stock/models/stock_move.py
```diff
@@ -361,6 +361,24 @@ def group_by_key(self):
 
         return self.groupby(lambda ml: ml.u_grouping_key)
 
+    def _prepare_extra_values_for_group(self, pickings, moves):
```
parisiale: You'll actually get the extra values only if the collections below will have a single element. Is this safe enough for the caller? Also, the method name doesn't say anything about what's going to happen... You gotta read the code.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1294,17 +1294,22 @@ def _new_picking_for_group(self, group_key, moves, origin=None):
...
+
         else:
-            # avoid misleading source document
-            if origin != picking.origin:
-                picking.origin = False
+            # avoid misleading values for extra fields
+            for field, value in kwargs.items():
+                current_value = getattr(picking, field, None)
+                if current_value and current_value != value:
+                    setattr(picking, field, False)
```
parisiale: The above is not very clear to me... The extra values are used to create a picking, if none or multiple ones are found, otherwise are used to unset the respective fields of the instance. I can't understand why by simply looking at the workflow. Would a docstring help perhaps?

addons/udes_stock/models/stock_move.py
ws-mps: It's not exactly the same logic, depends on the field type. the middle one uses.id and the other ones list(set()) and [0].
I will leave it like this, the next thing we add we factor, but I don't think we need anything else.

addons/udes_stock/models/stock_move.py
ws-mps: Thanks. Will add docstring, to explain.

addons/udes_stock/models/stock_move.py
ws-mps: I agree, this is because the docstring is missing.

addons/udes_stock/models/stock_picking.py
ws-mps: Yes, I will explain.

addons/udes_stock/models/stock_move.py
```diff
@@ -361,6 +361,24 @@ def group_by_key(self):
...
+
+        values = {}
+
+        origins = list(set(pickings.mapped('origin')))
```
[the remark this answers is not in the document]
ws-mps: Yes. Didn't wanted to add complexity to the code, just for 3 fields.

PR #118, opened 2018-12-24
[IMP] udes_stock: Add handle unpickable items flag to picking type

ws-mps:
When this flag is set to false, a message asking to contact team leader
will be raised instead of performing the usual unpickable item behaviour.

User-story: 2551

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -618,6 +618,18 @@ class StockPickingBatch(models.Model):
         return all([pick.is_valid_location_dest_id(location=location)
                     for pick in all_done_pickings])
 
+    def _check_unpickable_item(self):
+        """ Checks if all the picking types of the batch are allowed to handle
+            unpickable items. If one of them does not, it raises an error.
+        """
+        if not all(self.picking_type_ids.mapped('u_enable_unpickable_items')):
+            raise ValidationError(
+                _('This type of operation cannot handle unpickable items. '
+                  'Please, contact your team leader to resolve the issue. '
+                  'Press back when resolved.')
+            )
+
+
     def unpickable_item(self,
                         reason,
                         product_id=None,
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -618,6 +618,17 @@ def is_valid_location_dest_id(self, location_ref):
...
                     for pick in all_done_pickings])
 
+    def _check_unpickable_item(self):
+        """ Checks if the picking type of the batch is allowed to handle
+            unpickable items.
+            If it is not, it raises an error asking to contact team leader.
+        """
+        if not all(self.picking_type_ids.mapped('u_handle_unpickable_items')):
+            raise ValidationError(
+                _('Contact your team leader.')
```
parisiale: Exception messages should describe errors and failures. Then it should be up to the caller to catch them and decide what to do. The above message is just weird.

parisiale:
Looks good.
Just nitpick comments, no strong opinion.
👍

addons/udes_stock/models/stock_picking_batch.py
unirho: it's part of the AC's that they want this message by the looks of things =s

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -618,6 +618,17 @@ def is_valid_location_dest_id(self, location_ref):
         return all([pick.is_valid_location_dest_id(location=location)
                     for pick in all_done_pickings])
 
+    def _check_unpickable_item(self):
+        """ Checks if the picking type of the batch is allowed to handle
+            unpickable items.
+            If it is not, it raises an error asking to contact team leader.
+        """
+        if not all(self.picking_type_ids.mapped('u_handle_unpickable_items')):
```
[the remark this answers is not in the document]
parisiale: Ofc, you gotta invert the condition, but it's always good to get rid of the `not` in a clause.

addons/udes_stock/models/stock_picking_batch.py
ws-mps: I agree the message is a bit weird, I have changed it to something more meaningful and waiting for PO confirmation.

addons/udes_stock/models/stock_picking_batch.py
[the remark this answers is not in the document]
ws-mps: I was going to change it but having to invert the condition makes it more confusing. I have left the original statement since it matches to what we are semantically doing:
If not all picking types can handle unpickable items: bla bla bla.

PR #120, opened 2019-01-04
[IMP] udes_sale_stock: Restrict Outbound and Stock user access levels

richardwhitefoot:

### DIFF
--- a/addons/udes_sale_stock/models/partner_pii.py
+++ b/addons/udes_sale_stock/models/partner_pii.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+
+# @TODO: Tidy, forms field read-only
+
+from odoo import models, fields, api, SUPERUSER_ID, _
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+REDACTED_TXT = _('Redacted')
+PII_RELATIONS = [
+    'partner_id', 'res_partner_id', 'author_id', 'partner_address_id',
+    'owner_id', 'partner_shipping_id', 'partner_invoice_id',
+    'order_partner_id', 'customer_id'
+]
+
+
+@api.model
+def _read(self, result):
+    """Redact fields based on res.user.u_view_customers"""
+    if self.env.uid == SUPERUSER_ID or self.env.user.u_view_customers:
+        return result
+
+    for record in result:
+        for field in record:
+            if (field in PII_RELATIONS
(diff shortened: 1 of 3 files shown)


addons/udes_sale_stock/models/partner_pii.py
```diff
@@ -0,0 +1,74 @@
...
+
+        result = super(ResPartner, self).name_get()
+
+        if self.env.uid == SUPERUSER_ID or self.env.user.u_view_customers:
+            return result
+
+        for i, record in enumerate(result):
+            if isinstance(record, tuple) and \
+                self.sudo().search_count([('id', '=', record[0]),('customer', '=', True)]):
+                result[i] = (record[0], REDACTED)
```
parisiale: Not a big fan of modifying a collection while iterating it. Could you create a new one to be used as the return value? You should be using references anyway, so it should be ok.

addons/udes_sale_stock/models/partner_pii.py
```diff
@@ -0,0 +1,74 @@
...
+def _read(self, result):
+    """Redact fields based on res.user.u_view_customers"""
+
+    if self.env.uid == SUPERUSER_ID or self.env.user.u_view_customers:
+        return result
+
+    for record in result:          
+        for field in record:
+            if field in PII_RELATIONS and isinstance(record[field], tuple) and \
+                self.env['res.partner'].sudo().search_count([('id', '=', record[field][0]),('customer', '=', True)]):
```
parisiale: Too long. Besides, this bit of the if clause is aligned with the actual block which makes it confusing.

parisiale: Also, need space between tuples.

addons/udes_sale_stock/models/partner_pii.py
```diff
@@ -0,0 +1,74 @@
...
+                record[field] = (record[field][0], REDACTED)
+                _logger.info("Redacted: %s.%s", self._name, field)
```
parisiale: Do we need the l10n helper (the `_` underscore function)?

richardwhitefoot: Added in rather than constant

addons/udes_sale_stock/models/partner_pii.py
richardwhitefoot: OC, will change it

PR #125, opened 2019-01-09
[IMP] (2521) udes_stock: new move_line::suggested_locations response

parisiale:


addons/udes_stock/controllers/stock_move_line.py
```diff
@@ -19,13 +19,37 @@ def suggested_locations(self, move_line_ids):
...
-            locations = pick_loc if locations is None \
-                        else locations & pick_loc
+            pick_locs = pick.get_suggested_locations(pick_mls)
+            locations = pick_locs if locations is None \
+                        else locations & pick_locs
 
-        return locations.get_info()
+            if pick.picking_type_id.u_drop_location_constraint == "enforce_with_empty":
+
+                # TODO(ale): do we want the `enforce_with_empty` check in the policies???
```
parisiale: @ws-mps 	^^

addons/udes_stock/models/stock_move_line.py
```diff
@@ -808,8 +808,8 @@ def _validate_location_dest(self, location=None):
                       (location.name, picking.location_dest_id.name))
                 )
 
-            if picking.picking_type_id.u_drop_location_constraint == 'enforce':
-
+            if picking.picking_type_id.u_drop_location_constraint \
+                    in ['enforce', 'enforce_with_empty']:
```
ws-mps: I think in line 825 we shouldn't raise the error if the location is an empty location and constraint is `enforce_with_empty`

ws-mps: Forget about the example, that will accept any empty location, was trying to avoid to call get_empty_location(), but I don't think we can.
```
if constraint == enforce_with_empty and not location.quant_ids:
    locations += get_empty_locations()
if locations and location not in locations:
...    
```

parisiale: Yep, you're right, as the system is allowing to use empty locations if the constraint is `enforce_with_empty`, it shall not raise an error. 

I'm not sure about the `not location.quant_ids` clause because 1) the `_validate_location_dest` call can happen before `write` is completed (synchronously by overriding `write` - see above method) and you may not have the quant update by then (we're looking for that, right?) and 2) what we're trying to do here is to make the system more tolerant and allow using empty locations for whatever reason. I may be missing something; what's the reason for such clause?

ws-mps: You are right, was a copy paste from the previous example. Will update the example.

addons/udes_stock/README.md
```diff
@@ -721,13 +722,24 @@ that is an aray of numbers, containing the IDs of the move lines being dropped.
...
-  "result" : [
-    {"id": 1, "name": "Location 1", "barcode": "L00000100"},
-    {"id": 2, "name": "Location 2", "barcode": "L00000200"}]
+    "result" : [
+        { "title": "",
+          "locations": [
+            {"id": 1, "name": "Location 1", "barcode": "L00000100"},
+            {"id": 2, "name": "Location 2", "barcode": "L00000200"}]
+        },
+        { "title": "empty locations",
```
samueljsb:
```suggestion
        { "title": "empty",
```
Would this fix the UI problem ("the following empty locations location:")?

addons/udes_stock/controllers/stock_move_line.py
```diff
@@ -19,13 +19,32 @@ def suggested_locations(self, move_line_ids):
...
+        if locations:
+            response.append({"title": "",
+                             "locations": locations.get_info()})
+
+            if empty_locations:
+                # Ensure we don't have duplicates in the two recordsets
+                empty_locations = empty_locations - locations
+
+        if empty_locations:
+            response.append({"title": "empty locations",
```
samueljsb:
```suggestion
            response.append({"title": "empty",
```
Same as above

PR #130, opened 2019-01-24
[FIX] udes_stock: Unlink empty pickings after refactoring

joe-gb:
To ensure that any system generated canceled pickings are deleted
so that the system does not pollute the database with empty data.


addons/udes_stock/tests/test_splitting.py
```diff
@@ -337,6 +337,8 @@ def test02_maintain_single_pick_extra_info(self):
         self.picking.write({'origin': "Test origin 123",
                             'partner_id': partner.id})
 
+        # Take a copy of the original picking, as refactoring can delete records
+        original_picking = self.picking.copy()
```
ws-mps: What about just storing partner_id and origin? No need to copy the whole picking. (this applies to the other test)

PR #140, opened 2019-02-08
[FIX] udes_stock: Do not attempt to cancel empty pickings

mcb30:
Pick factorisation may result in some empty pickings.  The current
code attempts to cancel these by directly writing the "state" field to
the value "cancel", with a comment noting that action_cancel() does
not work on a picking with no associated stock moves.

The reason that action_cancel() does not work on a picking with no
associated stock moves is that the "state" field is a readonly field
computed from the associated stock move lines.  A picking with no
associated stock move lines is, by definition, in state "draft".

The current code seems to succeed in setting the state to "cancel" due
only to the permissiveness of the ORM, which (for performance reasons)
generally trusts code not to attempt undefined operations such as
writing to a non-invertible but stored computed field.

Fix by deleting any empty pickings, since there is no valid way for a
picking with no moves to exist in a cancelled state.

Deletions are performed via sudo() since most users will not have
permission to delete pickings.

Signed-off-by: Michael Brown <mbrown@fensystems.co.uk>

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -424,14 +424,10 @@ class StockMove(models.Model):
 
             Picking._new_picking_for_group(key, move_group, **values)
 
+        # Deactivate any empty picks
         empty_picks = pickings.filtered(lambda p: len(p.move_lines) == 0)
         if empty_picks:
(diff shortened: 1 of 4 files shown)


mcb30:
Obsoleted by PR #142

PR #148, opened 2019-02-16
[ADD] udes_sale_stock: Add flag in picking view for privacy wrapping

nocturnalastro:
User-stoy 1946

### DIFF
--- a/addons/udes_sale_stock/models/stock_move.py
+++ b/addons/udes_sale_stock/models/stock_move.py
@@ -9,6 +9,11 @@ class StockMove(models.Model):
 
     _inherit = 'stock.move'
 
+    u_is_privacy_wrapped = fields.Boolean(
+        string='Is privacy wrapped',
+        compute='_compute_privacy_wrapped',
+    )
+
     def _action_done(self):
         result = super(StockMove, self)._action_done()
         result.mapped('sale_line_id.order_id').check_delivered()
@@ -23,3 +28,9 @@ class StockMove(models.Model):
             self.mapped('sale_line_id.order_id').check_delivered()
 
         return result
+
+    def _compute_privacy_wrapped(self):
+        privacy = self.env.ref('udes_stock.privacy_wrapping')
+        for mv in self:
+            mv.u_is_privacy_wrapped = privacy in mv.mapped(
+                'sale_line_id.product_packaging')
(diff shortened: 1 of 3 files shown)


addons/udes_sale_stock/models/stock_move.py
```diff
@@ -23,3 +28,9 @@ def _action_cancel(self):
             self.mapped('sale_line_id.order_id').check_delivered()
 
         return result
+
+    def _compute_privacy_wrapped(self):
+        privacy = self.env.ref('udes_stock.privacy_wrapping')
+        for mv in self:
+            mv.u_is_privacy_wrapped = privacy in mv.mapped(
```
ws-mps: One move only links to one sale_line_id, so no need for mapped.

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -9,6 +9,11 @@ class StockMove(models.Model):
 
     _inherit = 'stock.move'
 
+    u_is_privacy_wrapped = fields.Boolean(
+        string='Is privacy wrapped',
+        compute='_compute_privacy_wrapped',
```
ws-mps: Another option is to use a related field, with readonly=True, related='sale_line_id.privacy_wrapping' ? Although it is not going to be a boolean field.

addons/udes_sale_stock/views/stock_picking_views.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_form_stock_picking" model="ir.ui.view">
+        <field name="name">udes_sale_picking_form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='move_lines']/tree/field[@name='product_id']" position="after">
+                <field name="u_is_privacy_wrapped" />
```
ws-mps: readonly ?

addons/udes_sale_stock/models/stock_move.py
nocturnalastro: I like that better, I'll ask what would be preferred tomorrow. :)

PR #149, opened 2019-02-19
[IMP] udes_stock: Add RPC timing boolean flag to warehouse config

richardwhitefoot:


addons/udes_stock/models/stock_warehouse.py
```diff
@@ -78,6 +78,8 @@ def _domain_u_damaged_location_id(self):
         string='Stock Investigation Picking Type',
         help="Picking type used to create stock investigation pickings."
     )
+    u_show_rpc_timing = fields.Boolean(string="Show RPC timing", default=False,
+                                       help="Show RPC call times on mobile UI")
```
parisiale: Perhaps "timings" instead of "times"? But I'm the foreigner... :P

parisiale:
👍

PR #150, opened 2019-02-20
[ADD] udes_sale_stock: `cancel_date` field to SaleOrderLine

parisiale:
Defining and assigning the timestamp of cancellation
to sale order lines.

User-story: 3020

### DIFF
--- a/addons/udes_sale_stock/models/sale_order_line.py
+++ b/addons/udes_sale_stock/models/sale_order_line.py
@@ -1,5 +1,7 @@
 """Sale order line"""
 
+from datetime import datetime
+
 from odoo import api, models, fields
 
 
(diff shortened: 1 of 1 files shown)


addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -9,19 +11,31 @@ class SaleOrderLine(models.Model):
...
+    cancel_date = fields.Datetime(string='Cancel Date',
+                                  help='Time of cancellation',
+                                  readonly=True, index=True)
 
     @api.depends('move_ids.state')
     def _compute_is_cancelled(self):
         location_customers = self.env.ref('stock.stock_location_customers')
+        now_date = datetime.now()
+
+        def not_cancelled_filter(m):
```
unirho: just as a question, do we like to do this kind of defining rather than using a lambda? I only see lambdas in the code so I dno if there is a reason for it or a preference or neither as the reason why.

PR #153, opened 2019-02-21
[FIX] udes_sale_stock: Set sale_line is_cancelled on picking.action_cancel

joe-gb:
Using api onchange is inefficient, it will calcualte for every state
change. Set this only when cancelling the final picking.

### DIFF
--- a/addons/udes_sale_stock/models/stock_move.py
+++ b/addons/udes_sale_stock/models/stock_move.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from odoo import api, fields, models
+from datetime import datetime
 
 import logging
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


addons/udes_sale_stock/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+        """
+        location_customers = self.env.ref('stock.stock_location_customers')
+        res = super(StockPicking, self).action_cancel()
+        for move in self.mapped('move_lines'):
+            sale_line = move.sale_line_id
+            if sale_line and not sale_line.is_cancelled:
+                not_cancelled = sale_line.move_ids.filtered(
+                    lambda m: m.state not in ['cancel'] and
+                              m.location_dest_id == location_customers)
+                sale_line.is_cancelled = len(not_cancelled) == 0
```
ws-mps: Call  `order.check_state_cancelled()` here and remove sale.order.line.compute_order_state().

ws-mps: Batch all the sale lines to set is_cancelled = True and do it in one call .write()

joe-gb: done

joe-gb: changed this a bit, looks mental but works .. using ORM API calls instead of loops

addons/udes_sale_stock/models/stock_picking.py
```diff
@@ -0,0 +1,30 @@
...
+
+        def not_cancelled_filter(m):
+            return m.state not in ['cancel']\
+                   and m.location_dest_id == location_customers
+
+        lines_to_cancel = self.filtered(lambda p: p.location_dest_id == location_customers) \
+            .mapped('move_lines.sale_line_id') \
+            .filtered(lambda s: len(s.move_ids.filtered(not_cancelled_filter)) == 0)
+
+        lines_to_cancel.write({'is_cancelled': True})
```
ws-mps: Merge writes into one

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -16,10 +17,27 @@ def _action_done(self):
...
             self.mapped('sale_line_id.order_id').check_delivered()
 
+        def not_cancelled_filter(m):
+            return m.state not in ['cancel'] \
+                   and m.location_dest_id == location_customers
+
+        lines_to_cancel = self.filtered(
+            lambda m: m.location_dest_id == location_customers) \
+            .mapped('sale_line_id') \
+            .filtered(
```
ws-mps: This can go to the line above.

PR #154, opened 2019-02-22
[IMP] udes_stock: when filtering with categories, allow users with some perm to pick from locs with missing categories

andrewm314:
[IMP] udes_stock: when filtering by location categories, allow users with any perms to pick from locations with missing/NULL categories


crwr45:
Issue arose due to incorrect configuration. location categories can be turned off for this.

PR #156, opened 2019-02-26
Story/3264

ws-mps:


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -481,10 +483,29 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode):
...
+                to_update['location_dest_id'] = dest_loc.id
             pickings = completed_move_lines.mapped('picking_id')
+            picking_type = pickings.mapped('picking_type_id')
+            picking_type.ensure_one()
+            if picking_type.u_scan_parent_package_end:
+                if not result_package:
+                    raise ValidationError(
+                        _('Expecting result package on drop off.')
+                    )
+                result_package = Package.get_package(result_package)
```
parisiale: Nasty nasty shadowing.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -78,6 +78,12 @@ class StockPickingType(models.Model):
         'scan something from pickings of this type'
     )
 
+    u_scan_parent_package_end = fields.Boolean(
+        string='Scan parent package at the end',
+        default=False,
+        help='If True, the user is asked to scan parent package on drop off.',
```
parisiale: Not true. That would depend on `u_user_scans` and `u_target_storage_format`. So, having such boolean option is misleading.

As discussed, we should move such logic from the frontend. And use an enumeration for the 3 possible cases (don't scan, scan at start, scan at the end). We should at least add a TODO here.

addons/udes_stock/controllers/stock_picking_batch.py
```diff
@@ -173,7 +173,8 @@ def create_batch_for_user(self,
     def update_batch(self, ident,
                      continue_batch,
                      move_line_ids,
-                     location_barcode=None):
+                     location_barcode=None,
+                     result_package=None):
```
parisiale: This needs to go in the specs (README).

parisiale: Maybe we can revisit the naming.
We're using `result_package_name` for `pick` and `result_package_id` for move line. This is extremely confusing.

parisiale: As agreed with @richardwhitefoot, will use `result_package_name`.

PR #158, opened 2019-03-01
[IMP] udes_stock: Adding picking date_done to form and tree views

joe-gb:


addons/udes_stock/views/stock_picking.xml
```diff
@@ -59,7 +59,12 @@
             <!-- origin (Source Document  field changed to required for goods-in -->
             <xpath expr="//field[@name='origin']" position="attributes">
                 <attribute name="attrs">{'required': [('picking_type_code', '=', 'incoming')]}</attribute>
-            </xpath>              
+            </xpath>
+
+            <!-- Add date done to form view -->
+            <xpath expr="//group/group[2]" position="inside">
```
ws-mps: Can we add it after the scheduled_date.
`<xpath expr="//field[@name='scheduled_date']" position="after">`

addons/udes_stock/views/stock_picking.xml
```diff
@@ -120,6 +125,11 @@
 
             </xpath>
 
+            <!-- Show date done -->
+            <xpath expr="//field[@name='state']" position="after">
```
ws-mps: As above.

PR #174, opened 2019-03-25
[IMP] udes_sale_stock: Add ability to cancel unfulfillable sale orders

joe-gb:
Added a method to cancel unfulfillable sale orders, logic will search
sale orders that need fulfilling and compare with stock availability,
cancelling orders that current stock holding is not enough for.

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 
+from collections import defaultdict
 from odoo import api, fields, models
 from odoo.addons.udes_stock.models import common
+import logging
+
+_logger = logging.getLogger(__name__)
+
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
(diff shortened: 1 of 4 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+        quants = Stock.search(domain)
+        available_quantity = sum(quants.mapped('quantity')) - \
+                             sum(quants.mapped('reserved_quantity'))
+        return available_quantity
+
+    @api.model
+    def cancel_orders_without_availability(self):
+        """ From the current list of unconfirmed SO lines, cancel lines that
+        cannot be fulfilled with current stock holding """
+        Order_line = self.env['sale.order.line']
```
ws-mps: OrderLine

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+
+        # Get order lines
+        order_lines = Order_line.search([('state', 'in', ['sale', 'draft'])])
+
+        # Get unreserved stock for each product in locations
+        location_ids = self.get_available_stock_locations()
```
ws-mps: Better use `locations`, since we are returning recordsets

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+        stock = {}
+        for pid in order_lines.mapped('product_id'):
```
ws-mps: Better use `product` since we are getting a product recordset.

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+                stock[pid] = stock.get(pid, 0) + \
+                             self.get_available_quantity(pid, location_ids)
+
+        # Create empty record sets for SO lines
+        can_fulfill = Order_line.browse()
+        cant_fulfill = Order_line.browse()
+
+        # Group lines by order
+        mls_grouped = order_lines.groupby(key=lambda x: x.order_id, sort=False)
```
ws-mps: Use `lines_grouped`, since `mls` is meaningless here.

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+
+        # Get order lines
+        order_lines = Order_line.search([('state', 'in', ['sale', 'draft'])])
+
+        # Get unreserved stock for each product in locations
+        location_ids = self.get_available_stock_locations()
+        stock = {}
+        for pid in order_lines.mapped('product_id'):
+                stock[pid] = stock.get(pid, 0) + \
+                             self.get_available_quantity(pid, location_ids)
```
squah-u: Can `order_lines.mapped('product_id')` return contain the same product multiple times?
Is the `stock.get(pid, 0)` unnecessary or can we double count available stock here?

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -40,6 +44,88 @@ def _set_priority(self):
...
+
```
squah-u: A [defaultdict](https://docs.python.org/2/library/collections.html#defaultdict-examples) could be used for `stock` in this function

addons/udes_sale_stock/models/sale_order.py
ws-mps: Good question, `mapped` does not return the same product mulitple times.

Good point, it is unnecessary.

addons/udes_sale_stock/models/sale_order.py
ws-mps: Agree.

joe-gb: never used defaultdict before, good shout 👍

PR #177, opened 2019-03-28
Merge/3490

ws-mps:


addons/udes_stock/models/stock_picking.py
```diff
@@ -1707,15 +1711,22 @@ def reserve_stock(self):
...
             processed = Picking.browse()
             by_type = lambda x: x.picking_type_id == picking_type
 
-            while to_reserve > 0 or reserve_all:
+            if not reserve_all:
+                ready_pickings = Picking.search([
+                    ('picking_type_id', '=', picking_type.id),
+                    ('state', '=', 'assigned'),
+                ])
+                to_reserve -= len(ready_pickings)
```
kdwyer: Could we use [search_count](https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L1451), as we aren't using `ready_pickings` for anything else?

ws-mps: Yes, we do.

PR #182, opened 2019-04-05
[IMP] udes_stock: Re-use existing pickings in refactor when appropriate

squah-u:
If the moves in a post-refactor picking match the moves in an original
picking exactly, reuse that picking instead of creating a new one.

User-story: 3591
Task: 3738

### DIFF
--- a/addons/udes_sale_stock/tests/__init__.py
+++ b/addons/udes_sale_stock/tests/__init__.py
@@ -1,4 +1,4 @@
 """UDES sale_Stock tests"""
 
 from . import common
-# from . import test_sale_order
+from . import test_cancel_unfulfillable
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -1342,7 +1345,24 @@ def _new_picking_for_group(self, group_key, moves, **kwargs):
...
         ])
+
+        if not picking and picking_for_reuse:
+            # Otherwise use the suggested picking
+            if (picking_for_reuse.picking_type_id.id != picking_type.id or
+                picking_for_reuse.location_id.id != src_loc.id or
+                picking_for_reuse.location_dest_id.id != dest_loc.id or
+                picking_for_reuse.state not in
+                    ('assigned', 'confirmed', 'waiting')):
+                raise ValueError(_('Cannot reuse picking because it is not '
```
ws-mps: This changes the current behaviour, if we cannot find a picking we create a new one. Therefore if the picking to reuse does not meet the search domain we should create a new picking.

PR #186, opened 2019-04-09
[IMP] udes_stock: Re-use existing picks in refactor when appropriate

squah-u:
If all the remaining moves of a picking are to be placed in a new picking,
the existing picking is reused instead of creating a new one.

User-story: 3591
Task: 3738

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1334,6 +1334,8 @@ class StockPicking(models.Model):
 
         group = Group.get_group(group_identifier=group_key,
                                 create=True)
+
+        # Look for an existing picking with the right group
         picking = self.search([
             ('picking_type_id', '=', picking_type.id),
             ('location_id', '=', src_loc.id),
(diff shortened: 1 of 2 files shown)


squah-u:
Previous abandoned merge request: https://github.com/unipartdigital/udes-open/pull/182

addons/udes_stock/models/stock_picking.py
```diff
@@ -1342,7 +1352,22 @@ def _new_picking_for_group(self, group_key, moves, **kwargs):
...
+        if not picking:
+            # Otherwise reuse the existing picking if all the moves
+            # already belong to it and it contains no other moves
+            # The picking_type_id, location_id and location_dest_id
+            # will match already
+            current_picking = moves.mapped('picking_id')
+            if (len(current_picking) == 1 and
+                current_picking.mapped('move_lines') == moves):
+                current_picking.write({ 'group_id': group.id })
+                current_picking.write(kwargs)
```
crwr45: each call to `write()` here will result in a trip to the database. Better to copy/update kwargs with the new value then `write` once.

PR #189, opened 2019-04-11
[FIX] udes_stock: reservation popup shows in-stock products.

kdwyer:
Story/3838
Task/3841

If an attempt to reserve stock for an individual picking fails, the
error message displayed to the user reports all products for the
picking, regardless of whether or not there is sufficient stock to
reserve each product.

This patch ensures only unreservable products are reported.

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1775,8 +1775,12 @@ class StockPicking(models.Model):
                     if unsatisfied:
                         if self:
                             # we need to construct our error message before the
-                            # changes are rolled back.
-                            moves = unsatisfied.mapped('move_lines')
+                            # changes are rolled back, and report only products
+                            # that are unreservable.
+                            not_done = lambda x: x.state not in (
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -1775,8 +1775,12 @@ def reserve_stock(self):
                     if unsatisfied:
                         if self:
                             # we need to construct our error message before the
-                            # changes are rolled back.
-                            moves = unsatisfied.mapped('move_lines')
+                            # changes are rolled back, and report only products
+                            # that are unreservable.
+                            done = ('done', 'assigned', 'cancelled')
```
ws-mps: This is a constant, define it or replace where we use `done` at next line.

kdwyer: @ws-mps constant inlined.

ws-mps: Thanks!

kdwyer:
In master as af67ec087, so PR may be closed.

PR #192, opened 2019-04-16
Story/3762

apatrick88:


addons/udes_stock/models/stock_picking.py
```diff
@@ -532,6 +537,9 @@ def create_picking(
             # picking.move_line_ids.mapped('result_package_id').write({'package_id': False})
             pass
 
+        if create_batch:
+            batch = PickingBatch.create_batch(picking_type_id,None)
```
crwr45: This should explicitly create a batch with only the picking that was just created. I haven't looked at the implementation of create_batch, but I think it can potentially pick the oldest unbatched picking of the given type. If there are other Internal Transfers in the system outside batches then the user might get a different one than the one that was just created.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -210,6 +210,12 @@ def _remove_unready_picks(self, picks=None):
             if confirmed_not_ready:
                 # Attempt to make confirmed picks ready
                 confirmed_not_ready.action_assign()
+                # TODO This has the undesired effect of having a cancelled
+                # picking with stock still assigned to it. When a picking is
+                # cancelled, this is fired before the relevant picking and batch
+                # state changes have all been made, resulting in the newly unreserved
+                # stock for the cancelled picking being reserved again via action_assign.
+                # Needs design and a rethink.
```
crwr45: Good summary

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -160,6 +160,6 @@ def _get_a_type(self, picking_type_name):
             lambda pt: pt.name == picking_type_name)
 
         if not put:
-            put = None
+            put = self.int_type_id
```
crwr45: Id like to see some explanation of why this was necessary in the commit. I'm sure we only just changed it to None.

PR #197, opened 2019-05-13
[ADD] Handle confirmed not ready picks

davidmcgaw:
added failing tests in class TestBatchState:
* test13_unreserves_stock_after_cancel_pick
* test14_unreserves_stock_after_unreserve_pick
Checking that when unreserving/cancelling a pick in a batch the stock gets unreserved

Task: 4072
Story: 4059

### DIFF
--- a/addons/udes_stock/tests/test_picking_batch.py
+++ b/addons/udes_stock/tests/test_picking_batch.py
@@ -1110,6 +1110,14 @@ class TestBatchState(common.BaseUDES):
         if call_done:
             picking.action_done()
 
+    def assert_number_of_reserved_apples_is(self, expected_amount):
+        """Asserts the amount of apples reserved"""
+        quant = self.env['stock.quant']._gather(self.apple, self.test_location_01)
+        self.assertEqual(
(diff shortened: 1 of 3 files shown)


addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1372,6 +1380,32 @@ def test12_computing_partial_assignment(self):
...
+
+    def test14_unreserves_stock_after_unreserve_pick(self):
+        """ Unreserve pick and check that stock is unreserved"""
+        self.draft_to_ready()
+        self.assign_user()
+        self.assert_number_of_reserved_apples_is(4)
+        self.assertEqual(self.picking01.state, 'assigned', 'Picking 1 should be assigned')
+        self.picking01.do_unreserve()
+        # Unreserve the pick and confirm we reach state done
+        self.picking01.do_unreserve()
```
apatrick88: do_unreserve() is called twice.

davidmcgaw: yes. I've removed the duplicated call, a similar thing was happening for test13 that I have also fixed

PR #203, opened 2019-05-20
[FIX] udes_security: Fix Trusted User write check and add tests

notcake:
[FIX] udes_security: Fix Trusted User write check

Fix the res.users write override to correctly handle all many2many write
operations. Also remove the assumption that many2many write operations have
the right number of elements and no more.

Task: 4209
User-story: 3917

[ADD] udes_security: Add tests for restrictions on granting Trusted User

Task: 4209
User-story: 3917

### DIFF
--- a/addons/udes_security/__init__.py
+++ b/addons/udes_security/__init__.py
@@ -2,3 +2,4 @@
 
 from . import controllers
 from . import models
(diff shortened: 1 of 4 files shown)


notcake:
Wrong user account.

PR #208, opened 2019-06-04
Epic/4189 story 3886

unirho:


addons/udes_stock/models/stock_location.py
```diff
@@ -428,20 +428,53 @@ def _get_stock_drift(self, adjustments_request):
...
             # determine the package
-
             package_name = adj['package_name']
             package = None
-
             if NO_PACKAGE_TOKEN in package_name:
-                pass
+                if products_and_quantities.get(product) == adj['quantity']:
+                    continue
+                else:
```
nocturnalastro: no need for the `else: pass`

unirho: so i should just combine the booleans right

unirho: oops just read it back, that would break with the elif and else

PR #210, opened 2019-06-07
Tidy up empty pickings

ws-mps:


addons/udes_stock/models/procurement.py
```diff
@@ -39,3 +39,17 @@ def get_group(self, group_identifier, create=False):
...
+        """ Override run to clean up empty pickings in case any refactoring has
+            occurred.
+        """
+        Picking = self.env['stock.picking']
+
+        super().run(product_id, product_qty, product_uom, location_id, name,
+                    origin, values)
+        Picking.unlink_empty()
+
+        return True
```
joe-gb: should we return true, or should we return the result of the super call?

addons/udes_stock/models/stock_picking.py
```diff
@@ -1159,16 +1159,36 @@ def action_confirm(self):
...
+            ])
+            records = records.filtered(lambda p: len(p.move_lines) == 0)
+
+        return records
+
     def unlink_empty(self):
         """
-            Delete pickings in self that are empty, locked and cancelled
+            Delete pickings in self that are empty, locked and marked=False.
+            If self is empty, find and delete any picking with the precious
```
squah-u: minor: typo: "precious"

addons/udes_stock/models/stock_picking.py
```diff
@@ -1159,16 +1159,36 @@ def action_confirm(self):
...
+        """
+        if self:
+            records = self.filtered(
+                lambda p:
+                    (len(p.move_lines) == 0
+                    and not p.u_mark
+                    and p.is_locked))
+        else:
+            records = self.search([
+                ('u_mark', '=', False),
```
squah-u: now that we're doing this query, it could be worth adding an index on `u_mark`

addons/udes_stock/models/procurement.py
ws-mps: The return of the super call is True.

addons/udes_stock/models/stock_picking.py
ws-mps: Haven't checked, but will add it if it is not there, thanks.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1159,16 +1159,36 @@ def action_confirm(self):
...
+            criterion.
             This is to prevent us leaving junk data behind when refactoring
         """
-        self.filtered(lambda p:
-                      (len(p.move_lines) == 0
-                       and not p.u_mark
-                       and p.is_locked)).unlink()
-        return self.exists()
+        records = self.find_empty_pickings_to_unlink()
+        records.unlink()
```
crwr45: My only comment here would be to add batching, in case the amount of data is large (e.g. job hasn't been run in a long time). Not a requirement though, since the job is run every 30 minutes by default.

ws-mps: Will add thanks.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -307,6 +307,13 @@ class StockPickingType(models.Model):
              'This is ignored if the number of pickings to reserve is 0.'
     )
 
+    u_auto_unlink_empty = fields.Boolean(
+        string='Auto Unlink Empty',
+        default=True,
+        help='When False, filter out empty pickings of this picking type when '
```
kdwyer: Can we explain what the flag does when it's enabled?  The double negative (unlink, False) could be difficult to understand.

PR #211, opened 2019-06-11
[FIX] udes_security: Further measures to tackle open redirects

apatrick88:
Story/4116


squah-u:
I think we need to check for both `scheme` or `netloc`. url parsing is really weird.
`http://localhost:8069/web/login?redirect=https:google.com` passes the new filter and works on Chrome and Firefox.

PR #212, opened 2019-06-13
[IMP] udes_stock: Prevent locations having more than one orderpoint.

kdwyer:
Story/3885

This patch adds a boolean flag to the location model.  If set, the flag
prevents having more than one orderpoint on a location.

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -96,6 +96,13 @@ class StockLocation(models.Model):
              "equipment to pick from it.",
     )
 
+    u_limit_orderpoints = fields.Boolean(
+        index=True,
+        string="Limit Orderpoints",
+        help="If set, allow only one orderpoint on this location "
+        "and its descendants.",
+    )
+
     def _prepare_info(self, extended=False, load_quants=False):
         """
             Prepares the following info of the location in self:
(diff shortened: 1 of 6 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -499,3 +505,38 @@ def apply_location_policy_change_to_descendants(self):
...
+        """
+        limited = self.search([('u_limit_orderpoints', '=', True)])
+        return bool(self.search([('id', 'child_of', limited.ids)]))
+
+
+class Orderpoint(models.Model):
+
+    _inherit = 'stock.warehouse.orderpoint'
+
+    @api.constrains('location_id')
```
ws-mps: onchange also :)

addons/udes_stock/models/stock_location.py
```diff
@@ -499,3 +506,44 @@ def apply_location_policy_change_to_descendants(self):
...
+        self.ensure_one()
+        limited = self.search([('u_limit_orderpoints', '=', True)])
+        return self in self.search([('id', 'child_of', limited.ids)])
+
+
+class Orderpoint(models.Model):
+
+    _inherit = 'stock.warehouse.orderpoint'
+
+    @api.onchange('locatiion_id')
```
ws-mps: typo

addons/udes_stock/models/stock_location.py
```diff
@@ -499,3 +506,45 @@ def apply_location_policy_change_to_descendants(self):
...
+        Orderpoint = self.env['stock.warehouse.orderpoint']
+        orderpoints = Orderpoint.search([
+            ('location_id', '=', self.location_id.id),
+        ])
+        orderpoints -= self
+        if not orderpoints:
+            return
+        if not self.location_id.limits_orderpoints():
+            return
+        names = ', '.join(x.product_id.name for x in orderpoints)
```
ws-mps: x.mapped('product_id.name') ?

addons/udes_stock/models/stock_location.py
```diff
@@ -499,3 +506,45 @@ def apply_location_policy_change_to_descendants(self):
...
+        raise ValidationError(
+            _('An order point for location {} already exists on '
+              '{}.').format(self.location_id.name, names))
+        return
```
ws-mps: Not needed.

addons/udes_stock/tests/test_limit_orderpoints.py
```diff
@@ -0,0 +1,56 @@
...
+        """ Tests cannot add an orderpoint if one already exists and the limit
+        orderpoints flag is set on the location.
+        """
+        self.test_output_location_01.u_limit_orderpoints = True
+        with self.assertRaises(ValidationError) as cm:
+            self.create_orderpoint(self.banana,
+                                   self.test_output_location_01,
+                                   5, 10)
+        ex = cm.exception
+        self.assertEqual('An order point for location Test output location 01 '
```
ws-mps: Does it worth to replace `Test output location 01` by  `%s` and use `self.test_output_location_01`? in case the name of the location changes the test is not going to fail since we don't care about the name of the location.

PR #222, opened 2019-07-11
[ADD] udes_stock: Refactor action to group pickings

samueljsb:
Add a refactor action to allow pickings to be grouped in draft batches.
A generic method allows methods to be written to group by an arbitrary
key function. Specific actions allow picks to be grouped by date or by
date and priority.

User-story: 4543

Signed-off-by: Samuel Searles-Bryant <samuel.searles-bryant@unipart.io>

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -519,3 +519,56 @@ class StockMove(models.Model):
             })
 
         return result_moves
+
+    @api.multi
+    def refactor_action_batch_pickings_by_date_priority(self):
+        """Batch pickings by date and priority."""
+        self._refactor_action_batch_pickings_by(
+            lambda picking: (picking.scheduled_date.split()[0], picking.priority)
+        )
+
+    @api.multi
+    def refactor_action_batch_pickings_by_date(self):
+        """Batch pickings by date."""
+        self._refactor_action_batch_pickings_by(
+            lambda picking: (picking.scheduled_date.split()[0],)
+        )
+
+    def _refactor_action_batch_pickings_by(self, by_key):
+        """Group picks in batches.
+
+        Move the pickings of the moves in this StockMove into draft batches grouped by a
+        given key.
+
+        Args:
+            by_key (function): The function to generate the key to group pickings by.
+                Should return a :obj:`tuple`.
+        """
+        PickingBatch = self.env["stock.picking.batch"]
+
+        # Find existing draft batches.
+        picking_types = self.mapped("picking_type_id")
+        batches = PickingBatch.search(
+            [("state", "=", "draft"), ("picking_type_ids", "in", picking_types.ids)]
+        )
+        batches.mapped("picking_ids")
+
+        # Index coherent batches by key.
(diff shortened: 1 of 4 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -519,3 +519,53 @@ def refactor_by_move_line_groups(self, groups):
...
+            by_key (function): The function to generate the key to group pickings by.
+                Should return a :obj:`tuple`.
+        """
+        PickingBatch = self.env["stock.picking.batch"]
+
+        # Precache pickings.
+        self.mapped("picking_id")
+
+        # Find existing draft batches.
+        batches = PickingBatch.search([("state", "=", "draft")])
```
ws-mps: Also filter by picking type. All the pickings in self should be of one picking type.

addons/udes_stock/models/stock_move.py
```diff
@@ -519,3 +519,53 @@ def refactor_by_move_line_groups(self, groups):
...
+
+    @api.multi
+    def refactor_action_batch_pickings_by_date(self):
+        """Batch pickings by date."""
+        self._refactor_action_batch_pickings_by(
+            lambda picking: (picking.scheduled_date,)
+        )
+
+    def _refactor_action_batch_pickings_by(self, by_key):
+        """Group picks in batches.
```
ws-mps: Mention that we are moving the pickings of the moves in self.

addons/udes_stock/models/stock_move.py
```diff
@@ -519,3 +519,53 @@ def refactor_by_move_line_groups(self, groups):
...
+
+        Args:
+            by_key (function): The function to generate the key to group pickings by.
+                Should return a :obj:`tuple`.
+        """
+        PickingBatch = self.env["stock.picking.batch"]
+
+        # Precache pickings.
+        self.mapped("picking_id")
```
ws-mps: Not sure this is needed, since we are doing the same in line 562 and we dont't have any other access to the pickings of the moves until that.

addons/udes_stock/models/stock_move.py
ws-mps: The reason is because we can have the same refactoring applied to different picking types, and we don't want the pickings of different picking types to be merged into the same batch.

samueljsb:
```python
    ("picking_type", "=", self.picking_type)
```
?

samueljsb: Or 
```python
    (self.picking_type_id, "in", "picking_type_ids")
```
?

ws-mps: `('picking_type_ids', 'in', picking_types.ids)`

samueljsb: `picking_type` isn't a variable - do you mean the `self.picking_type_id`?

ws-mps: Yes, something like:
`picking_types = self.mapped('picking_type_id')`
Because self can contain more than one stock move.

PR #225, opened 2019-07-24
[IMP] mrp: make waiting Manufacturing Orders not show as todo

przemub:
Since they already show as waiting, it was requested that they do not show as todo.

Story: 4344
Task: 4716
Signed-off-by: Przemysław Buczkowski <pbuczkowski@unipart.io>

### DIFF
--- a/addons/udes_mrp/__manifest__.py
+++ b/addons/udes_mrp/__manifest__.py
@@ -25,7 +25,7 @@
     # always loaded
     'data': [
         'report/mrp_production_templates.xml',
-
+        'views/mrp_production_views.xml'
     ],
 
     # only loaded in demonstration mode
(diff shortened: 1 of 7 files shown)


crwr45:
@przemub There should be no mention of the customer on a public repo. Please reword the commit message and the MR

przemub:
@crwr45 Changed, thanks.

addons/udes_mrp/__manifest__.py
```diff
@@ -19,13 +19,14 @@
     'depends': [
         'base',
         'mrp',
+        'stock',
```
crwr45: I wouldn't consider this necessary since the `mrp` module already depends on `stock`. since `udes_mrp` is now directly modifying `stock` things (`stock.picking.type`) it's OK to leave it as explicit now.

przemub: I went and removed it but it seems I haven't understood the latter sentence - should I leave it or remove it? :P

crwr45: I communicated poorly, and don't really mind, but I'd prefer not having it. Remove it.

PR #230, opened 2019-08-22
[ADD] udes_sale_stock: Improve characterisation of sale order state

unirho:
The check_delivered function did not account for a mixed terminal
state sale order, this has been addressed by confirming an order as
done when it's terminal pickings (i.e. all the last pickings in a
chain) are in a terminal state (done or cancelled).

User-story: 4516


nocturnalastro:
Already in master

PR #231, opened 2019-09-05
[IMP] udes_sale_stock,udes_stock: Adding auto-confirm for sale orders

joe-gb:
Adding config and functionality to automatically confirm sale orders
based on an x days in the  future value. Triggers so.action_confirm()
on all orders within x days (0 being today only)

Also adding a string to the UDES stock warehouse configs for clarity.

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -4,6 +4,8 @@ from collections import defaultdict
 from odoo import api, fields, models
 from odoo.addons.udes_stock.models import common
 import logging
+from datetime import timedelta, date
+
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -176,6 +178,30 @@ def check_state_cancelled(self):
...
+        """
+        days = self.env.ref('stock.warehouse0').u_so_auto_confirm_ahead_days
+        unconfirmed_states = ('draft', 'sent')
+        unconfirmed_so = self or self.search(
+            [('state', 'in', unconfirmed_states)]
+        )
+
+        # If ahead days set to -1, confirm all.
+        if days == -1:
+            return self.action_confirm()
```
squah-u: Is this supposed to be `unconfirmed_so.action_confirm()` ?

joe-gb: yes! thank you

PR #236, opened 2019-11-25
Story/9034

samclk:


addons/udes_stock/static/src/xml/create_planned_transfer_button.xml
```diff
@@ -13,15 +13,9 @@
         </t>
     </t>
 
-    <!--Replace the create button on a picking formview to have 'Planned Transfer' functionality instead of 'Immediate Transfer'-->
+    <!--Remove the create button on a picking form view-->
     <t t-extend="FormView.buttons">
         <t t-jquery="button.o_form_button_create" t-operation="replace">
-            <t t-if="widget and widget.modelName=='stock.picking'">
```
nocturnalastro: Don't we want this if the person is an admin or privilaged user?

addons/udes_stock/views/edi_menu_views.xml
```diff
@@ -6,5 +6,9 @@
     <menuitem id="edi_record_menu" name="UDES"
 	      parent="edi.record_menu" sequence="500"/>
 
+    <!-- EDI move tracker menu item -->
```
nocturnalastro: does udes_stock depennd on edi? if not this needs to live somewhere else.

joe-gb: it does!

addons/udes_stock/static/src/xml/create_planned_transfer_button.xml
joe-gb: don't think it's needed in form view, we still have the "create" on list view, and "new planned transfer" on the picking widgets - I see no use for a "Create" button when already looking at a picking! but it's one we can definitely ask to product :)

PR #238, opened 2019-12-02
[FIX] udes_stock: Don't recompute batch on picking cancel until done

joe-gb:
Recomputing the batch AFTER the picking cancel logic prevents quant
re-reservation being triggered on something that in in process of
being cancelled, and resolves the issue of batch-cancelled picks
still having reserved stock.

### DIFF
--- a/addons/udes_stock/tests/test_batch_pickings.py
+++ b/addons/udes_stock/tests/test_batch_pickings.py
@@ -322,3 +322,98 @@ class TestByDate(BaseUDES):
 
         picks.mapped("move_lines").action_refactor()
         self.assertFactorised()
+
+
+class TestCancelInBatch(BaseUDES):
+    """Tests for cancelling picks in a batch."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Setup for test class."""
+        super().setUpClass()
+        PickingBatch = cls.env["stock.picking.batch"]
+        cls.batch = PickingBatch.create({"name": "batch001",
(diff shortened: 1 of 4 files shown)


apatrick88:
LGTM

addons/udes_stock/tests/test_batch_pickings.py
```diff
@@ -322,3 +322,114 @@ def test04_fix(self):
...
+        self.batch.confirm_picking()
+
+        # Confirm batch state
+        self.assertEqual(self.batch.state, 'in_progress')
+
+        # Confirm reserved stock
+        reserved = {}
+        for move in self.pick1.move_line_ids:
+            reserved[move.product_id.name] = move.product_qty
+        self.assertEqual(reserved, {self.banana.name: 4, self.apple.name: 2})
```
davidmcgaw: This block of code is repeated several few times.

The generation of reserved could be done on one line using list comprehension.

``reserved = {move.product_id.name:move.product_qty for move in self.pick1.move_line_ids}``

Asserting the reserved amount for a pick could be a method of the test class, e.g.

```
def assertReserved(self, pick, expected):
    reserved = {move.product_id.name:move.product_qty for move in pick.move_line_ids}
    self.assertEqual(reserved, expected)
```

It would be less code and more readable IMHO

joe-gb: good suggestion, tyvm

PR #239, opened 2019-12-05
[REF] udes_stock: Refactor batch state computation

joe-gb:


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -113,6 +113,31 @@ def _compute_priority(self):
...
+                    batch.picking_ids.filtered(
+                        lambda p: p.state == "confirmed" and
+                                  len(p.mapped("move_lines").filtered(
+                                      lambda move: move.state
+                                                   not in (
+                                                   "draft", "cancel", "done")
+                                  )) > 0
+                    ).with_context(lock_batch_state=True).action_assign()
+                except:
+                    pass
```
apatrick88: Can we have some more detail here, maybe log the exception rather than consume it.

joe-gb: good spot! wasn't needed at all :)

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -113,6 +113,28 @@ def _compute_priority(self):
...
+        for batch in self:
+            if batch.state in ["waiting", "in_progress"]:
+                picks_to_assign = batch.picking_ids.filtered(
+                    lambda p: p.state == "confirmed"
+                    and p.mapped("move_lines").filtered(
+                        lambda move: move.state not in ("draft", "cancel", "done")
+                    )
+                )
+                if picks_to_assign:
+                    picks_to_assign.with_context(lock_batch_state=True).action_assign()
```
ws-mps: If all the pickings are successfully assigned, the state of the batch should change to ready if it was waiting?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -130,91 +152,66 @@ def _compute_state(self):
...
         for batch in self:
-            if batch.state in ['draft', 'done', 'cancel']:
+            if batch.state in ['draft', 'cancel']:
                 # Can not do anything with them don't bother trying
                 continue
 
-            ready_picks, other_picks = \
-                batch._calculate_pick_groups(batch.picking_ids)
+            picks = batch.picking_ids
+            if picks:
```
ws-mps: `if batch.picking_ids` because picks is not used

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -82,6 +82,14 @@ def _domain_u_damaged_location_id(self):
     u_show_rpc_timing = fields.Boolean(string="Show RPC timing", default=False,
                                        help="Show RPC call times on mobile UI")
 
+    u_remove_unready_batch = fields.Boolean(
```
ws-mps: Different behaviours might be wanted for different picking types.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -207,7 +209,6 @@ def test09_create_batch_user_already_has_completed_batch(self):
                               result_package_name=None)
 
         # check the picking is done and the backorder is not in the batch
-        self.assertEqual(picking.state, 'done')
```
ws-mps: This is not needed anymore?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1219,7 +1222,7 @@ def test08_check_computing_simple(self):
             self.batch01.confirm_picking()
 
         mocked_compute.assert_called_with(self.batch01)
-        self.assertEqual(
+        self.assertGreaterEqual(
```
ws-mps: Does this mean that the function is called more than once now?

joe-gb: yes, it can be called but may exit because of state_lock

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -132,91 +159,83 @@ def _compute_state(self):
             return
 
         for batch in self:
-            if batch.state in ['draft', 'done', 'cancel']:
+            if batch.state in ['draft', 'cancel']:
```
nocturnalastro: why not done?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -115,6 +115,33 @@ def _compute_priority(self):
...
+        """If configured, attempt to assign all the relevant pickings in self"""
+        if self.env.context.get("lock_batch_state"):
+            # State is locked so don't do anything
+            return
+
+        batches = self.filtered(
+            lambda b: (
+                b.state in ['waiting', 'in_progress']
+                and b.picking_ids # Nothing to act on
+                and all(b.picking_type_ids.mapped("u_auto_assign_batch_pick")) # Is it configured to remove
```
nocturnalastro: do you want to use check_remove_flag here as well?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -132,91 +161,76 @@ def _compute_state(self):
...
-            if confirmed_not_ready:
-                # Attempt to make confirmed picks ready
-                confirmed_not_ready.action_assign()
-
-            # Filter again as we are possibly calling action_assign which
-            # could make some of the previous not ready picks to now be ready
-            not_ready.filtered(not_ready_lam).write({'batch_id': False})
-            self._compute_state()
+        # Remove unready pick, if configured.
+        check_pick_config(unready_picks).write({"batch_id": False})
```
nocturnalastro: you might as well do it inline here I think :)

PR #244, opened 2020-01-01
[IMP] udes_stock: Update task to pass quant_ids to front end

unirho:


addons/udes_stock/models/stock_move_line.py
```diff
@@ -777,10 +777,12 @@ def _prepare_task_info(self):
         user_scans = picking.picking_type_id.u_user_scans
 
         if user_scans == 'product':
-            quants = self.get_quants()
             task['type'] = 'product'
             task['pick_quantity'] = sum(self.mapped('product_qty'))
-            task['quant_id'] = quants.get_info()[0]
+            quants = self.get_quants()
+
+            # make changes here to facilitate passing all the quants
```
nocturnalastro: Comment doesn't make sense

nocturnalastro:
Fix commit message then looks good to me.

PR #247, opened 2020-01-20
Story/9246 scan lots

williamsajr:
Changed behaviour for lot based tracking to allow multiple movelines to pick the same lot number. This is allowed behaviour for lot based tracking (but disallowed for serial based tracking). 

Additionally, a configuration option for scanning the lot number during the pick process was added.
The default behaviour is to scan tracking numbers, as it was before.

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -262,11 +262,12 @@ class StockMoveLine(models.Model):
 
     def _filter_by_products_info(self, products_info):
         """ Filter the move_lines in self by the products in product_ids.
-            When a product is tracked by lot number:
+            When a product is tracked by lot/serial number:
             - when they have lot_id set, they are also filtered by
               lot number and check that they are not done
             - when they have lot_name, it is checked to avoid repeated
-              lot numbers
+              lot numbers (except for lot based tracking where this is
+              allowed)
         """
         # get all move lines of the products in products_info
         move_lines = self.filtered(lambda ml: ml.product_id in products_info)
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -297,14 +298,14 @@ def _filter_by_products_info(self, products_info):
...
-                         product_mls.mapped('picking_id').name))
+                    if product.tracking != 'lot' and set(lot_numbers) != set(mls_lot_numbers):
+                        raise ValidationError(_('Lot numbers %s for product %s not found in '
+                                              'picking %s') %
+                                              (' '.join(diff), product.name,
+                                              product_mls.mapped('picking_id').name))
 
                 done_mls = product_mls_in_lot_numbers.filtered(lambda ml: ml.qty_done > 0)
-                if done_mls:
+                if product.tracking != 'lot' and done_mls == product_mls_in_lot_numbers:
```
nocturnalastro: same here

addons/udes_stock/models/stock_move_line.py
```diff
@@ -411,10 +412,12 @@ def _assert_result_package(self, result_package):
 
     def generate_lot_name(self, lot_names, product):
         """ If required, create a lot and return it's name in a list """
-        confirm_tracking = self.mapped('picking_id.picking_type_id').u_confirm_tracking
+        confirm_tracking = self.mapped('picking_id.picking_type_id').u_scan_tracking
```
nocturnalastro: this assumes that there is only one picking type (either make it explicit or move `u_scan_tracking` into the mapped

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -92,16 +92,15 @@ class StockPickingType(models.Model):
         help="Flag to indicate reservations should be rounded up to entire packages."
     )
 
-    # u_scan_tracking maybe better?
-    u_confirm_tracking = fields.Selection([
+    u_scan_tracking = fields.Selection([
```
nocturnalastro: you need to set `oldname` on the field, otherwise it will not update properly

PR #248, opened 2020-01-21
[IMP] stock_mobile: Allow stock check for locations with bad tracking

andrewm314:
Allow stock check for locations containing products with the wrong lot number tracking type. Will still complain if a user attempts to stock new check items that are badly formed.

 User-story: 9247

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -198,6 +198,11 @@ class StockLocation(models.Model):
 
         product = Product.get_product(int(request['product_id']))
 
+        # Ignore conflicts of tracked vs untracked if we are removing existing
+        # stock
+        if request['quantity'] == 0:
+            return
+
         if 'lot_name' not in request:
             if product.tracking != 'none':
                 raise ValidationError(
@@ -455,7 +460,7 @@ class StockLocation(models.Model):
         # Replace requests with values passed through by reserved filter
         adjustments_request = adj_reqs
 
-        # Readd them to adjustments_request
+        # Read them to adjustments_request
         new_adjs = []
         for product, quantity in no_package_vals.items():
             new_adjs.append({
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -198,6 +198,11 @@ def _validate_inventory_adjustment_request(self, request):
 
         product = Product.get_product(int(request['product_id']))
 
+        # Ignore conflicts of tracked vs untracked if we are removing existing
+        # stock
+        if request['quantity'] == 0:
```
apatrick88: Is this to cover the case when setting existing stock to zero for the old tracked/ untracked quant? Does this not give us a problem when doing a normal stock check (regardless of tracked) and setting quantity to zero for a location?

ws-mps: No, because quantity to zero is only set when something is not accounted during stock check.

addons/udes_stock/tests/test_location_pi.py
```diff
@@ -728,3 +728,42 @@ def test25_process_single_preceding_adjustments_request_mss(self):
...
+        No error raised if the request contains an inventory adjustment entry
+        for a tracked product without a lot name specified if quantity == 0
+        """
+        req = {
+            'location_id': self.test_location_01.id,
+            'inventory_adjustments': [
+                {
+                    'product_id': self.strawberry.id,
+                    'package_name': self.package_one.name,
+                    'quantity': 1
```
apatrick88: Should quantity here be 0 if the docstring mentions quantity 0?

addons/udes_stock/tests/test_location_pi.py
```diff
@@ -728,3 +728,42 @@ def test25_process_single_preceding_adjustments_request_mss(self):
...
+                    'product_id': self.apple.id,
+                    'package_name': self.package_one.name,
+                    'quantity': 0,
+                    'lot_name': 'awesome_lot'
+                }
+            ]
+        }
+
+        self.test_location_01._validate_perpetual_inventory_request(req)
+        self.assertTrue(True)
```
davidmcgaw: This assert is kind of pointless. Is it not possible to check something useful?

PR #250, opened 2020-02-06
[IMP] udes_stock; Link batches and retain user assignment.

kdwyer:
User-story/9326
Task/9394

Do not unassign batches from users at dropoff or release. Instead,
create a copy of the batch with a name that links to the original batch
name and includes those pickings that have not been completed.

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -469,7 +469,7 @@ class StockPickingBatch(models.Model):
             return None
 
         new_name = get_next_name(self, 'picking.batch')
-        batch = self.sudo().copy({'name': new_name})
+        batch = self.sudo().copy({'name': new_name, 'user_id': None})
         _logger.info('Created continuation batch %r, %s', batch, batch.name)
 
         pickings.write({'batch_id': batch.id})
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -468,7 +468,7 @@ class StockPickingBatch(models.Model):
         if not pickings:
             return None
 
-        new_name = get_next_name(self, 'Batch Picking')
+        new_name = get_next_name(self, 'picking.batch')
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -944,3 +974,36 @@ def mark_as_todo(self):
...
+
+    Returns:
+        The generated name, a string.
+    """
+    IrSequence = obj.env['ir.sequence']
+
+    # Name pattern for continuation object.
+    # Is two digits enough?
+    ir_sequence = IrSequence.search([('name', '=', name)], limit=1)
+    name_pattern = r'({}\d{{5}})-(\d{{2}})'.format(re.escape(ir_sequence.prefix))
```
ws-mps: Can we change '\d{{5}}' to any number of digits? Because when the sequence reaches 99999 will carry on to six digits anyway. Also 5 is default lenght, but not always since it is a setting.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -944,3 +974,36 @@ def mark_as_todo(self):
...
+        name - the name for the object's model in the ir_sequence table
+
+    Returns:
+        The generated name, a string.
+    """
+    IrSequence = obj.env['ir.sequence']
+
+    # Name pattern for continuation object.
+    # Is two digits enough?
+    ir_sequence = IrSequence.search([('name', '=', name)], limit=1)
```
ws-mps: It is better to search by code instead of name like `next_by_code()`

kdwyer: Yes - I used `code` originally, but looking at `ir_sequence` code seems to be undefined for most models.  But happy to use it if you think that doesn't matter.

ws-mps: The cases where we will end up using this function will be the ones with code. It would be good to do the same for `stock.picking` backorders.

kdwyer:
Closed in favour of #255

PR #251, opened 2020-02-06
[ADD] edi_notifier: Add module for automating notifications for edi

nocturnalastro:
Adds notifiers (primarily email) which can be trigged by cron or directly there
is two basic implimentations currently Missing (which is tiggered off of crons)
and success - which is called when action_execute completes.

This can only be accessed by a trusted user currently

User-story: 9285

# TODO: 
* ~Add basic email templates~
* ~Add logging~
* ~Add README~
* ~Add tests~


addons/edi_notifier/models/email_notifier.py
```diff
@@ -0,0 +1,101 @@
...
+
+class EdiEmailSuccessNotifier(models.AbstractModel):
+
+    _name = "edi.notifier.email.success"
+    _inherit = "edi.notifier.email"
+
+    def _should_notify(self, notifier, rec):
+        if rec.state != "done":
+            return False
+        return super()._should_notify(notifier, rec)
```
ws-mps: Do we want to do this the other way around? first check if the model has notifier and then check the condition of this type of notifier?

addons/edi_notifier/models/email_notifier.py
```diff
@@ -0,0 +1,101 @@
...
+    _name = "edi.notifier.email.failed"
+    _inherit = "edi.notifier.email"
+
+    def _should_notify(self, notifier, rec):
+        if rec.state == "done":
+            return False
+        return super()._should_notify(notifier, rec)
+
+
+class EdiEmailMissingNotifier(models.AbstractModel):
```
ws-mps: There is a small variant of this notifier, that checks between a range (like last hour) instead of since the last time checked.

nocturnalastro: # TODO
* Update to last_checked
* Factor out lower_bound
* Subclass for range

addons/edi_notifier/models/email_notifier.py
```diff
@@ -0,0 +1,126 @@
...
+        return rec.state != "done"
+
+
+class EdiEmailMissingNotifier(models.AbstractModel):
+
+    _name = "edi.notifier.email.missing"
+    _inherit = "edi.notifier.email"
+
+    can_use_crons = True
+    _timestamp_field = "x_last_checked_not_received"
```
ws-mps: Shouldn't this be last_checked only ?

addons/edi_notifier/models/edi_document.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class EdiDocumentType(models.Model):
+    _inherit = "edi.document.type"
+
+    notifier_ids = fields.Many2many(
+        "edi.notifier", column1="doc_id", column2="notifier_id", string="Notifiers"
+    )
+
+    x_last_checked_not_received = fields.Datetime("Last not recived check")
```
ws-mps: Typo: `recived`

ws-mps: Also do we want to change it to just last_checked ?

nocturnalastro: We might want to have multiple checks if we did we would need another field

PR #258, opened 2020-02-13
[IMP] udes_edi: Use env.ref instead of relying on doc_type_id.name

peterpratt13:


addons/udes_edi/security/edi_security.xml
```diff
@@ -14,7 +14,7 @@
             <field name="name">Hide EDI Raw Import Documents</field>
             <field name="model_id" search="[('model','=','edi.document')]" model="ir.model"/>
             <field name="groups" eval="[(4, ref('base.group_user'))]"/>
-            <field name="domain_force">[('doc_type_id.name', '!=', 'Raw Import')]</field>
+            <field name="domain_force">[('doc_type_id', '!=', user.env.ref('edi.raw_document_type').id)]</field>
```
ws-mps: * Use also the reference on the previous rule.
* Add comment to add a field in user if ever is needed to hide more edi document types.

PR #262, opened 2020-02-17
[ADD] udes_stock: Add reserve_pallet functionality for batches

squah-u:
[ADD] udes_stock: Add reserve_pallet functionality for batches

reserve_pallet reserves a single pallet for use in a batch and raises
an error if another batch has already reserved the pallet. Pallets are
automatically considered unreserved once the batch is done or a different
pallet is reserved for the batch.

User-story: 9391

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -62,6 +62,16 @@ class StockPickingBatch(models.Model):
         help=('Name of the batch from which this batch was derived')
     )
 
+    # This is a barcode and not a One2one to allow pallets that aren't
+    # in the system yet (because they are empty) to be reserved.
+    u_last_reserved_pallet_name = fields.Char(
+        string='Last Pallet Used',
+        index=True,
+        help="Barcode of the last pallet used for this batch. "
+             "If the batch is in progress, indicates the pallet currently in "
+             "use.",
+    )
+
     @api.depends('picking_ids',
                  'picking_ids.u_location_category_id')
     @api.one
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -986,6 +996,32 @@ def mark_as_todo(self):
...
+        batch.
+        """
+        PickingBatch = self.env['stock.picking.batch']
+
+        self.ensure_one()
+
+        conflicting_batch = PickingBatch.search([
+            ('u_last_reserved_pallet_name', '=', pallet_name),
+            ('state', '=', 'in_progress'),
+        ])
```
ws-mps: You can add and id != self.id to avoid to find self.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -986,6 +996,32 @@ def mark_as_todo(self):
...
+        if conflicting_batch and conflicting_batch != self:
+            raise ValidationError(
+                _("This pallet is already being used for batch %s.") %
+                conflicting_batch[0].name)
+
+        self.write({ 'u_last_reserved_pallet_name': pallet_name })
```
ws-mps: If you don't do the previous change, you can avoid this write if self == conflicting_batch.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -62,6 +62,16 @@ class StockPickingBatch(models.Model):
         help=('Name of the batch from which this batch was derived')
     )
 
+    # This is a barcode and not a One2one to allow pallets that aren't
+    # in the system yet (because they are empty) to be reserved.
+    u_last_reserved_pallet_name = fields.Char(
```
ws-mps: Do we want it readonly?

squah-u: It's now visible only for Trusted Users on the UI.

addons/udes_stock/models/stock_picking_batch.py
squah-u: Fixed!

PR #263, opened 2020-02-19
[IMP] udes_stock: Override _set_u_result_parent_package_id

unirho:
Overriden _set_u_result_parent_package_id to not add parent package
if the u_target_storage_format is a package/pallet of packages,
this facilitates correctly storing as packages in pick face from
returns putaway/internal transfer.

User-story: 9570

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -10,14 +10,21 @@ from ..common import check_many2one_validity
 from . import common
 
 import logging
+
 _logger = logging.getLogger(__name__)
 
 import random
 import time
 
 from psycopg2 import OperationalError, errorcodes
-PG_CONCURRENCY_ERRORS_TO_RETRY = (errorcodes.LOCK_NOT_AVAILABLE, errorcodes.SERIALIZATION_FAILURE, errorcodes.DEADLOCK_DETECTED)
+
+PG_CONCURRENCY_ERRORS_TO_RETRY = (
+    errorcodes.LOCK_NOT_AVAILABLE,
+    errorcodes.SERIALIZATION_FAILURE,
+    errorcodes.DEADLOCK_DETECTED,
+)
 MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
+
 
 def _update_move_lines_and_log_swap(move_lines, packs, other_pack):
     """ Set the package ids of the specified move lines to the one
@@ -31,7 +38,7 @@ def _update_move_lines_and_log_swap(move_lines, packs, other_pack):
     values = {
         "package_id": other_pack.id,
         "result_package_id": other_pack.id,
-        "lot_id": lots.id if len(lots) == 1 else False
+        "lot_id": lots.id if len(lots) == 1 else False,
     }
     move_lines.with_context(bypass_reservation_update=True).write(values)
     msg_args = (", ".join(packs.mapped("name")), other_pack.name)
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -1244,17 +1244,23 @@ def unlink_empty(self):
...
-        if package_move_lines:
-            package_move_lines.write({'u_result_parent_package_id': False})
+
+    def _set_u_result_parent_package_id(self):
+        """ Override function to avoid setting result_package_id when user
+        the storage format is package/pallet of products"""
+        package_pickings = self \
+            .filtered(
+                lambda p: p.picking_type_id.u_target_storage_format in
+                ('package' or 'pallet_packages')
```
ws-mps: ('package', 'pallet_packages')

ws-mps: In fact, this should be pallet products not pallet packages

ws-mps: Also we don't want to set it for products. So probably do it the other way around:
`pallet_package_pickings = XXX`
and use it when calling super.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1244,17 +1244,22 @@ def unlink_empty(self):
...
         super(StockPicking, pickings)._check_entire_pack()
-        package_move_lines = pickings \
-            .filtered(lambda p: p.picking_type_id.u_target_storage_format == 'package') \
-            .mapped("move_line_ids")
-        if package_move_lines:
-            package_move_lines.write({'u_result_parent_package_id': False})
+
+    def _set_u_result_parent_package_id(self):
+        """ Override function to avoid setting result_package_id when user
+        the storage format is package/pallet of products"""
```
ws-mps: products/package/pallet of products

addons/udes_stock/tests/test_picking.py
```diff
@@ -16,6 +16,30 @@ def setUpClass(cls):
                                               confirm=True)
         cls.SudoPicking = Picking.sudo(cls.inbound_user)
         cls.test_picking = cls.test_picking.sudo(cls.inbound_user)
+        cls.tangerine_lot = cls.create_lot(cls.tangerine.id, 1)
+
+    def picksByStorageType(self, storage_type=False):
```
nocturnalastro: snakecase

addons/udes_stock/tests/test_picking.py
```diff
@@ -16,6 +16,30 @@ def setUpClass(cls):
...
+        """
+            Generate and return picks and pallets for different storage formats
+        """
+        Package = self.env['stock.quant.package']
+        self.picking_type_in.u_target_storage_format = storage_type
```
nocturnalastro: do this in the test then you don't need to pass the parameter

addons/udes_stock/tests/test_picking.py
```diff
@@ -16,6 +16,30 @@ def setUpClass(cls):
...
+        mummy_pallet = Package.get_package('mummy_pallet', create=True)
```
nocturnalastro: Move these into setup.

addons/udes_stock/tests/test_picking.py
```diff
@@ -102,3 +126,16 @@ def test08_related_pickings(self):
...
+    def test09_only_pallets_of_packages_have_parent_package(self):
+        """
+            Test that only pallets of packages have a parent package added by
+           _check_entire_pack/_set_u_result_parent_package_id
+        """
+
+        pallet, pick = self.picksByStorageType("pallet_packages")
+        self.assertEqual(pallet, pick.move_line_ids.u_result_parent_package_id)
+
+        for storage_type in ("product", "pallet_products", "package"):
```
nocturnalastro: these should probably be different tests.

PR #269, opened 2020-03-18
[IMP] udes_security: Add flag to restrict a user to a single session

joe-gb:
u_restrict_to_single_session flag added which restricts a user to a
single session. Upon login all active sessions for a user are deleted.
(story/9772)

### DIFF
--- a/addons/udes_security/models/res_users.py
+++ b/addons/udes_security/models/res_users.py
@@ -1,14 +1,35 @@
 # -*- coding: utf-8 -*-
 import logging
 
-from odoo import api, models, SUPERUSER_ID
+from odoo import api, models, fields, SUPERUSER_ID
 from odoo.exceptions import AccessError
 from odoo.tools.translate import _
+from odoo.http import root, request
 
+_store = root.session_store
 _logger = logging.getLogger(__name__)
+
 
 class Users(models.Model):
     _inherit = 'res.users'
+
+    u_restrict_to_single_session = fields.Boolean(
+        string="Restrict user account to single login session",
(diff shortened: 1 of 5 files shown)


addons/udes_security/models/res_users.py
```diff
@@ -1,15 +1,36 @@
...
 
+    u_restrict_to_single_session = fields.Boolean(
+        string="Whether a user is restricted to having a single login session.",
+        default=False
+    )
+
+    @classmethod
+    def authenticate(cls, db, *args, **kwargs):
+        """ Override to clear all previous sessions if authenticated """
+        sessions = {sid: _store.get(sid) for sid in _store.list()}
```
kenvac: Can we move this inside if condition?

joe-gb: thank you! updated :)

addons/udes_security/tests/test_single_session.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def restrict_multiple_sessions(self, restrict):
+        with self.cursor() as cr:
+            env = self.env(cr)
+            env["res.users"].browse(self.user_1.id).u_restrict_to_single_session = restrict
+
+    def assertSessionValid(self, sids):
+        store_list = _store.list()
+        for sid in sids:
+            self.assertTrue((sid in store_list))
```
nocturnalastro: why double brackets?

joe-gb: good spot, fixed

PR #274, opened 2020-04-01
[ADD] udes_security: Added 'debug user' group

apatrick88:
Added a 'debug user' group that can only be added to by the main admin user, similarly to the 'trusted user' group. Only users in the 'debug user' group can user debug mode. If a user outside of this group tries to use any form of debug mode they will be redirected to a version of the page without debug mode enabled.

Story/9774

### DIFF
--- a/addons/udes_security/controllers/redirect.py
+++ b/addons/udes_security/controllers/redirect.py
@@ -1,11 +1,13 @@
 """Open redirect filter"""
 
 import logging
+import werkzeug
 from werkzeug import urls
 from werkzeug.exceptions import BadRequest
 from odoo import http
 from odoo.tools.translate import _
 from odoo.addons.web.controllers import main
+from odoo.http import request
 
 _logger = logging.getLogger(__name__)
 
@@ -50,7 +52,12 @@ class Home(main.Home):
     @http.route()
     @no_absolute_redirect
     def web_client(self, *args, **kwargs):
-        """Prevent absolute redirects on /web"""
+        """Prevent absolute redirects on /web/login,
+           also redirects requests from users without debug rights
+           to non-debug session"""
+        if request.debug and not request.env.user.browse(request.session.uid)\
+            .has_group('udes_security.group_debug_user'):
+                return werkzeug.utils.redirect('/web')
         return super().web_client(*args, **kwargs)
 
     @http.route()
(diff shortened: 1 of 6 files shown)


addons/udes_security/data/res_users.xml
```diff
@@ -2,6 +2,7 @@
   <data noupdate="1">
     <record id="base.user_root" model="res.users">
       <field name="groups_id" eval="[(4, ref('udes_security.group_trusted_user'))]"/>
+      <field name="groups_id" eval="[(4, ref('udes_security.group_debug_user'))]"/>
```
ws-mps, line 5: This line can be merged with previous one: `[(4, XXX), (4, YYY)]`

addons/udes_security/models/res_groups.py
```diff
@@ -13,9 +13,11 @@ class Groups(models.Model):
     @api.multi
     def write(self, values):
         group_trusted_user = self.env.ref('udes_security.group_trusted_user')
+        group_debug_user = self.env.ref('udes_security.group_debug_user')
```
ws-mps, line 16: I would suggest to have list of groups in a constant of the file and loop over it.

addons/udes_security/models/res_users.py
```diff
@@ -41,21 +41,24 @@ def authenticate(cls, db, *args, **kwargs):
...
-        self._check_trusted_user_grant(values)
+        self._check_user_group_grant(values)
 
         res = super(Users, self).write(values)
         return res
 
-    def _check_trusted_user_grant(self, values):
-        group_trusted_user = self.env.ref("udes_security.group_trusted_user")
+    def _check_user_group_grant(self, values):
+        group_trusted_user = self.env.ref('udes_security.group_trusted_user')
```
ws-mps, line 57: This can import the list from res_groups.py and also loop over it.

addons/udes_security/controllers/redirect.py
```diff
@@ -51,6 +53,9 @@ class Home(main.Home):
     @no_absolute_redirect
     def web_client(self, *args, **kwargs):
         """Prevent absolute redirects on /web"""
```
squah-u, line 55: This docstring could do with an update

squah-u: This method has two purposes now, not just one! It still prevents absolute redirects

ws-mps:
LGTM

PR #278, opened 2020-04-07
{IMP] udes_load_testing: recommend running _report in tearDown

kdwyer:
User-story/9878
Task/9985

Calling `_report()` in a test method means that it gets counted as a test,
and triggers the class's `setUp()` method.  Recommend that it be called
in the class's `tearDown` method instead.


ws-mps:
LGTM

PR #280, opened 2020-04-08
[ADD] udes_stock: Add indexes for stock.move.line

squah-u:
Task: 9993
User-story: 9881

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 from itertools import groupby
-from odoo import api, fields, models, _
+from odoo import api, fields, models, tools, _
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.float_utils import float_compare, float_round
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -1038,6 +1038,25 @@ def _prevent_write_after_done(self):
         if done_lines:
             raise ValidationError(_("Cannot update move lines that are already 'done'."))
 
+    @api.model_cr
+    def init(self):
+        """ Creates indexes for _check_resultant_package_level
+        """
+        super(StockMoveLine, self).init()
+
+        tools.create_index(
```
kdwyer, line 1047: :+1: I didn't know about these tools.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1038,6 +1038,25 @@ def _prevent_write_after_done(self):
...
+            self._cr,
+            "stock_move_line_state_result_package_id_index",
+            self._table,
+            ["state", "result_package_id"],
```
kdwyer, line 1051: If we're theoretically targeting Python3.5, I think the trialing comma on the final arguemnt isn't valid syntax.

addons/udes_stock/models/stock_move_line.py
squah-u: It's very much copied from Miquel's PR

addons/udes_stock/models/stock_move_line.py
squah-u: Here I was following the existing style for this file (eg. line 822)

PR #289, opened 2020-04-28
User-story: 10074 - Confirm orders in batches and committing

ws-mps:
* Add confirm_orders() method to sale order so that they are confirmed in batches and committing after each batch to give visibility.

* Add cron job to execute confirm orders but disabled.

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -5,6 +5,16 @@ from odoo import api, fields, models
 from odoo.addons.udes_stock.models import common
 import logging
 from datetime import timedelta, date
+import time
+
+from psycopg2 import OperationalError, errorcodes
+
+PG_CONCURRENCY_ERRORS_TO_RETRY = (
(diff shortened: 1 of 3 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -267,6 +277,55 @@ def get_current_demand(self, products=None):
...
+        to_confirm = self.search([('state', '=', 'draft')])
+
+        for _, batch in to_confirm.batched(size=1000):
+            tries = 0
+            while True:
+                try:
+                    with self.env.cr.savepoint():
+                        batch.action_confirm()
+                        break
+                except OperationalError as e:
```
nocturnalastro, line 293: why do a while loop as mat retries is a fixed value and we break on success anyway?

ws-mps: Just to mimic Odoo's retry behaviour, same as we did for stock reservation.

nocturnalastro: Fair enough

PR #292, opened 2020-05-13
[ADD] udes_stock: Add helper functions to retry on concurrrency errors

squah-u:
User-story: 10136

### DIFF
--- a/addons/udes_stock/tools/retry.py
+++ b/addons/udes_stock/tools/retry.py
@@ -1,0 +1,71 @@
+from functools import wraps
+import logging
+import time
+from psycopg2 import OperationalError, errorcodes
+
+_logger = logging.getLogger(__name__)
+
+PG_CONCURRENCY_ERRORS_TO_RETRY = (
+    errorcodes.LOCK_NOT_AVAILABLE,
+    errorcodes.SERIALIZATION_FAILURE,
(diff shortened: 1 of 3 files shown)


addons/udes_stock/tools/retry.py
```diff
@@ -0,0 +1,58 @@
...
+                # don't do this
+                raise
+            if tries >= max_retries:
+                _logger.info(
+                    "%s, maximum number of tries reached"
+                    % errorcodes.lookup(e.pgcode)
+                )
+                raise
+
+            wait_time = 1
```
apatrick88, line 47: Do we want `wait_time` as another function argumen as for `max_retries`? It'd give us more flexibility for longer running processes.

squah-u: Good idea, I'll add it

ws-mps: Good point. A parameter looks reasonable to me.

In fact, wait time usually multiplies by 2 after every retry, so after each retry waits a bit longer. So it would be good to have another parameter like "delaying factor" so by default is 2, but in some cases we might want to keep it the same (so we can use 1 as factor).

squah-u:
This code was found to not have the desired outcome. If the first attempt failed, all subsequent attempts would also fail due to how postgres transactions work.

PR #294, opened 2020-05-22
[FIX] Log message if method already exists on base module.

peterpratt13:
If the method already exists on the base model then log the message
instead of raising an import error.

User-story: 9138

### DIFF
--- a/addons/udes_common/models/models.py
+++ b/addons/udes_common/models/models.py
@@ -1,10 +1,13 @@
 """Base model enhancements"""
 
+import logging
 import itertools
 from operator import itemgetter
 from .. import tools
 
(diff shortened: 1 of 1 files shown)


addons/udes_common/models/models.py
```diff
@@ -15,7 +18,9 @@ def add_if_not_exists(cls):
 
     def wrapper(func):
         # pylint: disable=missing-docstring
-        if not hasattr(cls, func.__name__):
+        if hasattr(cls, func.__name__):
```
nocturnalastro, line 21: See https://github.com/unipartdigital/odoo-edi/pull/68#discussion_r430245931

PR #295, opened 2020-05-22
[9138] [FIX]: Fix the ordering of operation types in the UI

peterpratt13:
User-story: 9138


addons/udes_stock/data/picking_types.xml
```diff
@@ -69,7 +69,7 @@
     <!-- Create "Putaway" picking type -->
     <record id="picking_type_putaway" model="stock.picking.type">
       <field name="name">Putaway</field>
-      <field name="sequence" eval="13"/>
+      <field name="sequence" eval="11"/>
```
ws-mps, line 72: Let's keep more space between them: onbound at the 10, outbound 20, etc

PR #297, opened 2020-06-01
[ADD] udes_security: Add inactive_session_time_out_enabled system parameter

PeterC10:
Added system parameter "inactive_session_time_out_enabled" to allow the timeout check in auth_session_timeout to be disabled.

Story/9964

### DIFF
--- a/addons/udes_security/models/res_users.py
+++ b/addons/udes_security/models/res_users.py
@@ -99,3 +99,23 @@ class Users(models.Model):
                     )
                     % group.name
                 )
+
+    @api.model_cr_context
+    def _auth_timeout_enabled(self):
+        """ Pluggable method to check if session timeout is enabled """
+        IrConfigParameter = self.env["ir.config_parameter"]
+        
+        auth_timeout_enabled_parameter = IrConfigParameter.get_param(
+            "inactive_session_time_out_enabled"
+        )
+        return auth_timeout_enabled_parameter == "True"
+
+    @api.model
+    def _auth_timeout_check(self):
+        """ Override to not carry out timeout check if system parameter is disabled """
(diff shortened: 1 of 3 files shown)


addons/udes_security/models/res_users.py
```diff
@@ -99,3 +99,20 @@ def _check_is_admin(self, values, group):
                     )
                     % group.name
                 )
+
+    @api.model_cr_context
+    def _auth_timeout_enabled(self):
+        """ Pluggable method to check if session timeout is enabled """
+        auth_timeout_enabled_parameter = self.env["ir.config_parameter"].get_param("inactive_session_time_out_enabled")
+        auth_timeout_enabled = auth_timeout_enabled_parameter == "True"
+        return auth_timeout_enabled
```
williamsajr, line 108: Could directly return `auth_timeout_enabled_parameter == "True"`

PeterC10: Yeah, was thinking it might make it a bit clearer rather than returning the expression. But its a very simple expression so will amend.

PeterC10: Updated.

addons/udes_security/models/res_users.py
```diff
@@ -99,3 +99,21 @@ def _check_is_admin(self, values, group):
                     )
                     % group.name
                 )
+
+    @api.model_cr_context
+    def _auth_timeout_enabled(self):
+        """ Pluggable method to check if session timeout is enabled """
+        auth_timeout_enabled_parameter = self.env["ir.config_parameter"].get_param(
```
ws-mps, line 106: Do we want to use `IrConfigParameter` instead?

PeterC10: Yep, changed. :)

PR #300, opened 2020-06-04
[ADD] udes_sale_stock: Create index for sale order ordering

squah-u:
Task: 10004
User-story: 9977


squah-u:
Moved to story/9977a

PR #302, opened 2020-06-12
Story/9770 - Blocked file type filtering

PeterC10:
Prevent users from uploading/download executable files via a blocked file type list which is configurable by admin.

Added Blocked File Type entity which allows admin user to add file types which non-admin users will not be able to download.

Attachments matching blocked file types now hidden. Upload and Download controllers overridden to explicitly prevent blocked file types from being uploaded/downloaded.

Added override to BinaryFile/BinaryImage widgets so that the file type is checked and upload is prevented if file type is blocked.

Common executable file types added as Blocked File Types.

### DIFF
--- a/addons/udes_security/controllers/main.py
+++ b/addons/udes_security/controllers/main.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+
+import json
+import logging
+
+from odoo import http, SUPERUSER_ID, _
+from odoo.http import request
+from odoo.exceptions import UserError
+
+from odoo.addons.web.controllers.main import Binary, serialize_exception
+
+_logger = logging.getLogger(__name__)
+
+
+class BinaryExtension(Binary):
+    def _get_file_type(self, filename):
+        """Extract the file type from the filename by returning value
+           after the last '.' character"""
+        return filename.split(".")[-1].lower()
+
+    def _get_file_type_blocked(self, file_type):
+        """Return true if the file type is blocked, otherwise false"""
+        BlockedFileType = request.env["udes.blocked_file_type"].sudo()
+
+        search_args = [("name", "=", file_type)]
+        blocked_file_type_count = BlockedFileType.search_count(search_args)
+
+        return bool(blocked_file_type_count)
+
+    def _get_file_type_blocked_error_message(self, file_type):
+        """Return an error message stating that the file type has been blocked"""
+        return _("File type '%s' has been blocked by system administrator." % (file_type))
+
+    def _log_user_file_action_blocked(self, action, filename, user_id):
+        """Log a message when a user is blocked trying to upload/download a file"""
+        ResUsers = request.env["res.users"].sudo()
+
+        user = ResUsers.browse(user_id)
+
+        message = "Blocked attempt to %s file '%s' by user '%s'" % (action, filename, user.login)
+        _logger.info(message)
+
+        return True
+
+    @http.route(
+        [
+            "/web/content",
+            "/web/content/<string:xmlid>",
+            "/web/content/<string:xmlid>/<string:filename>",
+            "/web/content/<int:id>",
+            "/web/content/<int:id>/<string:filename>",
+            "/web/content/<int:id>-<string:unique>",
+            "/web/content/<int:id>-<string:unique>/<string:filename>",
+            "/web/content/<string:model>/<int:id>/<string:field>",
+            "/web/content/<string:model>/<int:id>/<string:field>/<string:filename>",
+        ],
+        type="http",
+        auth="public",
+    )
+    def content_common(
+        self,
+        xmlid=None,
+        model="ir.attachment",
+        id=None,
+        field="datas",
+        filename=None,
+        filename_field="datas_fname",
+        unique=None,
+        mimetype=None,
+        download=None,
+        data=None,
+        token=None,
+        access_token=None,
+        **kw
+    ):
+        """Override to prevent non-admin users from downloading blocked file types"""
+        user_id = request.session.uid
+
+        if user_id != SUPERUSER_ID:
+            # Check if the file type is blocked
+            if download:
+                record_filename = ""
+
+                if filename:
+                    record_filename = filename
+                elif id and model and filename_field:
+                    RecordModel = request.env[model].sudo()
+                    record = RecordModel.browse(id)
+                    record_filename = record[filename_field]
+
+                file_type = self._get_file_type(record_filename)
+                file_type_blocked = self._get_file_type_blocked(file_type)
+
+                # If the user is trying to download a blocked file type then
+                # prevent download and return an error message
+                if file_type_blocked:
+                    self._log_user_file_action_blocked("download", record_filename, user_id)
+
+                    download_error = {
+                        "message": "Unable to download file: File type blocked",
+                        "data": {"debug": self._get_file_type_blocked_error_message(file_type),},
+                    }
+
+                    return request.make_response((json.dumps(download_error)))
+
+        return super(BinaryExtension, self).content_common(
+            xmlid,
+            model,
+            id,
+            field,
+            filename,
+            filename_field,
+            unique,
+            mimetype,
+            download,
+            data,
+            token,
+            access_token,
+            **kw
+        )
+
+    @http.route("/web/binary/upload_attachment", type="http", auth="user")
+    @serialize_exception
+    def upload_attachment(self, callback, model, id, ufile):
+        """Override to prevent non-admin users from uploading blocked file types"""
+        user_id = request.session.uid
+        files = request.httprequest.files.getlist("ufile")
+
+        if user_id != SUPERUSER_ID:
+            for upload_file in files:
+                filename = upload_file.filename
+
+                if upload_file and filename:
+                    file_type = self._get_file_type(filename)
+                    file_type_blocked = self._get_file_type_blocked(file_type)
+
+                    # If the user is trying to upload a blocked file type then
+                    # prevent upload and return an error message
+                    if file_type_blocked:
+                        self._log_user_file_action_blocked("upload", filename, user_id)
+
+                        out = """<script language="javascript" type="text/javascript">
+                                    var win = window.top.window;
+                                    win.jQuery(win).trigger(%s, %s);
+                                </script>"""
+
+                        args = [{"error": self._get_file_type_blocked_error_message(file_type)}]
+
+                        return out % (json.dumps(callback), json.dumps(args))
+
+        return super(BinaryExtension, self).upload_attachment(callback, model, id, ufile)
--- a/addons/udes_security/models/blocked_file_type.py
+++ b/addons/udes_security/models/blocked_file_type.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class BlockedFileType(models.Model):
+    """
+    Blocked file type which the user is prevented from uploading or downloading
+    """
+
+    _name = "udes.blocked_file_type"
+    _description = "Blocked File Type"
+    _order = "name"
+
+    _sql_constraints = [("name_uniq", "unique (name)", "The File Type must be unique")]
+
+    name = fields.Char("File Type", required=True, index=True)
+    description = fields.Char("Description")
+    active = fields.Boolean("Active?", default=True)
+
+    def _format_name(self, name):
+        """Convert supplied name to lowercase if needed"""
+        formatted_name = name.lower().replace(".", "") if name else ""
+        return formatted_name
+
+    def _format_vals(self, vals):
+        """Format the required values in the supplied dictionary and 
+           return the updated values"""
+        formatted_vals = {}
+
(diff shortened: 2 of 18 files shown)


addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+            data,
+            token,
+            access_token,
+            **kw
+        )
+
+    @http.route("/web/binary/upload_attachment", type="http", auth="user")
+    @serialize_exception
+    def upload_attachment(self, callback, model, id, ufile):
+
```
ws-mps, line 106: Remove empty line to be consistent with `content_common`.

Do we need docstrings?

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+        BlockedFileType = request.env["udes.blocked_file_type"].sudo()
+
+        files = request.httprequest.files.getlist("ufile")
+
+        if request.session.uid != SUPERUSER_ID or True:
```
ws-mps, line 111: `or True`?

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+                    record_filename = record[filename_field]
+                    file_type = self._get_file_type(record_filename)
+
+                blocked_file_type_count = BlockedFileType.search_count([("name", "=", file_type)])
+
+                # If the user is trying to download a blocked file type then 
+                # prevent download and return an error message
+                if blocked_file_type_count:
+                    _logger.info(
+                        "Blocked attempt to download file %s" % (filename or record_filename)
```
ws-mps, line 74: I think it could be useful to know the user too.

Can we unify `filename` and `record_filename` ?

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+
+                    if blocked_file_type_count:
+                        out = """<script language="javascript" type="text/javascript">
+                                    var win = window.top.window;
+                                    win.jQuery(win).trigger(%s, %s);
+                                </script>"""
+
+                        args = [{"error": _("File type blocked by system administrator")}]
+                        _logger.info(
+                            "Blocked attempt to upload file %s" % ufile.filename
```
ws-mps, line 128: Same as above.

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+    @serialize_exception
+    def upload_attachment(self, callback, model, id, ufile):
+
+        BlockedFileType = request.env["udes.blocked_file_type"].sudo()
+
+        files = request.httprequest.files.getlist("ufile")
+
+        if request.session.uid != SUPERUSER_ID or True:
+            for ufile in files:
+                if ufile and ufile.filename:
```
ws-mps, line 113: `ufile.filename` is used 3 times, can we put it into a var?

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+    @http.route("/web/binary/upload_attachment", type="http", auth="user")
+    @serialize_exception
+    def upload_attachment(self, callback, model, id, ufile):
+
+        BlockedFileType = request.env["udes.blocked_file_type"].sudo()
+
+        files = request.httprequest.files.getlist("ufile")
+
+        if request.session.uid != SUPERUSER_ID or True:
+            for ufile in files:
```
ws-mps, line 112: This `ufile` is shadowing the function parameter, dangerous if we ever get to line 133.

addons/udes_security/controllers/main.py
PeterC10: Whoops, was most likely checking something out as admin user. Will remove.

addons/udes_security/controllers/main.py
PeterC10: Yeah should have docstring, will go amend.

addons/udes_security/controllers/main.py
PeterC10: Yeah, it's how Odoo do it in the upload_attachment function in web module hence why I copied that approach. But it is unnecessarily dangerous so will go rename that.

addons/udes_security/controllers/main.py
PeterC10: The user is informed below when ```download_error``` is returned which products a notification in the web client. If we raise an exception here the user will just be given a 500 error.

ws-mps: Haven't explained properly :) I mean also add the user that tried to download:
```Blocked attemp to download file XXX by user YYY```

File: addons/udes_security/models/blocked_file_type.py [hunk 7]
```diff
@@ -0,0 +1,125 @@
...
+
+    name = fields.Char("File Type", required=True, index=True)
+    description = fields.Char("Description")
+    active = fields.Boolean("Active?", default=True)
+
+    def _format_name(self, name):
+        """Convert supplied name to lowercase if needed"""
+
+        formatted_name = name or ""
+        if formatted_name:
```
nocturnalastro, line 26: is there anypoint in the ifs why not just do the lower and replace if it doesn't change it doesn't matter

PeterC10: I suppose not really, will go change this.

addons/udes_security/models/blocked_file_type.py
```diff
@@ -0,0 +1,125 @@
...
+            blocked_file_type_active = blocked_file_type.active
+
+            # If a blocked file type record is being deleted then it should be treated as inactive
+            if deleted:
+                blocked_file_type_active = False
+
+            attachment_search_args = [
+                ("datas_file_type", "=", blocked_file_type.name),
+                "|",
+                ("active", "=", True),
```
ws-mps, line 62: Is this to find any record? Odoo supports a context variable to the search `active_test=False` that will avoid to filter by active field.

PeterC10: Yeah it is. Ah that's a lot nicer, so I can do this?

```IrAttachment.with_context(active_test=False).search(attachment_search_args)```

File: addons/udes_security/models/blocked_file_type.py [hunk 9]
```diff
@@ -0,0 +1,125 @@
...
+                ("active", "=", False),
+            ]
+
+            attachments = IrAttachment.search(attachment_search_args)
+
+            attachment_vals = {"active": not blocked_file_type_active}
+            attachments.with_context(skip_active_check=True).write(attachment_vals)
```
ws-mps, line 69: Instead of doing write for each blocked type, we could store two recordsets: one for active=False and one for active=True and do two writes outside the loop.

addons/udes_security/controllers/main.py
PeterC10: Gotcha. Think that is more down to me reading too fast!

Good idea, will go add this. :)

addons/udes_security/models/blocked_file_type.py
```diff
@@ -0,0 +1,125 @@
...
+        attachments = IrAttachment.search(attachment_search_args)
+        attachments.with_context(skip_active_check=True).write({"active": True})
+
+        return True
+
+    @api.model
+    def create(self, vals):
+        """Override to ensure name is formatted correctly
+           and relevant attachments are set to active/inactive"""
+
```
ws-mps, line 91: There is no consistency between files about having empty line after docstring, please revise.

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,135 @@
...
+                if ufile and ufile.filename:
+                    file_type = self._get_file_type(ufile.filename)
+
+                    blocked_file_type_count = BlockedFileType.search_count(
+                        [("name", "=", file_type)]
```
ws-mps, line 117: In other files this is moved into a variable. Also this search_count is quite common, do we want a function for it?

addons/udes_security/controllers/main.py
PeterC10: Have removed this.

addons/udes_security/controllers/main.py
PeterC10: Have now refactored ```content_common``` and ```upload_attachment``` functions. Function ```_get_file_type_blocked``` is now used and returns `True` or `False` depending on whether the file type is blocked.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have now removed blank lines under docstrings.

addons/udes_security/controllers/main.py
PeterC10: Have now added missing docstrings.

addons/udes_security/controllers/main.py
PeterC10: Have now created a ```_log_user_file_action_blocked``` function to handle this, and that includes the user login in the message.

Have also unified ```filename``` and ```record_filename```.

addons/udes_security/controllers/main.py
PeterC10: Have now done this.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have now simplified this and removed the unnecessary if statements. It now uses one inline if which is used just incase ```name``` is ```False```.

addons/udes_security/controllers/main.py
PeterC10: This has now been done.

addons/udes_security/controllers/main.py
PeterC10: Have now changed ```ufile``` to ```upload_file``` to avoid this.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have now implemented this.

addons/udes_security/controllers/main.py
```diff
@@ -0,0 +1,152 @@
...
+        access_token=None,
+        **kw
+    ):
+        """Override to prevent non-admin users from downloading blocked file types"""
+        user_id = request.session.uid
+
+        if user_id != SUPERUSER_ID:
+            # Check if the file type is blocked
+            if download:
+                file_type = ""
```
ws-mps, line 82: This line is not needed anymore.

addons/udes_security/models/blocked_file_type.py
```diff
@@ -0,0 +1,119 @@
...
+            name = self._format_name(name)
+            formatted_vals["name"] = name
+
+        return formatted_vals
+
+    def _update_attachments_active_status(self, deleted=False):
+        """Set active field on attachments that match the supplied blocked file types"""
+        IrAttachment = self.env["ir.attachment"]
+
+        attachments_set_active = IrAttachment.browse()
```
ws-mps, line 43: nit: What about calling them active_attachments and inactive_attachments ? reads easier to me.

addons/udes_security/models/blocked_file_type.py
```diff
@@ -0,0 +1,119 @@
...
+        attachments_set_inactive = IrAttachment.browse()
+
+        for blocked_file_type in self:
+            blocked_file_type_active = blocked_file_type.active
```
ws-mps, line 47: nit: What about renaming it to is_blocked_file_type ?

addons/udes_security/models/blocked_file_type.py
```diff
@@ -0,0 +1,119 @@
...
+            if deleted:
+                blocked_file_type_active = False
+
+            attachment_search_args = [("datas_file_type", "=", blocked_file_type.name)]
+
+            attachments = IrAttachment.with_context(active_test=False).search(
+                attachment_search_args
+            )
+
+            set_attachments_active = not blocked_file_type_active
```
ws-mps, line 59: This is not really needed, we can just check `not blocked_file_type_active` at the if condition. (or `not is_blocked_file_type` if you rename as per previous comment).

addons/udes_security/models/blocked_file_type.py
PeterC10: Was on the fence about that, as ```active_attachments``` will most likely be assigned inactive attachments, but from the method it should be clear what we are doing with the recordsets so will rename. :)

ws-mps: You are right. I would then suggest `attachments_to_set_active`

addons/udes_security/controllers/main.py
PeterC10: Have removed this.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have now renamed these variables.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have now renamed this.

addons/udes_security/models/blocked_file_type.py
PeterC10: Have updated this so it just checks ```is_blocked_file_type``` directly. Have removed ```set_attachments_active```.

PR #307, opened 2020-06-25
Story/9859 - Migrate Stock pick refactoring functionality from UDES 11

PeterC10:
This story covers the migration of existing Stock pick refactoring from UDES 11 into UDES 13. One key difference is that rather than going into udes_stock, this functionality will go into its own udes_stock_functionality module.

A separate story will be raised to cover further improvements for this module (and refactoring to the refactoring functionality...!)


addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+        help="First pickings in the chain",
+    )
+    u_created_back_orders = fields.One2many(
+        "stock.picking",
+        string="Created Back Orders",
+        compute="_compute_related_picking_ids",
+        help="Created Back Orders",
+    )
+
+    def find_empty_pickings_to_unlink(self):
```
williamsajr, line 80: It may be better to move this function to udes_stock as that is where ```u_auto_unlink_empty``` is defined and is independent of refactoring

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+            pts = PickingType.search([("u_auto_unlink_empty", "=", False)])
+            if pts:
+                domain.append(("picking_type_id", "not in", pts.ids))
+        while True:
+            records = self.search(domain, limit=1000)
+            if not records:
+                break
+            yield records
+
+    def unlink_empty(self):
```
williamsajr, line 106: It may be better to move this function to udes_stock as that is where ```u_auto_unlink_empty``` is defined and is independent of refactoring

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,659 @@
...
+            self.apple.id, self.test_stock_location_01.id, 10, package_id=apple_pallet.id
+        )
+        apple_move = self.create_move(self.apple, 10, self.picking)
+
+        self.picking.action_assign()
+
+        apple_pick = apple_move.picking_id
+        self.assertEqual(self.picking, apple_pick)  # Check picking reuse
+        self.assertEqual(apple_pick.state, "assigned")
+        self.assertEqual(apple_pick.location_id.id, self.test_stock_location_01.id)
```
peterpratt13, line 311: Don't need the id's

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+
+        # add extra domain if there is any
+        if extra_domain:
+            domain.extend(extra_domain)
+
+        pickings = Picking.search(domain, order=order)
+
+        return pickings
+
+    def action_assign(self):
```
williamsajr, line 353: I'm wondering if these action functions should also be moved to udes_stock? We may want to remove junk for other situations as well

williamsajr: Do we also need a function for action_done to unlink empty? It seems to be done at the end of the UDES11 method

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,659 @@
...
+        self.assertEqual(pick2_move_line.package_id, cherry_pallet2)
+        self.assertEqual(pick2_move_line.result_package_id, cherry_pallet2)
+        self.assertEqual(pick2_move_line.product_id, self.cherry)
+
+        # Check that the original picking has been reused
+        self.assertTrue(self.picking.exists())
+
+    def test_split_pallet_different_products(self):
+        """Reserve self.picking with a pallet containing two different products
+        and check it splits correctly when reserved."""
```
peterpratt13, line 145: Nit: Other doc strings you ended with """ on a new line :)

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,437 @@
...
+        (will error out if not). Therefore the refactor at confirm should avoid
+        splitting/creating moves. Luckily this is unlikely to be an issue as
+        it makes little sense to split moves at this stage, before they have
+        stock reserved against them.
+        """
+        res = super(StockMove, self)._action_confirm(*args, **kwargs)
+        post_refactor_moves = res._action_refactor(stage="confirm")
+
+        if post_refactor_moves != res:
+            raise UserError(
```
williamsajr, line 130: Is this error still valid?

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,437 @@
...
+            values["partner_id"] = partners.id
+
+        dates_done = list(set(moves.mapped("date")))
+        if len(dates_done) == 1:
+            values["date_done"] = dates_done[0]
+
+        return values
+
+    def refactor_action_group_by_move_key(self):
+        """
```
williamsajr, line 259: Inconsistent docstring

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,437 @@
...
+
+        empty_picks = pickings.filtered(lambda p: len(p.move_lines) == 0)
+        if empty_picks:
+            _logger.info(_("Marking empty picks after splitting for clean up."))
+            empty_picks.write({"u_mark": False, "is_locked": True})
+
+        return self
+
+    def refactor_action_group_by_move_line_key(self):
+        """
```
williamsajr, line 312: Inconsistent docstring

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+        order = None
+
+        if states is None:
+            states = ["draft", "cancel", "waiting", "confirmed", "assigned", "done"]
+
+        warehouse = Users.get_user_warehouse()
+        if picking_type_ids is None:
+            picking_type_ids = warehouse.get_picking_types().ids
+
+        if self:
```
peterpratt13, line 298: Lots of ifs statements here, that look to just overwrite the domain that may have previously been set if someone is added lots of kwargs. If they should only set one, or a subset can we create a mapping to shrink it?

addons/udes_stock_refactoring/models/stock_picking.py
williamsajr: Also inconsistent docstring format

addons/udes_stock_refactoring/models/stock_picking.py
williamsajr: Also inconsistent docstring format

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+                raise ValidationError(
+                    _("Trying to unlink non empty pickings: " "%s" % non_empty_pickings.ids)
+                )
+            to_unlink.unlink()
+
+        return self - records
+
+    @api.model
+    def _new_picking_for_group(self, group_key, moves, **kwargs):
+        """
```
williamsajr, line 129: inconsistent docstring format

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+        location_id=None,
+        product_id=None,
+        backorder_id=None,
+        result_package_id=None,
+        picking_priorities=None,
+        picking_ids=None,
+        batch_id=None,
+        extra_domain=None,
+    ):
+
```
williamsajr, line 230: Super pedantic: blank line

williamsajr: + blank space after """

addons/udes_stock_refactoring/models/stock_picking.py
williamsajr: Also inconsistent docstring format

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,659 @@
+# -*- coding: utf-8 -*-
+
+from addons.udes_stock.tests import common
+
+
+class TestAssignSplitting(common.BaseUDES):
+    def setUp(self):
+        """
```
williamsajr, line 8: inconsistent docstring format

addons/udes_stock_refactoring/models/stock_quant_package.py
```diff
@@ -0,0 +1,49 @@
...
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+class StockQuantPackage(models.Model):
+    _inherit = "stock.quant.package"
+
+    def get_package(self, package_identifier, create=False, no_results=False):
```
peterpratt13, line 14: Looks like get_package, get_group are very similar. Might be useful to consider a mixin class for this, like mixin_stock_model in udes_stock.

addons/udes_stock_refactoring/data/stock_picking_types.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+     <!--Do we want to to change the u_move_line_key_format to {package_id.id}?  -->
```
apatrick88, line 4: Perhaps rename this file to 'picking_types.xml' instead? Fits with current convention in UDES 11.

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+    _inherit = "stock.picking"
+
+    @api.depends(
+        "move_lines",
+        "move_lines.move_orig_ids",
+        "move_lines.move_dest_ids",
+        "move_lines.move_orig_ids.picking_id",
+        "move_lines.move_dest_ids.picking_id",
+    )
+    def _compute_related_picking_ids(self):
```
ws-mps, line 23: This should go into another module.

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def _get_package_search_domain(self, package):
+        """Generate the domain for searching pickings of a package"""
+        return [
+            "|",
+            ("move_line_ids.package_id", "child_of", package.id),
+            ("move_line_ids.result_package_id", "child_of", package.id),
+        ]
+
+    def get_pickings(
```
ws-mps, line 214: Don't think this is really necessary.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,437 @@
...
+        return them as part of the set of completed moves.
+        """
+        done_moves = super(StockMove, self)._action_done()
+
+        post_refactor_done_moves = done_moves._action_refactor(stage="validate")
+
+        post_refactor_done_moves.push_from_drop()
+        return post_refactor_done_moves
+
+    def push_from_drop(self):
```
ws-mps, line 164: This should go into another module.

addons/udes_stock_refactoring/models/stock_quant_package.py
ws-mps: This module should depend on udes_stock which already has a function to do this. Although looks like we only need this function for get_pickings which is not really needed.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Was only being used by ```_compute_related_picking_ids``` which needs to go into another module anyway so will remove this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Into ```udes_stock``` or another stock addon module?

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Into ```udes_stock``` or another stock addon module?

addons/udes_stock_refactoring/data/stock_picking_types.xml
PeterC10: Yep will rename this.

addons/udes_stock_refactoring/models/stock_move.py
ws-mps: A new module. But not needed for the refactoring code.

PeterC10:
I've carried out a second run for the stock_refactoring module. A lot of unneeded code/functionality has been removed, with some remaining code cleaned up and doc strings should now have a consistent format.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: This has now been removed.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: This has now been removed.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: This has now been removed.

addons/udes_stock_refactoring/data/stock_picking_types.xml
PeterC10: This has now been renamed.

addons/udes_stock_refactoring/models/stock_quant_package.py
PeterC10: Yeah this method is no longer needed and has now been removed.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: This giant function is no longer needed and has now been removed. :)

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: This giant function is no longer needed and has now been removed. :)

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: I've gone back and forth on this a few times. For now I think all unlinking of empty pickings should be done in the stock refactoring module, as right now that is the only time you'd make a picking empty/want to auto delete empty pickings.

PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now updated the docstrings.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Good spot, have changed this and a few other occasions this happens in these unit tests.

PeterC10:
Regarding the removal of empty pickings:

1. At one point it seemed like we wouldn't need this at all, as most often when refactoring the original picking gets preserved. However when confirming a batch of pickings it is possible for the original picking to be left empty and in this case it should get removed so I have left this functionality in for now as it should match UDES 11.

2. I've switched back and forth a few times about whether the empty picking checks should go in the stock or stock_refactoring module. Right now I think it should stay in stock_refactoring as that is the only time we're explicitly setting a picking to empty and also want to delete it afterwards. But can be argued it is pretty generic functionality that a future stock addon module may need to make use of, so happy to move this again if needed.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: I believe it is as the `mrp` module is still part of core Odoo in 13.

addons/udes_stock/models/stock_move.py
```diff
@@ -1,10 +1,56 @@
 # -*- coding: utf-8 -*-
-from odoo import models, _
+from odoo import models, api, _
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    def _prepare_move_line(self, moves, qty, **kwargs):
```
ws-mps, line 8: Only one qty for a list of moves? shouldn't this be a list of quantities with the same length as moves recordset?

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,647 @@
...
+        simultaneously and check they result in one picking with two moves.
+        """
+        pallet = self.create_package()
+
+        self.create_quant(
+            self.elderberry.id, self.test_stock_location_01.id, 5, package_id=pallet.id
+        )
+        self.create_quant(
+            self.elderberry.id, self.test_stock_location_01.id, 10, package_id=pallet.id
+        )
```
peterpratt13, line 218: You can probably move all the create quants into a setup method, and just reuse them for each test, rather than continually having to repeat them.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,647 @@
...
+
+        cherry_pallet_move_lines = MoveLine.search(
+            [("product_id.name", "=", "Test product Cherry")]
+        )
+
+        cherry_pallet1_move_line = False
+        cherry_pallet2_move_line = False
+
+        for move_line in cherry_pallet_move_lines:
+            if move_line.package_id.name == cherry_pallet1.name:
```
peterpratt13, line 110: Do we need the .name here?

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,647 @@
...
+                cherry_pallet1_move_line = move_line
+            elif move_line.package_id.name == cherry_pallet2.name:
+                cherry_pallet2_move_line = move_line
+
+        self.assertTrue(cherry_pallet1_move_line)
+        self.assertTrue(cherry_pallet2_move_line)
+
```
peterpratt13, line 117: Do you also want to check there are 2 for completeness?

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,647 @@
...
+        self.create_quant(
+            self.cherry.id, self.test_stock_location_01.id, 10, package_id=cherry_pallet2.id
+        )
+
+        products_info = [{"product": self.cherry, "qty": 20}]
+        self.create_move(self.picking, products_info)
+
+        self.picking.action_assign()
+
+        cherry_pallet_move_lines = MoveLine.search(
```
peterpratt13, line 102: Can we not just get them from the picking?

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Yep hadn't done too much with the unit tests since moving them from UDES 11 but will have a go at simplifying/improving them.

addons/udes_stock/models/stock_move.py
PeterC10: Yeah good point, is way too restrictive with just the one quantity for n moves. Is a list of quantities safe enough though (if safe is the right word)? We could just have a ```moves_info``` dict with the move as the key and the quantity as the value.

apatrick88: We'd definitely want to be able link the quantity back to the required move, although we'd still want to make sure that any calls of `_prepare_move_line` are not overly complicated. Perhaps `_prepare_move_line` and `_prepare_move_lines`?

PeterC10: Yeah think that makes sense. That way you have a quick option for a single move line or a slightly more complex way (but still far quicker than creating move lines manually) of creating a batch of move lines.

PeterC10: Have now added a new `_prepare_move_lines` helper method which itself calls `_prepare_move_line` for each move and quantity supplied.

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,176 @@
...
+                values = {"group_id": group.id}
+                values.update(kwargs)
+
+                current_picking.write(values)
+                picking = current_picking
+
+        if not picking or len(picking) > 1:
+            # There was no suitable picking to reuse.
+            # Create a new picking.
+            values = {
```
nocturnalastro, line 92: should we get this default set of values via a method?

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,176 @@
...
+        return res
+
+    def action_done(self):
+        """Override action_done to unlink empty pickings if needed"""
+        res = super(StockPicking, self).action_done()
+        if self:
+            self.unlink_empty()
+        return res
+
+    def unlink_empty(self):
```
nocturnalastro, line 157: Do we want to do this for the picking type/s we're refactoring or and original set of picking because this could cause a lot of serialization errors.

addons/udes_stock_refactoring/models/stock_picking_batch.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @api.depends("picking_ids", "picking_ids.picking_type_id")
+    def _compute_picking_type(self):
+        """Compute the picking types from the linked pickings"""
+        for batch in self:
+            if batch.picking_ids:
+                batch.picking_type_ids = batch.picking_ids.picking_type_id
+            elif not isinstance(batch.id, models.NewId):
+                # If the picking ids are empty use the stored picking type ids
+                batch.picking_type_ids = batch.read(["picking_type_ids"])[0]["picking_type_ids"]
```
nocturnalastro, line 17: this last one is it not empty? I'm clearly missing something

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        Delete pickings in self that are empty, locked and marked=False.
+        If self is empty, find and delete any picking with the previous criterion.
+        This is to prevent us leaving junk data behind when refactoring
+        """
+        records = self.browse()
+        for to_unlink in self.find_empty_pickings_to_unlink():
```
nocturnalastro, line 164: Doesn't seem like that function is doing its job if we have to check if there empty. does it need a name change?

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,176 @@
...
+        domain = [
+            ("u_mark", "=", False),
+            ("is_locked", "=", True),
+        ]
+        if self:
+            domain.append(("id", "in", self.ids))
+        else:
+            pts = PickingType.search([("u_auto_unlink_empty", "=", False)])
+            if pts:
+                domain.append(("picking_type_id", "not in", pts.ids))
```
nocturnalastro, line 42: this logic seems backwards.

addons/udes_stock_refactoring/models/stock_picking_type.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    u_post_assign_action = fields.Selection(
+        selection=U_STOCK_REFACTORING_CORE_LIFECYCLE_ACTIONS,
+        string="Post Assign Action",
+        help="Choose the action to be taken after reserving a picking.",
+    )
+
+    u_post_validate_action = fields.Selection(
+        selection=U_STOCK_REFACTORING_CORE_LIFECYCLE_ACTIONS,
```
nocturnalastro, line 52: Do we need a U in front of the constant?

addons/udes_stock/tests/common.py
```diff
@@ -410,12 +410,24 @@ def create_move(cls, pickings, products_info, **kwargs):
     @classmethod
     def create_move_line(cls, moves, qty, **kwargs):
```
peterpratt13, line 411: Nit: Should probably be move not moves now

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: No because the original picking only has one move line, the other move line is in another picking that was created by refactoring.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Yep have added a check for this and also got rid of the for loop to check which move line and replaced it with a simpler check on the move lines recordset.

Note this test is now using apples instead of cherries as I created apple and banana pallets and quants in setUp.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Nope, removed. :)

Found a few instances of the checks being made on name/id/ids which weren't needed and have changed these.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now done this. Apple and Banana pallets and quants moved to setup. Have adjusted most tests to make use of these (so in some tests cherry or elderberry were replaced by apple or banana).

addons/udes_stock_refactoring/models/stock_picking_type.py
PeterC10: I only added it on the very unlikely chance that Odoo put in their own refactor functionality and also used the `STOCK_REFACTORING_CORE_LIFECYCLE_ACTIONS` constant variable. I'm probably being too paranoid so happy to remove the U if you think its a bit over the top. :)

addons/udes_stock/tests/common.py
PeterC10: Yep, good spot!

PeterC10: Have now renamed this. :)

addons/udes_stock_refactoring/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "UDES Stock Refactoring",
+    "summary": """UDES Stock Refactoring""",
+    "author": "Unipart Digital Team",
+    "website": "https://unipart.io",
+    "category": "Specific Industry Applications",
+    "license": "LGPL-3",
+    "application": True,
+    "version": "13.0.2.1.0",
```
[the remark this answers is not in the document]
PeterC10: I took this from the `udes_security` module but having checked some other modules there isn't really a pattern. Some others are 0.1 or even 0.01. `udes_stock` is set to 0.1 so maybe best to copy that?

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,385 @@
...
+    def _prepare_extra_info_for_new_picking_for_group(self, pickings, moves):
+        """
+        Given the group of moves to refactor and its related pickings,
+        prepare the extra info for the new pickings that might be created
+        for the group of moves.
+        Fields with more than one value are going to be ignored.
+        """
+        values = {}
+
+        origins = list(set(pickings.mapped("origin")))
```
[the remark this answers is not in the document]
PeterC10: Without `mapped` here an error is raised when origins is False:

`TypeError: 'bool' object is not iterable`

But agree we shouldn't really need it here, I think this just needs reworking a bit to get it work without `mapped` so will look into this.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,385 @@
...
+
+        However, in stock.move._split() in Odoo core, _action_confirm is
+        called on a single move and it expects a single move to be returned
+        (will error out if not). Therefore the refactor at confirm should avoid
+        splitting/creating moves. Luckily this is unlikely to be an issue as
+        it makes little sense to split moves at this stage, before they have
+        stock reserved against them.
+        """
+        res = super(StockMove, self)._action_confirm(*args, **kwargs)
+        post_refactor_moves = res._action_refactor(stage="confirm")
```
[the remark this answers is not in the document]
PeterC10: Have now added the exists check in action_confirm and action_done.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,385 @@
...
+           equivalent moves after they have been transformed.
+           The output moves may be identical to the input, may contain none
+           of the input moves, or anywhere in between.
+           The output should contain a functionally similar set of moves.
+        """
+        if stage is not None and stage not in U_STOCK_REFACTOR_STAGES.values():
+            raise UserError(_("Unknown stage for move refactor: %s") % stage)
+        moves = self
+
+        if self._context.get("disable_move_refactor"):
```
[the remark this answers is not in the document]
PeterC10: Don't think this was done intentionally in the UDES 11 stock module so have changed this.

addons/udes_stock_refactoring/models/stock_move.py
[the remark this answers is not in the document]
PeterC10: Have looked into this more and think `mapped` is still needed here. Can only access `origin` directly if the pickings recordset only has 1 picking record otherwise Odoo will complain that it is expecting a singleton.

Furthermore in `refactor_by_move_line_groups` we are potentially passing in multiple picking records to `_prepare_extra_info_for_new_picking_for_group`.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,647 @@
...
+        )
+        banana_pallet = self.create_package()
+        self.create_quant(
+            self.banana.id, self.test_stock_location_01.id, 10, package_id=banana_pallet.id
+        )
+
+        products_info = [{"product": self.apple, "qty": 10}, {"product": self.banana, "qty": 10}]
+        moves = self.create_move(self.picking, products_info)
+
+        apple_move = moves[0]
```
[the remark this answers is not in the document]
PeterC10: Had missed a few of these. I've now replaced these instances with filtered checks on the moves as we can't be 100% sure on the order the moves will be returned.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -21,22 +22,78 @@ def setUp(self):
...
     def test_reserve_one_pallet_per_product_split(self):
         """
         Reserve self.picking with one pallet of each product and check it
         splits correctly when reserved.
         """
-        apple_pallet = self.create_package()
-        self.create_quant(
-            self.apple.id, self.test_stock_location_01.id, 10, package_id=apple_pallet.id
-        )
+        apple = self.apple
```
peterpratt13, line 116: Is this necessary, why not just use self.apple. We don't seem to use it that much. (No strong preference though)

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -62,30 +56,62 @@ def _check_move_line(self, move_line, product, product_qty, package):
...
+    def _assert_picking_fields(
+        self,
+        picking,
+        state=None,
+        origin=None,
+        partner=None,
+        group_name=None,
+        location_id=None,
+        location_dest_id=None,
+        date_done=None,
```
peterpratt13, line 68: Could simplify this with just using kwargs
def _assert_picking_fields(self, picking, **kwargs):
    self.assertEqual(len(picking), 1)
    for key, val in kwargs.items():
        self.assertEqual(picking[key], val)

peterpratt13: Although this doesn't help with groups as names are being compared

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
nocturnalastro: We should probably agree on a scheme as I think that has to many digits for any of the ones I've come across.

addons/udes_stock_refactoring/models/stock_move.py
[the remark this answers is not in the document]
nocturnalastro: `Have looked into this more and think mapped is still needed here. Can only access origin directly if the pickings recordset only has 1 picking record otherwise Odoo will complain that it is expecting a singleton.`

Though Miquel had mentioned this had changed for 13. Maybe I'm miss remembering something NVM :)

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
PeterC10: Will go ask in the UDES Devs group. :)

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have added a `_get_default_new_picking_for_group_values` method for this. This method is now also used in conjunction with the `create_picking` helper method from `udes_stock`.

I've also split out the functionality in this method for removing misleading values from the picking into its own `_remove_misleading_values` method.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: David made the same comment the other day, so have removed these now. :)

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Yeah having to check groups by names made it a bit more awkward so felt it was better to use the named arguments for all fields being checked as there aren't that many.

PeterC10:
After speaking with Michele the refactor actions are now defined in their own classes, based off an abstract `refactor` class. This means we no longer need to specify various `refactor_action_xyz` functions and rely on string concatenation (which made following the code a bit of a chore) to find the right refactor action to use.

addons/udes_stock_refactoring/models/stock_picking_type.py
PeterC10: This constant has now been removed as refactor actions are now in their own classes.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Yep, can't think why it was done like this. Maybe the logic was a bit more complex before and got cut down over time.

Anyway have reversed this. :)

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: True, have renamed this to ```get_unmarked_pickings``` as that is all it does, it doesn't do any explicit checks to make sure the pickings it finds are empty.

addons/udes_stock_refactoring/models/stock_picking_batch.py
PeterC10: Really not sure why this was done, have not been able to trigger this either. Tempted to remove this elif.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Honestly I'm not sure. Have we run into this issue with UDES 11 in past?

addons/udes_stock_refactoring/models/refactor_batch_pickings_by_date.py
```diff
@@ -0,0 +1,32 @@
...
+        return "batch_pickings_by_date"
+
+    @classmethod
+    def description(cls):
+        """Set description of the refactor action."""
+        return "Batch Pickings by Date"
+
+    def do_refactor(self, moves):
+        """Batch pickings by date."""
+        moves._refactor_action_batch_pickings_by(
```
nocturnalastro, line 24: Does `_refactor_action_batch_pickings_by` return the moves?

PeterC10: No it doesn't. I think that is why there is a ```if new_moves is not None``` check in ```_action_refactor``` in Stock Move.

addons/udes_stock_refactoring/models/refactor_group_by_move_line_key.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+from itertools import groupby
+
+from odoo import models, fields, _
+
+from .refactor import Refactor, get_selection
+
+
+class ByMoveLineKey(Refactor):
```
nocturnalastro, line 9: You have `GroupByMoveKey` make these consistent :)

addons/udes_stock_refactoring/models/registry.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from abc import ABCMeta
+from addons.udes_common.models.models import groupby
+
+
+class RegistryMeta(ABCMeta):
```
nocturnalastro, line 6: As suggested locations and this is using this meta do we want to move that into the base module with the groupby and the other tools

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,355 @@
...
+                refactor_action = picking_type.u_post_assign_action
+            elif stage == "validate":
+                refactor_action = picking_type.u_post_validate_action
+
+            if refactor_action:
+                refactor_class = REFACTOR_REGISTRY[refactor_action]
+                return refactor_class(self.env)
+
+            return refactor_action
+        except KeyError:
```
nocturnalastro, line 46: Can we make the try except as small as possible :) Or use a get and if :)

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,355 @@
...
+        # this with the equivalent for move_line_key
+        # we need to think of how we want to do it
+
+        # The environment must include {'compute_key': True}
+        # to allow the keys to be computed.
+        if not self.env.context.get("compute_key", False):
+            return
+        for move in self:
+            move_vals = {
+                field_name: move[field_name]
```
nocturnalastro, line 61: Feel's like there should be a way of figuring out which fields we need for for the key then storing them 

then using them to power a lambda or only recalculating on changes to them, haven't thought about this enough though perhaps we should take a note to do an investigation to see if improves performance.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,355 @@
...
+           The output should contain a functionally similar set of moves.
+        """
+        if stage is not None and stage not in U_STOCK_REFACTOR_STAGES.values():
+            raise UserError(_("Unknown stage for move refactor: %s") % stage)
+        moves = self
+
+        if self.env.context.get("disable_move_refactor"):
+            return moves
+
+        refactor_moves = moves.filtered(
```
nocturnalastro, line 106: we could do something like 

```
refactor_lam = lambda m: m.picking_type_id and m.state not in ["draft", "cancel"]
if stage is not None:
    refactor_lam =  lambda m, lam=refactor_lam: U_STOCK_REFACTOR_STAGES[m.state] == stage and lam(m)

refactor_moves = moves.filtered(refactor_lam)
```

saves a loop over moves not sure if its worth it though probably be premature optimization.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,355 @@
...
+        called on a single move and it expects a single move to be returned
+        (will error out if not). Therefore the refactor at confirm should avoid
+        splitting/creating moves. Luckily this is unlikely to be an issue as
+        it makes little sense to split moves at this stage, before they have
+        stock reserved against them.
+        """
+        res = super(StockMove, self)._action_confirm(*args, **kwargs)
+        post_refactor_moves = res.exists()._action_refactor(stage="confirm")
+
+        if post_refactor_moves != res:
```
nocturnalastro, line 155: Could we check if MRP is installed?

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,355 @@
...
+            for picking_type in self.picking_id.picking_type_id
+        ):
+            raise UserError(
+                _("Cannot group moves when their picking type has no grouping key set.")
+            )
+
+        # force recompute on u_grouping_key so we have an up-to-date key:
+        return self.with_context(compute_key=True).groupby(lambda ml: ml.u_grouping_key)
+
+    def _prepare_extra_info_for_new_picking_for_group(self, pickings, moves):
```
nocturnalastro, line 208: (Feel free to resolve): Not sure this function is in the write place (on moves instead of picking)

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Good idea, modules are records just like everything else in Odoo so should be possible. Will go look into this.

addons/udes_stock_refactoring/models/refactor_group_by_move_line_key.py
PeterC10: Whoops, will amend. :)

PeterC10: This has now been renamed. :)

addons/udes_stock_refactoring/models/registry.py
PeterC10: Have now moved this into `udes_common` and updated `udes_stock_refactoring` and `suggested_locations` to take the RegistryMeta from there.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Yeah good point. Have now moved the try/except to only cover when it tries to actually find the class. Have also added a `stage_post_action_dict` in `_get_refactor_class` to replace the if statements.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have added this in, every little helps plus it cuts down the code a bit. :)

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have added a `get_module_installed` method in `udes_common` and now make use of it here so that the error message is only raised if the `mrp` is installed.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Yeah agreed. Have moved this over to Picking, and this also removed the need for the `pickings` argument, as they can now just be taken from `self`.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have raised issue #10656 to track this so it can be discussed more and an investigation sorted.

Will also add a `TODO` in this function.

PeterC10: Have added `TODO` for `compute_grouping_key` on Stock Move and Stock Move Line.

addons/udes_common/models/ir_module.py
```diff
@@ -0,0 +1,16 @@
+from odoo import models, api
+
+
+class Module(models.Model):
+    _inherit = "ir.module.module"
+
+    @api.model
+    def get_module_installed(self, module_name):
```
peterpratt13, line 8: Nit: would is_module_installed be better here a ti just returns boolean, rather than the module

addons/udes_common/models/ir_module.py
```diff
@@ -0,0 +1,16 @@
...
+        """
+        Returns true if the supplied module exists 
+        and is installed otherwise false.
```
peterpratt13, line 11: Super nit, feel free to ignore! Can you add a comma after is installed, I read it as is installed otherwise the first 2 times and just got confused about the false.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -28,23 +28,22 @@ def _get_refactor_class(self, picking_type, stage):
...
+            "validate": picking_type.u_post_validate_action,
+        }
 
-            if stage == "confirm":
-                refactor_action = picking_type.u_post_confirm_action
-            elif stage == "assign":
-                refactor_action = picking_type.u_post_assign_action
-            elif stage == "validate":
-                refactor_action = picking_type.u_post_validate_action
+        refactor_action = stage_post_action_dict.get(stage, None)
```
peterpratt13, line 37: Nit: None is default so not necessary - but feel free to leave in if you like the explicitness :)

addons/udes_common/models/ir_module.py
PeterC10: Yeah does sound more suitable, will rename. :)

addons/udes_common/models/ir_module.py
PeterC10: Always happy to make one-character changes! :smile: Will amend. :)

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Was more me being too paranoid, will remove the `None`. :)

peterpratt13: where is stage? or am i being daft?

peterpratt13: ignore being daft

PeterC10: Have now removed the unnecessary None default.

addons/udes_common/models/ir_module.py
PeterC10: Comma added. :)

addons/udes_common/models/ir_module.py
PeterC10: Have now renamed this.

addons/udes_stock/models/stock_move.py
williamsajr: Personally I'd prefer a single function that can accept a dictionary or a list of dictionaries similar to how odoo13 does .create()

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -1,4 +1,4 @@
-from odoo import models
+from odoo import models, fields
```
ws-mps, line 1: Is this needed?

addons/udes_stock/tests/test_stock_move.py
```diff
@@ -107,3 +124,65 @@ def test04_unreserve_initial_demand(self):
         new_move_lines = MoveLine.search([("product_id", "=", self.fig.id)])
         self.assertEqual(move_lines, new_move_lines)
         self.assertIn(pack2_ml, new_move_lines)
+
+    def test05_pepare_and_create_single_move_line(self):
+        """ Prepare and create a single move line and check values are correct """
```
ws-mps, line 129: Usually, move lines are created when stock is reserved. After setup there are only 2 units left for bananas, so probably change product_uom_qty to 2 instead of 5 to make it more realistic.

Also the state of the current pick is "assigned" and adding new moves in state draft without confirming/assigning them does not make much sense.

addons/udes_stock/tests/test_stock_move.py
```diff
@@ -107,3 +124,65 @@ def test04_unreserve_initial_demand(self):
...
+        move_line = self.Move._create_move_line(move_line_values)
+
+        # Check that one move line has been created and that the move now contains
+        # the created move line
+        self.assertEqual(len(move_line), 1)
+        self.assertEqual(move.move_line_ids, move_line)
+
+    def test06_pepare_and_create_multiple_move_lines(self):
+        """ Prepare and create a multiple move lines and check values are correct """
+        apple_uom_qty = 5
```
ws-mps, line 151: Same as test05.

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
ws-mps: What was the outcome of this ?

addons/udes_stock/models/stock_picking_type.py
PeterC10: Nope, was left in as some fields were added here in an earlier commit in this branch but no longer needed. Have removed. :)

addons/udes_common/models/registry.py
```diff
@@ -9,15 +9,18 @@ class RegistryMeta(ABCMeta):
 
     def __init__(cls, name, bases, dict):
         super(RegistryMeta, cls).__init__(name, bases, dict)
-        policy_name = cls.name()
+        refactor_action_name = cls.name()
```
ws-mps, line 12: Since this is generic now, can we change `refactor_action_name` to a neutral name? it is used in refactor but also in suggested locations.

PeterC10: Good spot, missed that. Will amend. Any preference on the name? I suppose it could just be `name`?

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,334 @@
...
+            for picking_type in self.picking_id.picking_type_id
+        ):
+            raise UserError(
+                _("Cannot group moves when their picking type has no grouping key set.")
+            )
+
+        # force recompute on u_grouping_key so we have an up-to-date key:
+        return self.with_context(compute_key=True).groupby(lambda ml: ml.u_grouping_key)
+
+    def refactor_by_move_groups(self, groups):
```
ws-mps, line 212: Is this only meant to be used inside the new group by move key class? If so, shouldn't this function be in the class?

ws-mps: Same for move_line.

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
PeterC10: To follow the OCA system:

Odoo major version (in this case `13.0`) followed by module `x.y.z` version numbers.

So 13.0 can stay, but it still needs to be decided what the module version is, 2.1.0, 0.1.0 or something else entirely.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,334 @@
...
+        called on a single move and it expects a single move to be returned
+        (will error out if not). Therefore the refactor at confirm should avoid
+        splitting/creating moves. Luckily this is unlikely to be an issue as
+        it makes little sense to split moves at this stage, before they have
+        stock reserved against them.
+        """
+        Module = self.env["ir.module.module"]
+
+        res = super(StockMove, self)._action_confirm(*args, **kwargs)
+        post_refactor_moves = res.exists()._action_refactor(stage="confirm")
```
ws-mps, line 157: We are doing `.exists()` always before calling `_action_refactor`, do we want to do it inside instead? In such case we should update `_action_refactor` docstring.,

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,218 @@
...
+        if len(partners) == 1:
+            values["partner_id"] = partners.id
+
+        dates_done = list(set(moves.mapped("date")))
+        if len(dates_done) == 1:
+            values["date_done"] = dates_done[0]
+
+        return values
+
+    def _get_package_search_domain(self, package):
```
ws-mps, line 170: Is this still needed?

addons/udes_stock_refactoring/models/stock_picking_batch.py
ws-mps: It was needed in O11 for other reasons, and probably not needed in O13. You can remove it. But it is mandatory to set a value on O13, so you will sill need an else.

addons/udes_stock_refactoring/models/stock_picking.py
ws-mps: Have we checked if we are leaving empties still ?

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
+# -*- coding: utf-8 -*-
+
+from addons.udes_stock.tests import common
+
+
+class TestSplittingBase(common.BaseUDES):
```
ws-mps, line 6: We could probably rename this to Refactoring instead of splitting. Same for the file name.

addons/udes_stock_refactoring/views/stock_picking_type_views.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_picking_type_form_udes_stock_refactoring" model="ir.ui.view">
+        <field name="name">stock.picking.type.form.udes_stock_refactoring</field>
+        <field name="model">stock.picking.type</field>
+        <field name="inherit_id" ref="stock.view_picking_type_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//form/sheet/group[last()]" position="after">
+                <group string="UDES: Pick Refactoring" groups="base.group_no_one">
```
ws-mps, line 10: nit: Picking Refactoring

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
...
+
+        products_info = [{"product": self.apple, "qty": 10}, {"product": self.banana, "qty": 10}]
+        moves = self.create_move(self.picking, products_info)
+
+        apple_move = moves.filtered(lambda m: m.product_id == self.apple)
+        banana_move = moves - apple_move
+
+        self.picking.action_assign()
+
+        apple_pick = apple_move.picking_id
```
ws-mps, line 127: We should check that they are different. Otherwise all the following assert could still pass without moving one of the moves into a new picking.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
...
+        self._assert_move_line_fields(apple_move_line, self.apple, 10, apple_pallet)
+
+        banana_move = banana_pick.move_lines
+        self._assert_move_fields(banana_move, self.banana, 10)
+
+        banana_move_line = banana_pick.move_line_ids
+        self._assert_move_line_fields(banana_move_line, self.banana, 10, banana_pallet)
+
+        # Check that the original pick has been reused
+        self.assertTrue(self.picking.exists())
```
ws-mps, line 147: You could also check that the original picking is either apple_pick or banana_pick.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
...
+
+        self.picking.action_assign()
+
+        apple_move_lines = MoveLine.search([("product_id.name", "=", "Test product Apple")])
+        self.assertEqual(len(apple_move_lines), 2)
+
+        apple_pallet_move_line = apple_move_lines.filtered(lambda ml: ml.package_id == apple_pallet)
+        other_apple_pallet_move_line = apple_move_lines - apple_pallet_move_line
+
+        self.assertTrue(apple_pallet_move_line.exists())
```
ws-mps, line 176: We have just search for those move lines, so .exists() is not needed.

addons/udes_stock_refactoring/tests/test_splitting.py
ws-mps: I cannot find this check :)

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
...
+    def test_reserve_simultaneously_2_picks_1_pallet_same_prods_splits_1_pick(self):
+        """
+        With 2 picks for 2 of the same products on one pallet.
+        Reserve both picks (action_assign) simultaneously.
+        Original pick should be marked as empty and deleted.
+        A new pick should be created with both the moves.
+        """
+        apple_pallet = self.apple_pallet
+
+        self.create_quant(self.apple.id, self.test_stock_location_01.id, 15, package_id=apple_pallet.id)
```
ws-mps, line 260: No need to create a new quant, just increase the quantity to 25.

addons/udes_stock_refactoring/models/stock_picking.py
ws-mps: Let's create an issue to revise this, because we could get rid of this function if we change stuff in Odoo core.

addons/udes_stock_refactoring/tests/test_splitting.py
```diff
@@ -0,0 +1,649 @@
...
+        fig_moveline = move_lines.filtered(lambda ml: ml.product_id == fig)
+        grape_moveline = move_lines - fig_moveline
+
+        self._assert_move_line_fields(fig_moveline, fig, 5, mixed_pallet)
+        self._assert_move_line_fields(grape_moveline, grape, 10, mixed_pallet)
+
+        # Check that the original pick has been reused
+        self.assertTrue(self.picking.exists())
+
+    def test_reserve_simultaneously_2_picks_1_pallet_same_prods_splits_1_pick(self):
```
ws-mps, line 251: nit: merges

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have raised issue #10698 to track this.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: This has now been changed.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now done this.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have now added a check to make sure that `self.picking` is either `apple_pick` or `banana_pick`.

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Have added a comment explaining how `_assert_picking_fields` will cover this as both `apple_pick` and `banana_pick` are checked with different `group_name` values (as a picking can only have one group).

addons/udes_stock_refactoring/views/stock_picking_type_views.xml
PeterC10: Have changed this. :)

addons/udes_stock_refactoring/tests/test_splitting.py
PeterC10: Will leave this till the very end incase further changes are needed to the tests.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Nope, removed. :) Was previously called by `get_pickings` which was removed in an earlier commit.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: As discussed I have moved the exists check to `_action_refactor` and updated the doc string, so it has now been removed from `_action_confirm`, `_action_assign` and `_action_done`.

addons/udes_common/models/registry.py
PeterC10: Have now changed this to `name`. :)

addons/udes_stock/tests/test_stock_move.py
PeterC10: Have changed `banana_uom_qty` to 2. Apple didn't actually have any quants, so have created a quant set to `apple_uom_qty` to make the test a bit more realistic.

PeterC10: Also confirmed the picking and assigned the stock at the end of the test.

addons/udes_stock/tests/test_stock_move.py
PeterC10: Have updated `product_uom_qty` to 2. Picking is now also confirmed with the stock assigned at the end of the test.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,334 @@
...
+    def _action_refactor(self, stage=None):
+        """
+        Refactor moves in self if any exist.
+        :param stage: One of confirm|assign|done, if set, filters the moves
+            which will be refactored to only the state(s) that match:
+                - 'confirm': confirmed, waiting
+                - 'assign': assigned, partially_available
+                - 'done': done
+
+           Methods doing a refactor are expected to take a single recordset of
```
williamsajr, line 95: Nitpic: indentation of this bit of the docstring

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+
+    @api.model
+    def _new_picking_for_group(self, group_key, moves, **kwargs):
+        """
+        Find existing picking for the supplied group, if none found create a new one.
+        Assign the moves to the picking and return it.
+        """
+        ProcurementGroup = self.env["procurement.group"]
+
+        picking_type = moves.mapped("picking_type_id")
```
williamsajr, line 90: We don't need the ```mapped()``` for relationships in odoo13

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+        picking_type.ensure_one()
+        src_loc = moves.mapped("location_id")
```
williamsajr, line 92: here too

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+        dest_loc = moves.mapped("location_dest_id")
```
williamsajr, line 93: and here

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+                ("state", "in", ["assigned", "confirmed", "waiting"]),
+            ]
+        )
+
+        if not picking:
+            # Otherwise reuse the existing picking if all the moves
+            # already belong to it and it contains no other moves
+            # The picking_type_id, location_id and location_dest_id
+            # will match already
+            current_picking = moves.mapped("picking_id")
```
williamsajr, line 114: ```mapped```

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+            if len(current_picking) == 1 and current_picking.mapped("move_lines") == moves:
```
williamsajr, line 115: ```mapped```

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -0,0 +1,210 @@
...
+            picking_values.update(kwargs)
+
+            picking = self.create_picking(picking_type, **picking_values)
+
+        else:
+            picking._remove_misleading_values(**kwargs)
+
+        moves.write({"group_id": group.id, "picking_id": picking.id})
+
+        move_lines = moves.mapped("move_line_ids")
```
williamsajr, line 137: ```mapped```

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -0,0 +1,334 @@
...
+        """
+        Move = self.env["stock.move"]
+        Picking = self.env["stock.picking"]
+
+        pickings = self.picking_id
+
+        result_moves = Move.browse()
+
+        for key, ml_group in groups:
+            touched_moves = ml_group.mapped("move_id")
```
williamsajr, line 258: ```mapped```

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have now fixed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Have now removed this.

addons/udes_stock_refactoring/models/stock_move.py
PeterC10: Note: ```get_module_installed``` was renamed ```is_module_installed``` after this suggestion: https://github.com/unipartdigital/udes-open/pull/307#pullrequestreview-446292141

addons/udes_stock_refactoring/models/stock_move.py
williamsajr: You missed this one

PeterC10: Definitely gone now!

PR #309, opened 2020-07-03
 User-story: 10561 - Do not bypass_reservation_update at _unreserve_inital_demand

ws-mps:
By default it is now bypassed, but in some cases still it shouldn't.

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -10,24 +10,26 @@ _logger = logging.getLogger(__name__)
 
 # Map move state to the refactor stage.
 STAGES = {
-    'confirmed': 'confirm',
-    'waiting': 'confirm',
-    'assigned': 'assign',
-    'partially_available': 'assign',
-    'done': 'validate'
+    "confirmed": "confirm",
+    "waiting": "confirm",
+    "assigned": "assign",
+    "partially_available": "assign",
+    "done": "validate",
 }
+
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
-    u_grouping_key = fields.Char('Key', compute='compute_grouping_key')
+    u_grouping_key = fields.Char("Key", compute="compute_grouping_key")
 
     def _unreserve_initial_demand(self, new_move):
         """ Override stock default function to keep the old move lines,
             so there is no need to create them again.
         """
         self.mapped("move_line_ids").filtered(lambda x: x.qty_done == 0.0).write(
-            {'move_id': new_move, 'product_uom_qty': 0})
+            {"move_id": new_move, "product_uom_qty": 0}
+        )
 
     def _prepare_info(self):
         """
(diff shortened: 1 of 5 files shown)


addons/udes_stock/tests/test_move.py
```diff
@@ -100,3 +100,31 @@ def test03_cancel_backorder_cancels_remaining(self):
...
+
+        self.assertEqual(picking.state, 'assigned')
+        self.assertEqual(apple_quant.reserved_quantity, 4)
+
+        partial_pick_apples = [{'barcode': self.apple.barcode, 'qty': 2}]
+
+        picking.update_picking(product_ids=partial_pick_apples, validate=True, create_backorder=True)
+        self.assertEqual(picking.state, 'done')
+        backorder = Picking.get_pickings(backorder_id=picking.id)
+        self.assertEqual(backorder.state, 'assigned')
```
apatrick88: Is it worth asserting the remaining reserved quantity on the backorder picking too?

addons/udes_stock/models/stock_move.py
```diff
@@ -10,25 +10,26 @@
...
 
     def _unreserve_initial_demand(self, new_move):
         """ Override stock default function to keep the old move lines,
             so there is no need to create them again
         """
-        self.mapped('move_line_ids')\
-            .filtered(lambda x: x.qty_done == 0.0)\
-            .write({'move_id': new_move, 'product_uom_qty': 0})
+        self.with_context(bypass_reservation_update=False).mapped("move_line_ids").filtered(
+            lambda x: x.qty_done == 0.0
```
williamsajr, line 31: I don't think this change will work as qty_done is not zero, so this change will not have any effect. We probably want to create a new move line for the remainder on the new move instead anyway

addons/udes_stock/tests/test_move.py
```diff
@@ -100,3 +100,31 @@ def test03_cancel_backorder_cancels_remaining(self):
...
+        Picking = self.env['stock.picking']
+        apple_quant = self.create_quant(self.apple.id, self.test_location_01.id, 4)
+        picking = self.create_picking(self.picking_type_pick, products_info=self.pack_4apples_info, confirm=True, assign=True)
+
+        self.assertEqual(picking.state, 'assigned')
+        self.assertEqual(apple_quant.reserved_quantity, 4)
+
+        partial_pick_apples = [{'barcode': self.apple.barcode, 'qty': 2}]
+
+        picking.update_picking(product_ids=partial_pick_apples, validate=True, create_backorder=True)
```
williamsajr: I think we should be updating the move lines (qty_done)

addons/udes_stock/tests/test_move.py
ws-mps: The problem was it wasn't getting assigned at all so if the state is assigned means we have found stock, so the quantity is not really needed.

PR #311, opened 2020-07-08
[ADD] udes_stock: Restrict inventory adjustment on reserved stock

williamsajr:
A new config option may be specified on the warehouse to disallow
inventory adjustments on reserved quants. A warning will appear
if attempted and the user will be blocked from validating with
this setting or if not performed by a debug user.

### DIFF
--- a/addons/udes_stock/models/stock_inventory.py
+++ b/addons/udes_stock/models/stock_inventory.py
@@ -2,6 +2,7 @@
 
 from odoo import api, models, fields, _
 from odoo.exceptions import ValidationError, UserError
+from odoo.addons import decimal_precision as dp
 
 
 class StockInventory(models.Model):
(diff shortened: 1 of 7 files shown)


addons/udes_stock/models/stock_inventory.py
```diff
@@ -24,19 +25,94 @@ def action_done(self):
...
+        'company_id',
+        'prod_lot_id',
+        'partner_id'
+    )
+    def _compute_reserved_qty(self):
+        """Compute the reserved quantity for the line."""
+        if not self.product_id:
+            self.reserved_qty = 0
+            return
+        reserved_qty = sum([x.reserved_quantity for x in self._get_quants()])
```
davidmcgaw, line 112: maybe call the variable q or quants instead of x?

addons/udes_stock/tests/test_stock_inventory.py
```diff
@@ -0,0 +1,87 @@
...
+from . import common
+
+class TestStockInventory(common.BaseUDES):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestStockInventory, cls).setUpClass()
+        cls.apple_quant = cls.create_quant(cls.apple.id, cls.test_location_01.id, 5)
+
+    def setUp(self):
```
davidmcgaw, line 11: please could you add doc strings to all methods

addons/udes_stock/models/stock_inventory.py
williamsajr: Changed

addons/udes_stock/tests/test_stock_inventory.py
williamsajr: Added

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -95,6 +95,12 @@ def _domain_u_damaged_location_id(self):
         help="Maximum depth for package hierarchy. I.e. a value of 2 would limit the number of levels in hierarchy to 2, one level of packages(with no subpackages) inside an outer package."
     )
 
+    u_inventory_adjust_reserved = fields.Boolean(
+        string="Invetory Adjust Reserved Quants",
```
apatrick88, line 99: Typo 'Inventory'

williamsajr: Good spot

addons/udes_stock/views/stock_inventory.xml
```diff
@@ -16,5 +16,33 @@
...
+                </xpath>
+            </field>
+        </record>
+
+        <record id="view_adjust_reserved" model="ir.ui.view">
+            <field name="name">stock.inventory.adjust.reserved.view.form</field>
+            <field name="model">stock.inventory</field>
+            <field name="priority">20</field>
+            <field name="arch" type="xml">
+                <form string="Adjust Reserved Quants?">
```
apatrick88, line 36: Do we want to call them reserved quants or reserved stock? Might be more user friendly.

addons/udes_stock/tests/test_stock_inventory.py
```diff
@@ -0,0 +1,89 @@
+from odoo.exceptions import ValidationError
+from . import common
+
+class TestStockInventory(common.BaseUDES):
+
+    @classmethod
+    def setUpClass(cls):
+        """Add a quant in a location to the setUpClass method"""
```
apatrick88, line 8: Ok, I'm being super nit-picky and I don't like saying it, but some of your docstrings end with a full stop and some don't, I'm not going to hold anything up over that though!

addons/udes_stock/views/stock_inventory.xml
williamsajr: Good idea, I'll change this

addons/udes_stock/tests/test_stock_inventory.py
williamsajr: Changed to be consistent

addons/udes_stock/models/stock_inventory.py
```diff
@@ -24,19 +25,94 @@ def action_done(self):
...
+        'package_id',
+        'product_uom_id',
+        'company_id',
+        'prod_lot_id',
+        'partner_id'
+    )
+    def _compute_reserved_qty(self):
+        """Compute the reserved quantity for the line."""
+        if not self.product_id:
+            self.reserved_qty = 0
```
davidmcgaw, line 110: is there a unit test that covers this path?

williamsajr: I don't think we will hit this bit of logic here as product_id is required so I'm going to remove it. I'll add tests for this function anyway though

PR #312, opened 2020-07-08
Story/10465

peterpratt13:


addons/udes_stock/controllers/stock_picking.py
```diff
@@ -76,6 +76,16 @@ def is_compatible_package(self, ident, package_name):
...
         return picking.is_compatible_package(package_name)
 
+
+    @http.route('/api/stock-picking/<ident>/can_handle_partials',
+                type='json', methods=['GET'], auth='user')
+    def can_handle_partials(self, ident):
+        """ Check if the picking can handle partials """
+        Picking = request.env['stock.picking']
+        picking = Picking.browse(int(ident))
+        return picking.can_handle_partials()
```
squah-u, line 86: Could you update the udes_stock readme to include documentation of the new route?

addons/udes_stock/models/stock_picking.py
```diff
@@ -237,9 +239,11 @@ def _compute_state(self):
         """ Prevent pickings to be in state assigned when not able to handle
             partials, so they should remain in state waiting or confirmed until
             they are fully assigned.
+
+            Add a context variable to the can_handle_partials call
```
ws-mps, line 243: Would be good to improve this explanation, isn't this so that in can_handle_partials() we can distinguish if we are in the compute state function?

Have we considered using a parameter instead of context variable?

Do we want to call the field `computing_state` instead?

PR #319, opened 2020-07-16
Story/10340

kdwyer:


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -520,6 +527,13 @@ def add_extra_pickings(self, picking_type_id):
             raise ValidationError(_("Can only add work to ephemeral batches"))
 
         pickings.write({'batch_id': self.id})
+        PickingType = self.env['stock.picking.type']
```
ws-mps, line 530: We usually have this at the start of the function.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1021,17 +1035,32 @@ def reserve_pallet(self, pallet_name):
...
+            domain.extend([('id', '!=', self.id), ('u_last_reserved_pallet_name', '=', pallet_name)])
+        conflicting_batch = PickingBatch.search(domain)
         if conflicting_batch:
             raise ValidationError(
                 _("This pallet is already being used for batch %s.") %
                 conflicting_batch[0].name)
 
-        self.write({ 'u_last_reserved_pallet_name': pallet_name })
+        write_data = {'u_last_reserved_pallet_name': pallet_name}
+        if self.picking_ids.picking_type_id.u_reserve_pallet_per_picking:
```
ws-mps, line 1052: This will fail if picking_ids contains more than one picking.

addons/udes_stock/models/stock_picking_batch.py
kdwyer: Removed.

addons/udes_stock/models/stock_picking_batch.py
kdwyer: These have all been updated.

addons/udes_stock/controllers/stock_picking_batch.py
```diff
@@ -201,7 +203,20 @@ def reserve_pallet(self, ident, pallet_name):
             raise ValidationError(
                 _("The specified batch is not assigned to you."))
 
-        batch.reserve_pallet(pallet_name)
+        # TODO this is messy
```
ws-mps, line 206: Everywhere we assume the parameter is the type we expect, so you can remove the try except and go for: 
` pick = StockPicking.browse(int(picking_id))`

addons/udes_stock/controllers/stock_picking_batch.py
```diff
@@ -201,7 +203,20 @@ def reserve_pallet(self, ident, pallet_name):
...
+        if picking_id is not None:
+            picking = None
```
ws-mps, line 208: You can just get this out of the if and inside the if: `picking = StockPicking.browse(int(picking_id))`

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1017,43 @@ def reserve_pallet(self, pallet_name):
...
-            ('u_last_reserved_pallet_name', '=', pallet_name),
-        ])
+        reserve_pallet_per_picking = self.picking_ids.filtered('picking_type_id.u_reserve_pallet_per_picking')
+
+        if reserve_pallet_per_picking:
+            conflicting_picking = Picking.search([
+                ('u_reserved_pallet', '=', pallet_name),
+                ('state', 'not in', ['draft', 'cancelled', 'done'])
+            ])
+            conflicting_batch = conflicting_picking.batch_id
```
ws-mps, line 1032: We might want to raise the error with a conflicting picking instead of the batch. So probably two different error messages.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1017,43 @@ def reserve_pallet(self, pallet_name):
...
+                ('u_last_reserved_pallet_name', '=', pallet_name),
+            ])
         if conflicting_batch:
             raise ValidationError(
                 _("This pallet is already being used for batch %s.") %
                 conflicting_batch[0].name)
 
-        self.write({ 'u_last_reserved_pallet_name': pallet_name })
+        if reserve_pallet_per_picking:
+            # We expect at least one picking has already been assigned to the
```
ws-mps, line 1045: Reserve pallet should either reserve a pallet for the batch (stored at `u_last_reserved_pallet_name`) or for the picking (stored at `u_reserved_pallet`). So at this point we should write to picking straightaway.
We can add a check at the start of the function to ensure we have a picking and it is in the batch in self.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1017,43 @@ def reserve_pallet(self, pallet_name):
...
         PickingBatch = self.env['stock.picking.batch']
 
         self.ensure_one()
 
-        conflicting_batch = PickingBatch.search([
-            ('id', '!=', self.id),
-            ('state', '=', 'in_progress'),
-            ('u_last_reserved_pallet_name', '=', pallet_name),
-        ])
+        reserve_pallet_per_picking = self.picking_ids.filtered('picking_type_id.u_reserve_pallet_per_picking')
```
ws-mps, line 1025: If this is true we should check we have `picking != None`.

Batch has a field u_picking_type_ids or similar from which you can get `u_reserve_pallet_per_picking` with a shorter code.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -334,6 +334,14 @@ class StockPickingType(models.Model):
         " for any empty picking in the system.",
     )
 
+    # Maybe `reserve_pallet_per_picking` as we don;t know about sale orders in
```
ws-mps, line 337: Agree about the name.
Do we have another setting to enable matrix picking? or this is the setting that will be used to know it?

kdwyer: This would be the setting.

addons/udes_stock/controllers/stock_picking_batch.py
ws-mps: nit: `picking = XXX`

addons/udes_stock/models/stock_picking_batch.py
kdwyer: > We can add a check at the start of the function to ensure we have a picking

Are we assuming that the front end will always specify a picking if matrix picking is enabled? (@squah-u ?)

addons/udes_stock/models/stock_picking_type.py
ws-mps, line 337: typo: don;t

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,165 @@ def test04_rereservation_for_same_batch(self):
...
+        # enable unpickable items by default
+        cls.picking_type_pick.u_enable_unpickable_items = True
+
+        cls.picking_type_pick.u_reserve_pallet_per_picking = True
+
+    def setUp(self):
+        super().setUp()
+        Package = self.env['stock.quant.package']
+
+        self.package_one = Package.get_package("test_package_one", create=True)
```
ws-mps, line 1728: Do we need these packages?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,165 @@ def test04_rereservation_for_same_batch(self):
...
+        self.package_two = Package.get_package("test_package_two", create=True)
+        self.package_three = Package.get_package("test_package_three", create=True)
+        self.package_four = Package.get_package("test_package_four", create=True)
+
+        self.create_quant(self.apple.id,
+                          self.test_location_01.id,
+                          300,
+                          package_id=self.create_package().id)
```
ws-mps, line 1736: Do we want to use the packages created above?

addons/udes_stock/tests/test_picking_batch.py
kdwyer: No, they're copy-pasta from the original reservation tests - the test code needs some clean up.

addons/udes_stock/models/stock_picking_batch.py
squah-u: The matrix pick flow will always specify a picking while the regular pick flow will never specify a picking. It would be technically possible for `u_reserve_pallet_per_picking` and the normal pick flow to be both enabled at the same time.

addons/udes_stock/models/stock_picking_batch.py
kdwyer: New exception message added

addons/udes_stock/models/stock_picking_type.py
kdwyer: I've removed the comment, as it was redundant anyway

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -510,14 +518,22 @@ def add_extra_pickings(self, picking_type_id):
...
+
         picking_priorities = self.get_batch_priority_group()
         pickings = Picking.search_for_pickings(picking_type_id, picking_priorities)
 
         if not pickings:
             raise ValidationError(_("No more work to do."))
 
-        if not self.u_ephemeral:
-            raise ValidationError(_("Can only add work to ephemeral batches"))
+        reserve_pallet_per_picking = self.picking_type_ids.filtered('u_reserve_pallet_per_picking')
```
ws-mps, line 530: Does this work without lambda ?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -474,6 +474,14 @@ def _create_batch(self, user_id, picking_type_id, picking_priorities=None, picki
         if not picking:
             return None
 
+        picking_type = picking.mapped('picking_type_id')
+        reserve_pallet_per_picking = picking_type.filtered('u_reserve_pallet_per_picking')
```
ws-mps, line 478: Does this work without lambda?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -474,6 +474,14 @@ def _create_batch(self, user_id, picking_type_id, picking_priorities=None, picki
...
+        if reserve_pallet_per_picking:
+            max_reservable_pallets = picking_type.mapped('u_max_reservable_pallets')[0]
```
ws-mps, line 480: We should only have one picking type, so no need for mapped or [0]. We could add ensure_one() after line 477 to enforce this.

addons/udes_stock/models/stock_picking_batch.py
ws-mps, line 530: Does this work without lambda?

Won't be better to get the picking type from `pickings` like we have done at `_create_batch`?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -510,14 +518,22 @@ def add_extra_pickings(self, picking_type_id):
...
         pickings = Picking.search_for_pickings(picking_type_id, picking_priorities)
 
         if not pickings:
             raise ValidationError(_("No more work to do."))
 
-        if not self.u_ephemeral:
-            raise ValidationError(_("Can only add work to ephemeral batches"))
+        reserve_pallet_per_picking = self.picking_type_ids.filtered('u_reserve_pallet_per_picking')
+        if reserve_pallet_per_picking:
+            max_reservable_pallets = self.picking_type_ids.mapped('u_max_reservable_pallets')[0]
```
ws-mps, line 532: If we get the picking type from pickings, then this line does not need `.mapped()` and neither need `[0]`.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1033,45 @@ def reserve_pallet(self, pallet_name):
...
-        if conflicting_batch:
-            raise ValidationError(
-                _("This pallet is already being used for batch %s.") %
-                conflicting_batch[0].name)
+        reserve_pallet_per_picking = self.picking_type_ids.filtered('u_reserve_pallet_per_picking')
+        if reserve_pallet_per_picking:
+            if not picking:
+                raise ValidationError('A picking must be specified if pallets are reserved per picking.')
+        if reserve_pallet_per_picking and picking and not self.picking_ids & picking:
+            return
```
ws-mps, line 1046: Do we want to raise an error in this case?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1033,45 @@ def reserve_pallet(self, pallet_name):
...
 
-        self.write({ 'u_last_reserved_pallet_name': pallet_name })
+        if reserve_pallet_per_picking:
+            conflicting_picking = Picking.search([
+                ('u_reserved_pallet', '=', pallet_name),
```
ws-mps, line 1050: Do we want a compound index for this search?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1033,45 @@ def reserve_pallet(self, pallet_name):
...
+                ('state', 'not in', ['draft', 'cancelled', 'done'])
+            ])
+            if conflicting_picking:
+                raise ValidationError(
+                    _("This pallet is already being used for picking %s.") %
+                    conflicting_picking[0].name)
+        else:
+            conflicting_batch = PickingBatch.search([
+                ('id', '!=', self.id),
```
ws-mps, line 1059: Do we have an index for this search?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1017,21 +1033,45 @@ def reserve_pallet(self, pallet_name):
...
+                ('state', '=', 'in_progress'),
+                ('u_last_reserved_pallet_name', '=', pallet_name),
+            ])
+            if conflicting_batch:
+                raise ValidationError(
+                    _("This pallet is already being used for batch %s.") %
+                    conflicting_batch[0].name)
+
+        if reserve_pallet_per_picking:
+            # The front end will always add a picking to a batch and then
```
ws-mps, line 1069: Not strictly true, we can create a batch with 4 and assign 4. The thing is that reserve_pallet will receive only one picking as parameter, that is the one we are updating.

addons/udes_stock/README.md
```diff
@@ -126,6 +126,8 @@ A lot of custom UDES functionality is specfied at the picking type level. This i
 | u_enable_confirmations     | boolean | Flag to indicate if user has to perform any confirmation after picking a product.|
 | u_use_part_pallets         | boolean | Flag to indicate if each pallet is validated and processed at goods in as oppose scanning the whole ASN.|
 | u_auto_unlink_empty        | boolean | Flag to indicate whether to unlink empty pickings when searching for any empty picking in the system.|
+| u_reserve_pallet_per_picking | boolean | Flag to indicate whether pallets are reserved on pickings rather than pallets. |
```
ws-mps, line 129: typo: rather than batches.

addons/udes_stock/controllers/stock_picking_batch.py
```diff
@@ -192,6 +192,8 @@ def reserve_pallet(self, ident, pallet_name):
 
         @param pallet_name - Barcode of the pallet to be reserved.
         """
+        StockPicking = request.env['stock.picking']
```
ws-mps, line 195: Nit: `Picking` for consistency with other code?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+        cls.pack_4apples_info = [{'product': cls.apple,
+                                  'qty': 4}]
+        cls.pack_4bananas_info = [{'product': cls.banana,
+                                  'qty': 4}]
+
+        cls.picking_type_pick.u_reserve_pallet_per_picking = True
+
+    def setUp(self):
+        super().setUp()
+        self.create_quant(self.apple.id,
```
ws-mps, line 1724: If we are creating the same quants for every test, can we just add it to setUpClass?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+                          300,
+                          package_id=self.create_package().id)
+        self.create_quant(self.banana.id,
+                          self.test_location_02.id,
+                          300,
+                          package_id=self.create_package().id)
+
+    def tearDown(self):
+        super().tearDown()
+        self.picking_type_pick.u_max_reservable_pallets = 10
```
ws-mps, line 1735: Is this really needed? any change to the picking type should be rolled back when test ends.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+        return super().create_batch(user=user, u_ephemeral=True, **kwargs)
+
+    def test01_associates_pallet_with_initial_batch_picking(self):
+        batch = self.create_batch(user=self.outbound_user)
+        picking1 = self.create_picking(self.picking_type_pick,
+                                       products_info=self.pack_4apples_info,
+                                       confirm=True,
+                                       assign=True,
+                                       batch_id=batch.id)
+        batch.state = 'in_progress'
```
ws-mps, line 1747: Nit: we can call `batch.mark_as_todo()` which will compute the state of the batch, which should be in progress.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+        batch.reserve_pallet('UDES11111', picking=picking1)
+
+        self.assertEqual(picking1.u_reserved_pallet, 'UDES11111')
+
+    def test02_associates_pallet_with_batch(self):
```
ws-mps, line 1752: Can we find a better name? We are reserving two pallets to two pickings not to a batch.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+    def tearDown(self):
+        super().tearDown()
+        self.picking_type_pick.u_max_reservable_pallets = 10
+
+    def create_batch(self, user=None, **kwargs):
+        return super().create_batch(user=user, u_ephemeral=True, **kwargs)
+
+    def test01_associates_pallet_with_initial_batch_picking(self):
+        batch = self.create_batch(user=self.outbound_user)
+        picking1 = self.create_picking(self.picking_type_pick,
```
ws-mps, line 1742: nit: `picking`

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+                                       products_info=self.pack_4apples_info,
+                                       confirm=True,
+                                       assign=True,
+                                       priority='0')
+        picking2 = self.create_picking(self.picking_type_pick,
+                                       products_info=self.pack_4bananas_info,
+                                       confirm=True,
+                                       assign=True,
+                                       priority='1')
+        pickings = [pickings.id, picking2.id]
```
ws-mps, line 1909: nit: `picking_ids = `

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,213 @@ def test04_rereservation_for_same_batch(self):
...
+    def test10_raises_exception_if_batch_created_with_more_than_max_reservable_pallets(self):
+        Batch = self.env['stock.picking.batch']
+
+        self.picking_type_pick.u_max_reservable_pallets = 1
+        pickings = self.create_picking(self.picking_type_pick,
+                                       products_info=self.pack_4apples_info,
+                                       confirm=True,
+                                       assign=True,
+                                       priority='0')
+        picking2 = self.create_picking(self.picking_type_pick,
```
ws-mps, line 1904: If we don't care about  individual pickings, we can use:
`pickings |=  XXX` 
then we don't need line 1909 and we can use pickings.ids in line 1917.

addons/udes_stock/views/stock_picking_type.xml
```diff
@@ -56,6 +56,7 @@
                     <field name="u_num_reservable_pickings" />
                     <field name="u_reserve_batches" />
                     <field name="u_auto_unlink_empty" />
+                    <field name="u_reserve_pallet_per_picking" />
```
ws-mps, line 59: Do we want to add the other field?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -1006,32 +1022,61 @@ def mark_as_todo(self):
...
-        conflicting_batch = PickingBatch.search([
-            ('id', '!=', self.id),
-            ('state', '=', 'in_progress'),
-            ('u_last_reserved_pallet_name', '=', pallet_name),
-        ])
-        if conflicting_batch:
-            raise ValidationError(
-                _("This pallet is already being used for batch %s.") %
-                conflicting_batch[0].name)
+        reserve_pallet_per_picking = self.picking_type_ids.filtered('u_reserve_pallet_per_picking')
```
ws-mps, line 1048: Does this work without lambda?

addons/udes_stock/models/stock_picking_batch.py
kdwyer: No - this search was already present, it's been indented by my changes.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,193 @@ def test04_rereservation_for_same_batch(self):
...
+                                       products_info=self.pack_4bananas_info,
+                                       confirm=True,
+                                       assign=True)
+        batch.add_extra_pickings(self.picking_type_pick.id)
+        batch.reserve_pallet('UDES11112', picking=picking2)
+
+        self.assertEqual(sorted(batch.mapped('picking_ids.u_reserved_pallet')), ['UDES11111', 'UDES11112'])
+
+    def test03_associates_pallet_with_added_extra_picking(self):
```
ws-mps, line 1771: Isn't this test the same as test02 but just with a different assert at the end?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,193 @@ def test04_rereservation_for_same_batch(self):
...
+        picking2 = self.create_picking(self.picking_type_pick,
+                                       products_info=self.pack_4bananas_info,
+                                       confirm=True,
+                                       assign=True)
+        batch.add_extra_pickings(self.picking_type_pick.id)
+        batch.reserve_pallet('UDES11112', picking=picking2)
+
+        self.assertEqual(picking2.u_reserved_pallet, 'UDES11112')
+
+    def test04_raises_error_on_reservation_conflict(self):
```
ws-mps, line 1784: suggestion: test04_raises_error_on_reservation_conflict -> test04_raises_error_on_reservation_conflict_different_batch ?

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1705,3 +1705,193 @@ def test04_rereservation_for_same_batch(self):
...
+        picking3 = batch2.picking_ids
+
+        expected_error = 'This pallet is already being used for picking %s.' % \
+                         picking1.name
+
+        with self.assertRaisesRegex(ValidationError, expected_error,
+                                    msg='Incorrect error thrown'):
+            batch2.reserve_pallet('UDES11111', picking=picking3)
+
+    def test05_rereservation_for_same_batch(self):
```
ws-mps, line 1804: suggestion: test05_raises_error_on_reservation_conflict_same_batch

kdwyer: Much better; I should apply my preferred naming convention (test_verb_condition) when copying tests from other classes :smile:

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -335,6 +335,18 @@ class StockPickingType(models.Model):
...
     )
 
+    u_reserve_pallet_per_picking = fields.Boolean(
+        string="Reserve one pallet per picking",
+        default=False,
+        help="If enabled, each picking in a batch will be associated with an "
+        "individual pallet")
+
+    u_max_reservable_pallets = fields.Integer(
+        string="Maximum pallets that may be simultaneoulsy reserved in a batch.",
```
squah-u, line 345: typo: `simultaneously`

addons/udes_stock/README.md
squah-u: should this say `batches`?

PR #321, opened 2020-07-20
[ADD] udes_stock: Suggest locations by product lot numbers

williamsajr:


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -244,6 +244,7 @@ class StockPickingType(models.Model):
                 "Exactly Match The Move Line Destination Location",
             ),
             ("by_products", "By Products"),
+            ("by_product_lot", "By Product and Batch Number"),
```
PeterC10, line 247: Should this be Batch Number or Lot Number? I know they're interchangeable but I've seen a few instances of lot number on the desktop UI but not batch.

ws-mps: We should check this with PO.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1879,6 +1879,51 @@ def _get_suggested_location_by_products(self, move_line_ids, products=None):
...
+
+        quant_domain = [
+            ("product_id", "in", products.ids),
+            ("location_id", "child_of", self.location_dest_id.ids),
+            ("location_id.u_blocked", "=", False),
+            ("location_id.barcode", "!=", False),
+        ]
+        if lots:
+            quant_domain.append(("lot_id", "in", lots.ids))
+        else:
```
ws-mps, line 1916: Is this really needed ?

addons/udes_stock/models/stock_picking.py
```diff
@@ -1879,6 +1879,51 @@ def _get_suggested_location_by_products(self, move_line_ids, products=None):
 
         return suggested_locations
 
+    def _get_suggested_location_by_product_lot(self, move_line_ids, products=None):
```
ws-mps, line 1882: If the main difference is adding the lot to the search domain, I would suggest to merge product and product by lot into one with extra func.

addons/udes_stock/models/stock_picking.py
williamsajr: I guess technically not. The only case it may be useful is as a failsafe incase a product has changed from not tracked to lot-tracked. This way quants without lots are lumped together rather than being mixed with the those with lots

addons/udes_stock/models/stock_picking.py
williamsajr: Unfortunately due to the way these functions are called is by looking for a function with the name ```_get_suggested_location_by```_by policy_ so we'd have to change quite a lot to get them to use a single function. I could potentially call ```_get_suggested_location_by_product``` from ```_get_suggested_location_by_product_lot``` and filter those locations but I think that would lead to a less efficient function as I would need to do some re-work (re-aquire the quants and filter)

addons/udes_stock/models/stock_picking_type.py
williamsajr: Spoke to product and will change all references of batch to lot

addons/udes_stock/models/stock_picking.py
williamsajr: Removed this as it's probably too much of an edge case

addons/udes_stock/models/stock_picking.py
williamsajr: Added a new function called by both policies

addons/udes_stock/models/stock_picking.py
```diff
@@ -1863,14 +1885,25 @@ def _get_suggested_location_by_products(self, move_line_ids, products=None):
...
-            ]
-        ).mapped("location_id")
+        if match_lots:
+            if len(products) > 1:
+                raise ValidationError(_("Cannot drop different products by lot number."))
+            lot_tracking = products.tracking == "lot"
+            if lot_tracking:
+                lots = move_line_ids.filtered(lambda ml: ml.product_id in products).mapped("lot_id")
+                if len(lots) != 1:
+                    raise ValidationError(_("Expecting a single lot number for a package/pallet "
```
ws-mps, line 1895: I don't think we need the `package/pallet` reference here.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1863,14 +1885,25 @@ def _get_suggested_location_by_products(self, move_line_ids, products=None):
         if not products:
             raise ValidationError(_("Products missing to suggest location for."))
```
PeterC10, line 1886: Nit: Error message reads a bit oddly to me. Maybe "No products to suggestion location for." instead? Feel free to resolve though, probably just me. :)

williamsajr: I'm going to leave this one as it is as it was a pre-existing error message

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+        cls.test_stock_quant_03 = cls.create_quant(
+            cls.apple.id,
+            cls.test_location_03.id,
+            10,
+        )
+
+        # Create a new lot tracked product
+        cls.uglyfruit = cls.create_product('Ugly Fruit', tracking='lot')
+
+    def create_and_assign_putaway_pick(
```
ws-mps, line 215: nit: `_putaway_picking`

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+            product.id,
+            self.received_location.id,
+            10,
+        )
+        if lot:
+            test_quant.lot_id = lot
+
+        products_info = [{'product': product, 'qty': 10}]
+        self.picking_type_putaway.u_drop_location_policy = drop_policy
+        test_picking = self.create_picking(self.picking_type_putaway,
```
ws-mps, line 232: nit: just `picking` instead of `test_picking` is enough.
If we want to make clear the type: `putaway_picking`

Applies to the tests below.

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+        with self.assertRaises(ValidationError) as e:
+            test_picking.get_suggested_locations(test_picking.move_line_ids)
+            self.assertEqual(
+                e.exception.name,
+                "Expecting a single lot number "
+                "when dropping by product and lot."
+            )
+
+    def test04_get_suggested_location_by_product_lot_multiple_products(self):
+        """Test that an error is raised if we use want to suggest a location
```
ws-mps, line 311: typo: `if we use want to`

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+        test_tangerine_quant_1.lot_id = self.test_stock_quant_01.lot_id
+        test_tangerine_quant_2 = self.create_quant(
+            self.tangerine.id,
+            self.received_location.id,
+            10,
+        )
+        test_tangerine_quant_2.lot_id = self.test_stock_quant_02.lot_id
+
+        products_info = [{'product': self.tangerine, 'qty': 20}]
+        self.picking_type_putaway.u_drop_location_policy = "by_product_lot"
```
ws-mps, line 293: Can we unify how we create pickings? In some tests we use the aux function and sometimes we don't.

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+            10,
+        )
+        test_quant.lot_id = self.test_stock_quant_01.lot_id
+        uf_lot = self.create_lot(self.uglyfruit.id, "TEST_UF_LOT001")
+        test_quant = self.create_quant(
+            self.uglyfruit.id,
+            self.received_location.id,
+            10,
+        )
+        test_quant.lot_id = uf_lot
```
ws-mps, line 326: We should do this on the create_quant.

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,241 @@ def test12_package_has_no_parent_package(self):
...
+            test_picking.move_line_ids
+        )
+
+        # Assert that only a single location is returned;
+        # the empty location.
+        self.assertEqual(len(suggested_locations), 1)
+        self.assertTrue(self.test_location_04 in suggested_locations)
+
+    def test03_get_suggested_location_by_product_lot_multiple_lots(self):
+        """Test that an error is raised if we use want to suggest a location
```
ws-mps, line 276: Typo: `if we use want to`

addons/udes_stock/tests/test_picking.py
williamsajr: Changed so that all tests use the aux function. This function was changed to accept a list of dicts that are used to setup the quants/movelines.

addons/udes_stock/tests/test_picking.py
```diff
@@ -167,3 +167,213 @@ def test12_package_has_no_parent_package(self):
...
+        for info in products_info:
+            test_quant = self.create_quant(
+                info["product"].id,
+                self.received_location.id,
+                info["qty"],
+            )
+            if info.get("lot"):
+                test_quant.lot_id = info["lot"]
+            # Remove lot from dictionary (if present) so that it may be used in create_picking
+            info.pop("lot", None)
```
ws-mps, line 230: We can move this line to before line 227:
`lot = info.php("lot", None)`
And then check `if lot:` instead of get.

williamsajr: Good spot

PR #322, opened 2020-07-23
Story/10704

williamsajr:
Change to add lots to stock reports. I advice looking at the commits in turn to understand the reasoning a bit better

I re-factored how the reports are created by splitting data gathering and writing to a file. This was done for a few reason:
- The function was a bit of a mess. It's (slightly) less of a mess now 
- To allow testing to be done on the data without having to try read a .xls file
- To make the functions more general, with the potential to export the data in a non xls format in future by just creating a new write function (such as csv)

I also plan on running this module through black once approved

### DIFF
--- a/addons/udes_report/models/stock_export.py
+++ b/addons/udes_report/models/stock_export.py
@@ -62,94 +62,78 @@ class StockExport(models.TransientModel):
             1) part number, location, package ID and quantity;
             2) part number, number of packages and total quantity.
         '''
-        locations = self.included_locations - self.excluded_locations
+        file_name = 'warehouse_stock_{TIMESTAMP}.xls'.format(TIMESTAMP=datetime.now())
 
+        data = self._generate_stock_file_data()
+        self._write_workbook(file_name, "Stock File", data, send_file_via=send_file_via)
+
+    def _generate_stock_file_data(self):
+        '''
+        Generate data for the stock file.
+
+        This will contain 2 sheets that may then be exported to any format
+        (currently only excel)
+        Sheets:
+            1) part number, location, package ID and quantity;
+            2) part number, number of packages and total quantity.
+        '''
+        Quant = self.env['stock.quant']
+        Product = self.env['product.product']
+
+        locations = self.included_locations - self.excluded_locations
         if not locations:
             raise UserError(_("The specified list of Stock Locations is empty."))
 
-        Quant = self.env['stock.quant']
-        Product = self.env['product.product']
-        Location = self.env['stock.location']
         all_quants = Quant.search([('location_id', 'child_of', locations.ids)])
-
-        by_prod = lambda x: x.product_id.id
-        by_location = lambda x: x.location_id.id
-        by_package_name = lambda x: x.package_id.name or ''
-        by_lot_name = lambda x: x.lot_id.name or ''
-
-        get_quants_by_loc = lambda quant_union: {
-            Location.browse(loc): Quant.union(*_quants)
-            for loc, _quants in groupby(quant_union.sorted(key=by_location),
-                                        key=by_location)}
 
         quants_by_prod_by_loc = OrderedDict()
         prod_summary = defaultdict(lambda: {'n_pkgs': 0, 'qty': 0})
 
-        # Determine if lots should be shown on the report
-        show_lots = False
-        if Product.search([("tracking", "=", "lot")]):
-            show_lots = True
+        for prod, quants in all_quants.groupby('product_id'):
+            quants_by_prod_by_loc[prod] = {loc: grouped_quants
+                                           for loc, grouped_quants in quants.groupby("location_id")}
 
-        for prod, _quants in groupby(all_quants.sorted(key=by_prod),
-                                     key=by_prod):
-            quant_prods = Product.browse(prod)
-            quants_u = Quant.union(*_quants)
-            quants_by_prod_by_loc[quant_prods] = get_quants_by_loc(quants_u)
+        # 1/2) 'Stock File': one entry for each prod, package and lot (if applicable)
+        stock_file_data = {"name": 'Stock File'}
+        # Determine if lots/serials should be shown on the report
+        show_tracking = bool(Product.search([("tracking", "!=", "none")], order="id", limit=1))
 
-        def get_prod_qty(quants):
-            return sum(Quant.union(*quants).mapped('quantity'))
+        headers = ['Part Number', 'Location', 'Package', 'Quantity']
+        if show_tracking:
+            headers.insert(3, 'Lot/Serial Number')
+        stock_file_data["column_titles"] = headers
 
-        _logger.info(_('Creating Excel file'))
-        file_name = 'warehouse_stock_{TIMESTAMP}.xls'.format(
-            TIMESTAMP=datetime.now())
-        wb = xlwt.Workbook()
-
-        # 1/2) 'Stock File': one entry for each (prod, package, lot) combination
-        if show_lots:
-            row_titles = ['Part Number', 'Location', 'Package', 'Lot Number', 'Quantity']
-        else:
-            row_titles = ['Part Number', 'Location', 'Package', 'Quantity']
-
-        stock_sheet = _create_sheet(wb, 'Stock File', row_titles)
-
-        row = 0
+        rows = []
         for prod, quants_by_location in quants_by_prod_by_loc.items():
             for location, quants in quants_by_location.items():
-                for pkg_name, grouped_quants in groupby(quants.sorted(key=by_package_name),
-                                                        key=by_package_name):
-                    # If lots are not present then this will not change the rows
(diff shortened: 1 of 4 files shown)


davidmcgaw:
Tests LGTM!

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,103 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
+
+        all_quants = Quant.search([('location_id', 'child_of', locations.ids)])
 
         quants_by_prod_by_loc = OrderedDict()
         prod_summary = defaultdict(lambda: {'n_pkgs': 0, 'qty': 0})
 
-        for prod, _quants in groupby(all_quants.sorted(key=by_prod),
-                                     key=by_prod):
+        # Determine if lots/serials should be shown on the report
+        show_tracking = bool(Product.search([("tracking", "!=", "none")]))
```
apatrick88, line 111: Use `search_count` here instead of `search`.

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,103 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
-                                                       'Package',
-                                                       'Quantity'])
+        if show_tracking:
+            stock_file_data["column_titles"] = ['Part Number',
+                                                'Location',
+                                                'Package',
+                                                'Lot/Serial Number',
+                                                'Quantity']
+        else:
+            stock_file_data["column_titles"] = ['Part Number', 'Location', 'Package', 'Quantity']
```
apatrick88, line 127: Can we split this out to avoid repeating common columns? If we want to maintain column order:
```
headers = ['Part Number', 'Location', 'Package']

if show_tracking:
   headers.append('Lot/Serial Number')

headers.append('Quantity')

stock_file_data["column_titles"] = headers
```

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,103 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
+                               'Quantity': qty}
+                        if show_tracking:
+                            row['Lot/Serial Number'] = lot_name
+                        rows.append(row)
+                        prod_summary[prod]['qty'] += qty
                     if pkg_name:
                         prod_summary[prod]['n_pkgs'] += 1
+        stock_file_data['rows'] = rows
 
         # 2/2) 'Stock Summary': one entry for each prod
```
apatrick88, line 150: I think the counter part to this `# 1/2 ....` is now missing.

addons/udes_report/models/stock_export.py
davidmcgaw: alternatively you could use insert to add in the lot column

addons/udes_report/models/stock_export.py
williamsajr: Changed

addons/udes_report/models/stock_export.py
williamsajr: Changed as suggested

addons/udes_report/models/stock_export.py
williamsajr: Good spot, added an edited version back in

addons/udes_report/models/stock_export.py
```diff
@@ -165,35 +199,47 @@ def run_movement_file_export(self, context=None, send_file_via='user'):
...
+            data.append(move_data)
 
-        self._write_workbook(wb, file_name, "Movement File",
-                             send_file_via=send_file_via)
+        return data
 
     @api.model
-    def _write_workbook(self, workbook, file_name, doc_title,
-                        send_file_via='user'):
+    def _write_workbook(self, file_name, doc_title, data, send_file_via="user"):
```
ws-mps, line 223: Data is an existing variable that is used later (line 197). I would suggest to rename one of them.

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,99 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
-        _logger.info(_('Creating Excel file'))
-        file_name = 'warehouse_stock_{TIMESTAMP}.xls'.format(
-            TIMESTAMP=datetime.now())
-        wb = xlwt.Workbook()
-
-        # 1/2) 'Stock File': one entry for each (prod, package) pair
+        # 1/2) 'Stock File': one entry for each prod, package and lot (if applicable)
+        stock_file_data = {"name": 'Stock File'}
+        # Determine if lots/serials should be shown on the report
+        show_tracking = bool(Product.search_count([("tracking", "!=", "none")]))
```
PeterC10, line 118: Nit: Can also add a ```limit=1``` and ```order=id desc``` as you just need to know if any results are returned, would probably only give the most minimal of performance increases though. :)

PeterC10: Actually as it's ```search_count``` then ```order``` probably isn't relevant. Feel free to resolve anyway. :)

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,99 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
-                    row += 1
-                    qty = get_prod_qty(grouped_quants)
-
-                    stock_sheet.write(row, 0, prod.default_code)
-                    stock_sheet.write(row, 1, location.display_name)
-                    stock_sheet.write(row, 2, pkg_name)
-                    stock_sheet.write(row, 3, qty)
-
-                    prod_summary[prod]['qty'] += qty
+                    # If lots are not present then this will not change the rows
```
ws-mps, line 130: Can we rephrase this comment please?
Line 138 is where rows are not going to get changed if lots not present, this for will at least get on group of quants from the groupby, that is the quants without lot name.

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,99 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
+                    for lot_name, lot_grouped_quants in groupby(grouped_quants,
```
ws-mps, line 131: We can now replace `groupby(grouped_quants, key=by_lot_name)` by `grouped_quants.grupby(by_lot_name)` and `lot_grouped_quants` will be a recordset already, so we don't need `Quant.union` at line 133, and we can do something like: `lot_grouped_quants.total_quantity()` which does the sum.

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,99 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
+        if show_tracking:
+            headers.insert(3, 'Lot/Serial Number')
+        stock_file_data["column_titles"] = headers
 
-        row = 0
+        rows = []
         for prod, quants_by_location in quants_by_prod_by_loc.items():
             for location, quants in quants_by_location.items():
-                for pkg_name, grouped_quants in groupby(quants,
+                for pkg_name, grouped_quants in groupby(quants.sorted(key=by_package_name),
```
ws-mps, line 128: Same comment as line 131.

addons/udes_report/models/stock_export.py
```diff
@@ -62,79 +62,99 @@ def run_stock_file_export(self, context=None, send_file_via='user'):
...
+            raise UserError(_("The specified list of Stock Locations is empty."))
+
+        all_quants = Quant.search([('location_id', 'child_of', locations.ids)])
 
         quants_by_prod_by_loc = OrderedDict()
         prod_summary = defaultdict(lambda: {'n_pkgs': 0, 'qty': 0})
 
-        for prod, _quants in groupby(all_quants.sorted(key=by_prod),
-                                     key=by_prod):
+        for prod, _quants in groupby(all_quants.sorted(key=by_prod), key=by_prod):
```
ws-mps, line 110: We can use new groupby so we don't need union.

addons/udes_report/models/stock_export.py
williamsajr: Renamed

addons/udes_report/models/stock_export.py
williamsajr: Changed to 'Product.search([("tracking", "!=", "none")], order="id", limit=1)' as discussed on skype

addons/udes_report/models/stock_export.py
williamsajr: Removed comment as it was probably unnecessary.

 Also changed to use groupby

addons/udes_report/models/stock_export.py
williamsajr: Changed, looks much better

addons/udes_report/models/stock_export.py
williamsajr: Changed

addons/udes_report/models/stock_export.py
williamsajr: Changed

PR #323, opened 2020-07-24
[FIX] udes_closed: Unpickable items in two pickings

kdwyer:
User-story/10878
Task/10880

`StockPickingBatch.unpickable_item` fails if there is more than one
unpickable picking in a batch for the product.

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -759,7 +759,6 @@ class StockPickingBatch(models.Model):
                   'Press back when resolved.')
             )
 
-
     def unpickable_item(self,
                         reason,
                         product_id=None,
@@ -770,11 +769,8 @@ class StockPickingBatch(models.Model):
         """
         Given an unpickable product or package, find the related
         move lines in the current batch, backorder them and refine
-        the backorder (by default it is canceled).
-        Then create a new stock investigation picking for the
-        unpickable stock.
-        If the picking was the last one of the batch, the batch is
-        set as done.
+        the backorder. Then create a new stock investigation picking
+        for the unpickable stock.
 
         An unpickable product requires at least the location_id and
         optionally the package_id and lot_name.
@@ -782,6 +778,7 @@ class StockPickingBatch(models.Model):
         self.ensure_one()
 
         Package = self.env['stock.quant.package']
+        Picking = self.env['stock.picking']
         Location = self.env['stock.location']
         Product = self.env['product.product']
 
@@ -792,7 +789,6 @@ class StockPickingBatch(models.Model):
         package = Package.get_package(package_name) if package_name else None
         allow_partial = False
         move_lines = self.get_available_move_lines()
-
         if product:
             if not location:
                 raise ValidationError(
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -773,15 +772,14 @@ def unpickable_item(self,
         the backorder (by default it is canceled).
```
ws-mps, line 772: By default is not cancelled anymore, can you remove this although it is not related to your changes, please?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -834,29 +831,36 @@ def unpickable_item(self,
...
 
-        moves = picking.move_lines
-        if raise_stock_investigation:
+            if raise_stock_investigation:
+                to_investigate |= picking
+
+            else:
+                picking.batch_id = False
+
+        if raise_stock_investigation and to_investigate:
```
ws-mps, line 861: Is there any case where we can have `raise_stock_investigation` and not `to_investigate` ?

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -834,29 +831,36 @@ def unpickable_item(self,
...
 
-        if len(picking.move_line_ids - move_lines):
-            # Create a backorder for the affected move lines if
-            # there are move lines that are not affected
-            original_picking_id = picking.id
-            picking = picking._backorder_movelines(move_lines)
+            if len(picking.move_line_ids - move_lines):
+                # Create a backorder for the affected move lines if
+                # there are move lines that are not affected
+                original_picking_id = picking.id
```
ws-mps, line 852: `original_picking_id` should now be a mapping from "picking -> original_picking_id" since we have multiple pickings.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -834,29 +831,36 @@ def unpickable_item(self,
...
+                picking = picking._backorder_movelines(move_lines)
 
-        moves = picking.move_lines
-        if raise_stock_investigation:
+            if raise_stock_investigation:
+                to_investigate |= picking
+
```
ws-mps, line 857: Remove empty line.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -804,24 +804,142 @@ def test27_unpickable_item_product_ok_two_picks(self):
...
         move_line = picking.move_line_ids[0]
 
         batch.unpickable_item(product_id=move_line.product_id.id,
                               location_id=move_line.location_id.id,
                               reason=reason)
-        # after unickable all the quant should be reserved
+        # after unpickable all the quant should be reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+        # picking state should be confirmed
+        self.assertEqual(picking.state, 'confirmed',
```
ws-mps, line 883: Do we want to check this for both pickings? (Same for line 886)

addons/udes_stock/models/stock_picking_batch.py
kdwyer: There isn't, but `Picking.raise_stock_inv` gets upset if it's called on an empty recordset.

ws-mps: Which I think it should never happen, since to_investigate comes from pickings which comes from move_lines for which we raise an error if it is empty.

addons/udes_stock/models/stock_picking_batch.py
kdwyer: Done

addons/udes_stock/tests/test_picking_batch.py
kdwyer: Done

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -834,6 +834,178 @@ def test27_unpickable_item_product_ok_two_picks(self):
...
+        """
+        Tests that calling unpickable item with a product in two pickings, one
+        of which also includes a line for another product, preserves the move
+        for the pickable product.
+        """
+        Picking = self.env['stock.picking']
+        Batch = self.env['stock.picking.batch']
+        Batch = Batch.sudo(self.outbound_user)
+
+        quant = self.create_quant(self.apple.id, self.test_location_01.id, 4)
```
PeterC10, line 961: Nit: Could make the variables more descriptive, e.g. `apple_quant` and `banana_quant`. Same for products info and picking variables.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -834,6 +834,178 @@ def test27_unpickable_item_product_ok_two_picks(self):
...
+                                      assign=True)
+
+        products_info2 = [{'product': self.apple, 'qty': 1},
+                          {'product': self.banana, 'qty': 1}]
+        picking_2 = self.create_picking(self.picking_type_pick,
+                                        products_info=products_info2,
+                                        confirm=True,
+                                        assign=True)
+        pickings = picking | picking_2
+        print('TEST PICKINGS', pickings)
```
PeterC10, line 977: Wild print statement

kdwyer: Eliminated, thanks.

addons/udes_stock/tests/test_picking_batch.py
kdwyer: I've removed the `_2` suffix and prefixed those variable names with `mixed_` -  I think what we want to express is that there is more than one product involved, rather than the specific product names.  Thanks for the nudge to make this improvement.

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -865,29 +869,27 @@ def unpickable_item(self,
...
+                moves = picking.move_lines
+                if picking.exists() \
+                        and picking.state == 'assigned' \
+                        and original_picking_ids[picking] is not None\
+                        and not picking.picking_type_id.u_post_assign_action:
+                    # A backorder has been created, but the stock is
+                    # available; get rid of the backorder after linking the
+                    # move lines to the original picking, so it can be
+                    # directly processed
+                    picking.move_line_ids.write({'picking_id': original_picking_id})
```
ws-mps, line 882: This is still using the old `original_picking_id`.
We can just create a var before line 874 `original_picking_id = original_picking_ids[picking]` and use it also in line 876.

PR #325, opened 2020-07-30
[FIX] udes_open: Do not suggest locations with partially available lines

kdwyer:
User-story/10682
Task/10863

The suggested locations feature should consider partially available move
lines when determining available locations.

### DIFF
--- a/addons/udes_stock/tests/test_picking.py
+++ b/addons/udes_stock/tests/test_picking.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from . import common
-from odoo.exceptions import UserError, ValidationError
+from odoo.exceptions import ValidationError
 
 class TestGoodsInPicking(common.BaseUDES):
 
@@ -176,7 +176,6 @@ class TestSuggestedLocation(common.BaseUDES):
         super(TestSuggestedLocation, cls).setUpClass()
 
         Location = cls.env['stock.location']
-        Picking = cls.env['stock.picking']
 
         cls.test_location_03 = Location.create({
             "name": "Test location 03",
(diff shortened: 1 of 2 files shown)


addons/udes_stock/tests/test_picking.py
```diff
@@ -376,3 +375,30 @@ def test08_get_suggested_location_by_product_all_lots(self):
         # and the location with matching product but different lot.
         self.assertEqual(len(suggested_locations), 2)
         self.assertEqual(self.test_location_01 + self.test_location_02, suggested_locations)
+
+    def test09_considers_partially_available_move_lines_when_suggesting_location(self):
```
joe-gb, line 379: I realise the method name pretty much spells it out but please could we add a docstring briefly describing the tests as we do with the others?

addons/udes_stock/tests/test_picking.py
```diff
@@ -376,3 +375,30 @@ def test08_get_suggested_location_by_product_all_lots(self):
...
+        self.test_location_05 = Location.create({
+            "name": "Test location 05",
+            "barcode": "LTEST05",
+            "location_id": self.stock_location.id,
+        })
+
+        picking1 = self.create_and_assign_putaway_picking(products_info, drop_policy)
+        picking2 = self.create_and_assign_putaway_picking(products_info, drop_policy)
+        picking1.apply_drop_location_policy()
+        picking1.move_line_ids.write({'state': 'partially_available'})
```
ws-mps, line 399: It would be good to get it in partially_available without changing the state manually. Have we tried to increase the move quantity by one unit? something like `picking1.move_lines[0].product_uom_qty += 1`. Otherwise I would add a comment that we are changing the state manually just to keep the test simpler.

addons/udes_stock/tests/test_picking.py
kdwyer: Added, please review

addons/udes_stock/tests/test_picking.py
kdwyer: Changed, please review.

addons/udes_stock/tests/test_picking.py
```diff
@@ -376,3 +375,33 @@ def test08_get_suggested_location_by_product_all_lots(self):
...
+        self.test_location_05 = Location.create({
+            "name": "Test location 05",
+            "barcode": "LTEST05",
+            "location_id": self.stock_location.id,
+        })
+
+        picking1 = self.create_and_assign_putaway_picking(products_info, drop_policy)
+        picking2 = self.create_and_assign_putaway_picking(products_info, drop_policy)
+        picking1.apply_drop_location_policy()
+        picking1.move_lines[0].product_uom_qty += 1
```
ws-mps, line 402: Shall we assert that the state is now partially available ?

kdwyer: asserted

PR #326, opened 2020-07-30
Story/10703 - Restrict Multi Lot Transfers flag on Picking Types

PeterC10:
Added Restrict Multi Lot Transfers flag to Picking Types. When validating a picking a check is carried out to see if the Picking Type restricts multi lot transfers. If it does and the move lines contain more than one unique lot an exception is raised.

Added Unit Test for multi lot picking restriction in TestPickingType.

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -279,6 +279,46 @@ class StockPicking(models.Model):
                     % picking.name
                 )
 
+    def assert_not_lot_restricted(self):
+        """
+        Assert that the pickings in self are not restricted based on lot.
+        If picking operation type restricts multi lot transfers then check for number of unique
+        lots in move lines and raise an error if picking contains more than one. Pickings that
+        contain a mix of set lot and empty lot move lines will also raise an error.
+        """
+        restrict_multi_lot_pickings = self.filtered(
+            lambda p: p.picking_type_id.u_restrict_multi_lot_pickings
+        )
+
+        for pick in restrict_multi_lot_pickings:
+            # Get the move lines tracked by lot or by none
+            non_serial_move_line_ids = pick.move_line_ids.filtered(
+                lambda ml: ml.product_id.tracking != "serial"
+            )
+
+            if non_serial_move_line_ids:
+                # Count the number of unique lots in the move lines
+                lots = non_serial_move_line_ids.mapped("lot_id")
+                lot_count = len(lots)
+
+                if lot_count:
+                    if lot_count > 1:
+                        raise ValidationError(_("Cannot validate transfer with multiple lots."))
+
+                    # Count the move lines without a lot set
+                    empty_lot_move_line_ids = non_serial_move_line_ids.filtered(
+                        lambda ml: not ml.lot_id
+                    )
+                    empty_lot_count = len(empty_lot_move_line_ids)
+
+                    if lot_count and empty_lot_count:
+                        raise ValidationError(
+                            _(
+                                "Cannot validate transfer with both set lot and empty lot "
+                                "move lines."
+                            )
+                        )
+
     def action_assign(self):
         """
         Unlink empty pickings after action_assign, as there may be junk data
@@ -299,6 +339,7 @@ class StockPicking(models.Model):
         # Prevent recomputing the batch stat
         batches = mls.mapped("picking_id.batch_id")
         next_pickings = self.mapped("u_next_picking_ids")
+
         res = super(StockPicking, self.with_context(lock_batch_state=True)).action_done()
 
         # just in case move lines change on action done, for instance cancelling
@@ -308,6 +349,8 @@ class StockPicking(models.Model):
         # batches of the following stage should also be recomputed
         all_picks = picks | picks.mapped("u_next_picking_ids")
         all_picks.with_context(lock_batch_state=False)._trigger_batch_state_recompute()
+
+        self.assert_not_lot_restricted()
 
         extra_context = {}
         if hasattr(self, "get_extra_printing_context"):
--- a/addons/udes_stock/tests/test_picking_type.py
+++ b/addons/udes_stock/tests/test_picking_type.py
@@ -630,3 +630,204 @@ class TestPickingType(common.BaseUDES):
         picking.move_line_ids.write({
             'location_dest_id': self.test_location_02.id
         })
+
+    def test20_assert_single_lot_picking_validation_allowed(self):
+        """With a picking type set to restrict multi lot pickings, assert
+           that picking with only one lot in its move lines can still be
+           validated.
+        """
+        tangerine_qty = 20
+
+        # Set internal transfer to restrict multi lot pickings
+        self.picking_type_internal.u_restrict_multi_lot_pickings = True
+
+        # Create lots for tangerine product
+        tangerine_lot = self.create_lot(self.tangerine.id, "T1")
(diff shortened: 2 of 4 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -299,6 +299,14 @@ def action_done(self):
         # Prevent recomputing the batch stat
         batches = mls.mapped("picking_id.batch_id")
         next_pickings = self.mapped("u_next_picking_ids")
+
+        # If operation type does not allow multiple lots in a single transfer then check for number 
+        # of different lots in move lines and raise an error if more than one
+        if self.picking_type_id.u_restrict_multi_lot_pickings:
+            lots = mls.mapped("lot_id")
```
nocturnalastro, line 306: When does lot_name get turned into a lot because if its after this we may have a bug.

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -630,3 +630,108 @@ def test19_enforce_with_empty_preprocessed_location(self):
...
+
+        multi_lot_picking_strawberry_move_line.qty_done = strawberry_qty
+        multi_lot_picking_tangerine_move_line.qty_done = tangerine_qty
+
+        # Validate the multi lot picking which should raise an exception as it contains
+        # multiple lots
+        with self.assertRaises(ValidationError) as error:
+            multi_lot_picking.action_done()
+
+        expected_error_message = 'Cannot validate transfer with multiple lots.'
```
nocturnalastro, line 736: I think assert raises allows you to specify the message.

addons/udes_stock/tests/test_picking_type.py
```diff
@@ -630,3 +630,108 @@ def test19_enforce_with_empty_preprocessed_location(self):
         picking.move_line_ids.write({
             'location_dest_id': self.test_location_02.id
         })
+
+    def test20_assert_multi_lot_picking_validation_prevented(self):
```
nocturnalastro, line 634: This seems like two tests. One to check it works OK with one lot. One to check it fails if there's more than one.

PeterC10: Good point, will split them out.

addons/udes_stock/models/stock_picking.py
PeterC10: Will need to go look into that.

addons/udes_stock/models/stock_picking.py
```diff
@@ -299,6 +299,18 @@ def action_done(self):
...
         next_pickings = self.mapped("u_next_picking_ids")
+
+        # If operation type does not allow multiple lots in a single transfer then check for number 
+        # of different lots in move lines and raise an error if more than one
+        restrict_multi_lot_pickings = self.filtered(
+            lambda p: p.picking_type_id.u_restrict_multi_lot_pickings
+        )
+
+        for pick in restrict_multi_lot_pickings:
+            lots = pick.move_line_ids.mapped("lot_id")
```
williamsajr, line 310: One thing to bear in mind is that if we are tracking by serial number rather than lots then this will pull out all the serial numbers as well. So we either need to decide if we want to filter out products tracked by serial or edit the messages to make it clear that u_restrict_multi_lot_pickings applies to serial numbers as well

PeterC10: Hmm good point, one for me to check with product most likely. :)

williamsajr: Probably getting it to apply to both would be easier and give more functionality, would just require changing messages and possible the config name to say 'tracking' or 'lot/serial' instead of 'lot'. May be best to check with product though

williamsajr: One other thing with this is if we have one move line without tracking and one with tracking then the length would be 1 and this would pass. Is this what we want?

PeterC10: True, `mapped` won't return any False values. I guess there's not a way to force it to do this so might need a different approach here.

PeterC10: This could be worth adding as another test actually. Picking with lot and non-lot tracked products.

PeterC10: I will go check that other point with product also. I would assume we would want to count no lot as 1.

addons/udes_stock/models/stock_picking.py
PeterC10: So `lot_name` will get turned into `lot` if the lot doesn't already exist in the super call, so I will move this check after that.

addons/udes_stock/models/stock_picking.py
PeterC10: After speaking with product:

1. This should only apply to products tracked by lot, as otherwise for serial numbers we would end up with cases where products need to be moved one by one.

2. No lot should count as 1 for the lot cout, so this will need to be adjusted.

addons/udes_stock/models/stock_picking.py
PeterC10: Have now changed this so that the check is carried out after the super call of `action_done` is run.

addons/udes_stock/tests/test_picking_type.py
PeterC10: I was able to check the error message directly using `assertRaisesRegex` so have now used this in the new tests.

addons/udes_stock/tests/test_picking_type.py
PeterC10: This test has now been split into two. :)

addons/udes_stock/models/stock_picking.py
PeterC10: Have now pushed a change so that a lot restricted picking cannot contain a move line with a lot and then another with no lot set (basically for a non-lot tracked product).

addons/udes_stock/models/stock_picking.py
```diff
@@ -309,6 +310,40 @@ def action_done(self):
...
+            non_serial_move_line_ids = pick.move_line_ids.filtered(
+                lambda ml: ml.product_id.tracking != "serial"
+            )
+
+            if non_serial_move_line_ids:
+                lot_count = 0
+
+                # Count the number of unique lots in the move lines
+                lots = non_serial_move_line_ids.mapped("lot_id")
+                lot_count += len(lots)
```
williamsajr, line 330: Could we directly set lot_count = len(lots) to remove line 326?

PeterC10: Yep good spot. Problem with tinkering with the code so much, will go amend this.

PeterC10: Have now changed this.

addons/udes_stock/models/stock_picking.py
```diff
@@ -309,6 +310,41 @@ def action_done(self):
         all_picks = picks | picks.mapped("u_next_picking_ids")
         all_picks.with_context(lock_batch_state=False)._trigger_batch_state_recompute()
 
+        # If operation type does not allow multiple lots in a single transfer then check for number
```
nocturnalastro, line 313: Can we move this to another function? so we don't make action_done get bigger an harder to understand.

PeterC10: Good point. Have moved this logic into its own `assert_not_lot_restricted` function.

PR #328, opened 2020-07-30
[ADD] udes_sale_stock: Add unit tests for sale order state computation

squah-u:
Task: 10903
User-story: 10383

### DIFF
--- a/addons/udes_sale_stock/tests/test_sale_order_state.py
+++ b/addons/udes_sale_stock/tests/test_sale_order_state.py
@@ -1,0 +1,163 @@
+from . import common
+
+
+class TestSaleOrderState(common.BaseSaleUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestSaleOrderState, cls).setUpClass()
+
+        Customer = cls.env["res.partner"]
+
+        cls.customer = Customer.create(
+            {
+                "name": "Bob",
+                "street": "1 Bobs House",
+                "street2": "Bobs street",
+                "city": "The city of Bob",
+                "zip": "EN1 4LS",
+            }
+        )
+
+        cls.apple_quant = cls.create_quant(
+            cls.apple.id,
+            cls.test_location_01.id,
+            30,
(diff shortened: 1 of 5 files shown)


addons/udes_sale_stock/tests/test_sale_order_state.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+        cls.quant = cls.create_quant(
+            cls.apple.id, cls.test_location_01.id, 30, package_id=cls.create_package().id
+        )
+
+        # Create sale order
+        sale = cls.create_sale(cls.customer)
+        cls.sale_line = cls.create_sale_line(sale, cls.apple, 15)
+        sale.requested_date = "2020-01-03"
```
ws-mps, line 28: This can be added as a parameter at `create_sale`. Is this to avoid datetime conversions?

addons/udes_sale_stock/tests/test_sale_order_state.py
```diff
@@ -0,0 +1,110 @@
...
+            # Complete the picking
+            self.assertEqual(picking.state, "assigned")
+            self.complete_picking(picking)
+            self.assertEqual(picking.state, "done")
+
+        # Check that the sale order is done
+        self.assertEqual(self.sale.state, "done")
+        self.assertFalse(self.sale_line.is_cancelled)
+
+    def test02_test_last_picking_cancelled(self):
```
ws-mps, line 87: Can we add another test with a sale order of two lines and cancel one line but not the second one?

addons/udes_sale_stock/tests/test_sale_order_state.py
squah-u: It was a carry over from test_cancel_unfulfillable.py. I'll change it

addons/udes_sale_stock/tests/test_sale_order_state.py
squah-u: A new test has been added

PR #329, opened 2020-08-03
Story/10508

kdwyer:


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -334,10 +345,32 @@ def confirm_orders(self):
             # allow serialisation error to propagate.
             self.env.cr.commit()
             self.invalidate_cache()
+        if exception_data:
+            # Raise an exception including details of the exceptions that have
+            # suppressed, in case we want to expose this information somewhere.
+            collected_exceptions = '\n\n'.join(exception_data)
+            _logger.error(collected_exceptions)
+            raise CombinedException('At least one error occurred while confirming orders.',
+                                    collected_exceptions=collected_exceptions) from None
```
nocturnalastro, line 354: from None?

nocturnalastro: NVM I found why

kdwyer: There's no benefit in chaining exceptions here, because this isn't a "real" exception.  At least, that's what I thought?

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -334,10 +345,32 @@ def confirm_orders(self):
...
 
         return True
 
 
+class CombinedException(ValidationError):
```
williamsajr, line 359: This might be better in a separate file, maybe `exceptions.py`

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -334,10 +345,32 @@ def confirm_orders(self):
             # allow serialisation error to propagate.
             self.env.cr.commit()
             self.invalidate_cache()
+        if exception_data:
+            # Raise an exception including details of the exceptions that have
+            # suppressed, in case we want to expose this information somewhere.
```
squah-u, line 350: Is this comment missing a word?

kdwyer: How about: 

> Raise an exception that includes details of the exceptions that have being suppressed, in case we want to expose this information somewhere else

?

squah-u: Sounds good. I thought a `been` had gotten lost in the original comment

kdwyer: Hah, good point.   I can never see my own mistakes.

Although in the above, s/being/been :smile:

PR #330, opened 2020-08-03
[IMP] udes_sale_stock: Add optional states argument to confirm orders

williamsajr:
Add ability to specify states to confirm to allow custom
states defined outside the module to be used.

User-story: 10609

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -299,11 +299,19 @@ class SaleOrder(models.Model):
         """ Override to disable tracking by default """
         return super(SaleOrder, self.with_context(tracking_disable=True)).action_confirm()
 
+    def _get_confirmation_domain(self):
+        """Returns the domain for sale orders to attempt confirmation."""
+        return [("state", "=", "draft")]
+
     def confirm_orders(self):
-        """Process sale and cancel, confirm or ignore sale orders according to
-           order processing configuration.
+        """Attempt to confirm sale orders.
+
+        This is done in batches to reduce the chance of concurrency errors
+        when confirming large numbers of orders at once. If one batch fails
+        the other batches may still be confirmed, attempting to maximise the number of
+        orders that may be confirmed.
         """
-        to_confirm = self.search([("state", "=", "draft")])
+        to_confirm = self.search(self._get_confirmation_domain())
         for _, batch in to_confirm.batched(size=1000):
             tries = 0
             while True:


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -299,11 +299,11 @@ def action_confirm(self):
         """ Override to disable tracking by default """
         return super(SaleOrder, self.with_context(tracking_disable=True)).action_confirm()
 
-    def confirm_orders(self):
+    def confirm_orders(self, states=["draft"]):
         """Process sale and cancel, confirm or ignore sale orders according to
            order processing configuration.
         """
-        to_confirm = self.search([("state", "=", "draft")])
+        to_confirm = self.search([("state", "in", states)])
```
apatrick88, line 306: Depending on the outcome of the other comment, is this still needed?

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -299,11 +299,11 @@ def action_confirm(self):
         """ Override to disable tracking by default """
         return super(SaleOrder, self.with_context(tracking_disable=True)).action_confirm()
 
-    def confirm_orders(self):
+    def confirm_orders(self, states=["draft"]):
```
PeterC10, line 302: Not sure if this is likely to cause an issue here but it can be dangerous using mutable objects as default arguments.

Might be safer to have states default as `None` and then at the start of the method do:

```
if states is None:
    states = ["draft"]
```

addons/udes_sale_stock/models/sale_order.py
ws-mps: The other option is to add a function that generates the domain to search orders, so it can be overriden.

williamsajr: Implemented Miquel's suggestion as it looks better than my optional argument

addons/udes_sale_stock/models/sale_order.py
williamsajr: No longer needed after the change to remove the optional argument

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -299,11 +299,15 @@ def action_confirm(self):
         """ Override to disable tracking by default """
         return super(SaleOrder, self.with_context(tracking_disable=True)).action_confirm()
 
+    def _get_confirmation_domain(self):
+        """Returns the domain for sale orders to attempt confirmation."""
+        return [("state", "=", "draft")]
+
     def confirm_orders(self):
         """Process sale and cancel, confirm or ignore sale orders according to
```
ws-mps, line 307: It is not from your current changes but this docstring is not very helpful.

williamsajr: Updated

PR #333, opened 2020-08-11
[FIX] udes_security: Credentials checked before resetting signup token

apatrick88:
Ensured that the strength and history of the password are checked prior to resetting the signup token. This means that any discrepancies are found earlier and will allow the user to subsequently provide a valid password successfully.

Story/1380

### DIFF
--- a/addons/udes_security/models/res_partner.py
+++ b/addons/udes_security/models/res_partner.py
@@ -1,0 +1,14 @@
+from odoo import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def write(self, vals):
+        password = self._context.get("password")
+        if password:
+            partner_user = self.user_ids and self.user_ids[0]
+
+            partner_user._check_password(password)
+        return super(ResPartner, self).write(vals)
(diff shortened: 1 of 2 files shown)


addons/udes_security/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def write(self, vals):
+        password = self._context.get("password")
+        if password:
+            partner_user = self.user_ids and self.user_ids[0] or False
```
nocturnalastro, line 11: I don't think you need the False is needed here.

apatrick88: Updated

addons/udes_security/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.multi
+    def write(self, vals):
+        password = self._context.get("password")
```
PeterC10, line 9: Why `self._context` instead of `self.env.context`?

apatrick88: `self._context` is just a shortcut for `self.env.context`, they do the same thing, I think `_context` looks a little cleaner.

PR #337, opened 2020-08-20
[ADD] udes_stock,udes_stock_move: Add Create Picking For Package action

squah-u:


addons/udes_stock_package_move/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "UDES Stock Package Move",
```
ws-mps, line 2: Can we call it UDES Stock Move ? At the moment we are moving packages but would be good to also move any quant.

ws-mps: Same for the name of the module.

ws-mps: Can we add a README, please?

addons/udes_stock_package_move/tests/test_change_package_location.py
```diff
@@ -0,0 +1,121 @@
...
+
+        cls.empty_package = Package.create({})
+        cls.package_1 = cls.create_package(cls.test_stock_location_01)
+        cls.package_2 = cls.create_package(cls.test_stock_location_01)
+        cls.package_3 = cls.create_package(cls.test_stock_location_01)
+        cls.package_4 = cls.create_package(cls.test_stock_location_02)
+        cls.packages = cls.package_1 | cls.package_2 | cls.package_3 | cls.package_4
+
+    @classmethod
+    def create_package(cls, location):
```
ws-mps, line 22: Note: `create_package` already exists in `common.BaseUDES`. Shall we rename this one ?

addons/udes_stock_package_move/tests/test_change_package_location.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Create a package containing apples and bananas at the given location
+        """
+        Package = cls.env["stock.quant.package"]
+
+        package = Package.create({})
```
ws-mps, line 28: Can use `cls.create_package()` so you don't need line 26 and 28.

addons/udes_stock_package_move/tests/test_change_package_location.py
```diff
@@ -0,0 +1,121 @@
...
+        cls.create_quant(cls.apple.id, location.id, 10, package_id=package.id)
+        cls.create_quant(cls.banana.id, location.id, 10, package_id=package.id)
+        return package
+
+    def complete_picking(self, picking):
```
ws-mps, line 33: Wondering if we want this one in `common.BaseUDES`, since this one will be needed for other modules.

What do you think about making the `action_done` optional, with a parameter `validate=true` by default ?

addons/udes_stock_package_move/tests/test_change_package_location.py
```diff
@@ -0,0 +1,121 @@
...
+        """ Marks a picking and all its moves as done """
+        for move_line in picking.move_line_ids:
+            move_line.qty_done = move_line.product_uom_qty
+        picking.action_done()
+
+    def test01_move_empty_package(self):
+        """ Test that moving an empty package fails """
+        ChangePackageLocation = self.env["change_package_location"]
```
ws-mps, line 41: We can store this in `self.ChangePackageLocation` at setupClass, since it is going to be used by all the tets.

addons/udes_stock_package_move/wizard/change_package_location.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from odoo import models, api, fields, _
+
+
+class ChangePackageLocation(models.TransientModel):
+    _name = "change_package_location"
```
ws-mps, line 6: Should contain the module name: `XXX.YYY`.

addons/udes_stock_package_move/tests/test_change_package_location.py
ws-mps: In fact, we can move it into a helper function since we are always doing:
```
ChangePackageLocation = self.env["change_package_location"]
wizard = ChangePackageLocation.create({})
wizard = wizard.with_context(active_ids=XXX)
```

addons/udes_stock_package_move/tests/test_change_package_location.py
```diff
@@ -0,0 +1,121 @@
...
+        """ Test moving a single package """
+        ChangePackageLocation = self.env["change_package_location"]
+        Picking = self.env["stock.picking"]
+
+        wizard = ChangePackageLocation.create({})
+        wizard = wizard.with_context(active_ids=self.package_2.ids)
+
+        wizard.location_dest_id = self.test_stock_location_02
+
+        picking = Picking.browse([wizard.create_picking()["res_id"]])
```
ws-mps, line 61: QQ: browse only accepts a list of IDs ? Before it was also accepting one ID.

addons/udes_stock_package_move/wizard/change_package_location.py
```diff
@@ -0,0 +1,45 @@
...
+        Users = self.env["res.users"]
+        warehouse = Users.get_user_warehouse()
+        return warehouse.int_type_id
+
+    def create_picking(self):
+        Package = self.env["stock.quant.package"]
+
+        self.ensure_one()
+
+        packages = Package.browse(self.env.context["active_ids"])
```
ws-mps, line 30: What happens if any of the quants of the packages is reserved? Probably we want a check it and raise an error before calling `create_picking`.

addons/udes_stock/models/stock_quant.py
```diff
@@ -53,6 +53,17 @@ def create_picking(self, picking_type, **kwargs):
                 - picking
         """
         Picking = self.env["stock.picking"]
+
+        params = dict(kwargs)
+
+        if "location_id" not in params:
+            locations = self.location_id
+            location = locations.get_common_ancestor()
+            if location:
```
ws-mps, line 62: What happens if we cannot find common ancestor?

addons/udes_stock/tests/test_stock_location.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from .common import BaseUDES
+
+
+class TestStockLocation(BaseUDES):
```
ws-mps, line 5: I would add another test with multiple levels: 
```
Stock > Zone 01 > Loc 01-01
Stock > Zone 01 > Loc 01-02
Stock > Zone 02 > Loc 02-01
Stock > Zone 02 > Loc 02-02
```
* Locations Loc 02-02 and Loc 01-01

addons/udes_stock/models/stock_picking.py
```diff
@@ -230,3 +230,16 @@ def _create_move(self, move_values):
         """
         Move = self.env["stock.move"]
         return Move.create(move_values)
+
+    def open_form_view(self):
```
ws-mps, line 234: Do we need this one here because it might be needed in other modules?

nocturnalastro (approved):

addons/udes_stock/models/stock_quant.py
```diff
@@ -53,6 +53,17 @@ def create_picking(self, picking_type, **kwargs):
                 - picking
         """
         Picking = self.env["stock.picking"]
+
+        params = dict(kwargs)
```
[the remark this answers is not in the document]
squah-u: I thought there was a review comment expressing a preference for taking a copy, but it looks like I was dreaming

Fixed

addons/udes_stock/tests/test_stock_quant.py
```diff
@@ -131,7 +131,24 @@ def test06_create_picking_from_single_quant_assign(self):
         # Check QuantA is now reserved
         self.assertEqual(self.quantA.reserved_quantity, 10)
 
-    def test07_create_picking_from_single_quant_priority(self):
+    def test07_create_picking_from_single_quant_assign_correct_quant(self):
```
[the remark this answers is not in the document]
squah-u: It's been split out. Be warned that github does not display commits in the right order

addons/udes_stock/models/stock_quant.py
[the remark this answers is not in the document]
ws-mps: That good to know.

addons/udes_stock_move/wizard/change_package_location.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def create_picking(self):
+        Package = self.env["stock.quant.package"]
+
+        self.ensure_one()
+
+        packages = Package.browse(self.env.context["active_ids"])
+
+        # Complain if there are any reserved quants in the packages
+        reserved_packages = packages.filtered(
```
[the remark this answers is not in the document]
squah-u: added stock.quant.package assert_not_reserved, keeping the name of the udes11 version but not the implementation

PR #341, opened 2020-09-02
[IMP] udes_sale_stock: Allow client_order_ref to be specified for copy

williamsajr:
User-story: 10608

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -76,8 +76,9 @@ class SaleOrder(models.Model):
 
     @api.multi
     def copy(self, default=None):
-        """Append '(copy)' to Customer Reference for duplicated Order"""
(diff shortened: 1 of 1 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -76,8 +76,9 @@ def init(self):
 
     @api.multi
     def copy(self, default=None):
-        """Append '(copy)' to Customer Reference for duplicated Order"""
```
nocturnalastro, line 79: can't we just set the copy flag on the field?

williamsajr: This code would always ignore any```client_order_ref ``` that had been specified in ```default```. The idea of the change is to prevent it from overwriting it when a value is passed to it but allow it to continue with the existing behaviour when nothing is passed

apatrick88 (approved):

nocturnalastro (approved):

PR #343, opened 2020-09-08
Delete button for Pickings removed in UI for non-trusted users

PeterC10:
Ported restriction from UDES 11. This restriction was for debug user in UDES 11, but for UDES 13 trusted/debug user are being merged.


apatrick88 (approved):

PeterC10:
Closed PR as we're going to be handling this in a different module.

PR #344, opened 2020-09-11
Story/9943 - Advanced Picking Information migration from UDES 11

PeterC10:


addons/udes_stock/models/stock_picking.py
```diff
@@ -10,6 +10,116 @@ class StockPicking(models.Model):
...
+    )
+    u_next_picking_ids = fields.One2many(
+        "stock.picking",
+        string="Next Pickings",
+        compute="_compute_related_picking_ids",
+        help="Next pickings",
+    )
+    u_created_backorder_ids = fields.One2many(
+        "stock.picking",
+        string="Created Back Orders",
```
squah-u, line 35: Do we want to change this to `Backorders` as well?

PeterC10: Not sure, both variants are used throughout UDES.

Will ask in the group and see what the consensus is. Can then update the other instances of Backorders/Back Orders so that they're consistent. :)

addons/udes_stock/models/stock_picking.py
```diff
@@ -10,6 +10,116 @@ class StockPicking(models.Model):
...
+
+            picking.u_quantity_done = total_qty_done
+            picking.u_total_quantity = total_qty_todo
+            picking.u_has_discrepancies = has_discrepancies
+
+    @api.depends("move_line_ids", "move_line_ids.result_package_id")
+    def _compute_num_pallets(self):
+        """ Compute the number of pallets from the operations """
+        for picking in self:
+            picking.u_num_pallets = len(picking.move_line_ids.result_package_id)
```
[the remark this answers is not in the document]
PeterC10: In Odoo 13 you only need to use mapped for non-relational fields. For example the output below shows the same output with/without mapped:

`x = self.test_stock_location_01`
`y = self.test_goodsout_location_01`

`(x + y).mapped("write_uid")`
`res.users(1,)`
`(x + y).write_uid`
`res.users(1,)`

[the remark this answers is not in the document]
PeterC10: And in the UI:

image
image

apatrick88 (approved):

addons/udes_stock/models/stock_picking.py
PeterC10: Have now replaced instances of 'Back order' with 'Backorder'.

davidmcgaw (approved):
Tests LGTM!

addons/udes_stock/models/stock_picking.py
```diff
@@ -10,6 +10,116 @@ class StockPicking(models.Model):
...
+        "move_lines.move_orig_ids.picking_id",
+        "move_lines.move_dest_ids.picking_id",
+    )
+    def _compute_related_picking_ids(self):
+        """ Compute previous/next picking """
+        for picking in self:
+            picking.u_prev_picking_ids = picking.move_lines.move_orig_ids.picking_id
+            picking.u_next_picking_ids = picking.move_lines.move_dest_ids.picking_id
+
+            picking.u_created_backorder_ids = self.search([("backorder_id", "=", picking.id)])
```
ws-mps, line 93: In 11 there is check for picking.id , is it not needed?

PeterC10: No, because computed fields are only run on saved records, so the record will always have an ID set.

joe-gb (approved):

PR #345, opened 2020-09-14
[IMP] udes_stock: Removed f-strings from translated strings in Stock Model Mixin

PeterC10:
Removed f-strings in get_or_create method of MixinStockModel as they do not play nicely with Odoo translations.

Story/10305

### DIFF
--- a/addons/udes_stock/models/mixin_stock_model.py
+++ b/addons/udes_stock/models/mixin_stock_model.py
@@ -55,15 +55,17 @@ class MixinStockModel(models.AbstractModel):
                 results = self.create({"name": identifier})
             elif self.MSM_CREATE and create:
                 raise ValidationError(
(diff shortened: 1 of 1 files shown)


squah-u:
`model_name` may have been intentionally left inside the `_()` so that we can translate the model name. 

I remember flagging the string interpolations within `_()`s in the original PR and some but not all of them got moved out in the end. I can't remember if we had a discussion about it or what the outcome was. @ws-mps might know

PeterC10:
> `model_name` may have been intentionally left inside the `_()` so that we can translate the model name.
> 
> I remember flagging the string interpolations within `_()`s in the original PR and some but not all of them got moved out in the end. I can't remember if we had a discussion about it or what the outcome was. @ws-mps might know

Yep, just come off a call with @ws-mps and we don't need to change this. Closing PR. :)

PR #360, opened 2020-10-12
[IMP] udes_stock: Improve descrepancy highlighting of stock moves

joe-gb:
On the stock picking screen, once a move has been complete (or has
partial qty assigned) it should be clearly shown with colour coding
to highlight descrepancies. (story/11411)

### DIFF
--- a/addons/udes_stock/views/stock_picking.xml
+++ b/addons/udes_stock/views/stock_picking.xml
@@ -109,10 +109,10 @@
 
             <!-- Change colour of the rows depending on new values -->
             <xpath expr="//field[@name='move_lines']/tree" position="attributes">
-                <attribute name="decoration-danger">ordered_qty &lt; quantity_done</attribute>
+                <attribute name="decoration-danger">(ordered_qty != quantity_done) and quantity_done > 0</attribute>
                 <attribute name="decoration-success">ordered_qty == quantity_done</attribute>
                 <attribute name="decoration-muted">scrapped == True or state == 'cancel'</attribute>
-                <attribute name="decoration-bf">(ordered_qty == quantity_done) or (ordered_qty &lt; quantity_done)</attribute>
+                <attribute name="decoration-bf">quantity_done > 0</attribute>
             </xpath>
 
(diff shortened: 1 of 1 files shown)


kenvac (approved):
LGTM

addons/udes_stock/views/stock_picking.xml
```diff
@@ -109,10 +109,10 @@
 
             <!-- Change colour of the rows depending on new values -->
             <xpath expr="//field[@name='move_lines']/tree" position="attributes">
-                <attribute name="decoration-danger">ordered_qty &lt; quantity_done</attribute>
+                <attribute name="decoration-danger">ordered_qty &lt; quantity_done or ((ordered_qty > quantity_done) and (state == 'done'))</attribute>
```
ws-mps, line 112: I don't see why it is different `ordered_qty &lt; quantity_done` should also cover `((ordered_qty > quantity_done) and (state == 'done'))`. The only difference is `&lt;` vs `>`. Have we tried changing that? or do we need both?

ws-mps, line 112: Do we want to consolidate `&lt;` vs `>` ?

addons/udes_stock/views/stock_picking.xml
```diff
@@ -109,10 +109,10 @@
...
                 <attribute name="decoration-success">ordered_qty == quantity_done</attribute>
                 <attribute name="decoration-muted">scrapped == True or state == 'cancel'</attribute>
-                <attribute name="decoration-bf">(ordered_qty == quantity_done) or (ordered_qty &lt; quantity_done)</attribute>
+                <attribute name="decoration-bf">(ordered_qty == quantity_done) or (ordered_qty &lt; quantity_done) or ((ordered_qty > quantity_done) and (state == 'done'))</attribute>
```
nocturnalastro, line 115: do we want the same for both?

nocturnalastro: Ah wait nvm

ws-mps (approved):

unirho (approved):

PR #362, opened 2020-10-14
Story/11414 list goods in

nocturnalastro:


unirho (approved):

addons/udes_stock/controllers/stock_picking.py
```diff
@@ -111,3 +111,26 @@ def warn_picking_precondition(self, ident):
 
         picking = Picking.browse(ident)
         return picking.action_warn_picking_precondition()
+
+    @http.route("/api/stock-picking/list", type="json", methods=["GET"], auth="user")
+    def list_ready_pickings(
```
ws-mps, line 116: Is this endpoint needed? Can't we get the same result with get_pickings ?

ws-mps (approved):

PR #363, opened 2020-10-14
Story/11461 - Delivery Control picking type functionality

PeterC10:
Allows for Delivery Control pickings to be completed and cancelled without the need for any moves.

Goods In picking generated when Delivery Control picking completed (if it doesn't have one linked already). Cancelling a Delivery Control picking will also cancel the linked Goods In, and Goods In cannot be completed if the linked Delivery Control picking is in a pending state.

Computed first/previous/next picking fields overridden to take into account Delivery Control pickings.

Unit tests added and udes_delivery_control module added to udes_open Dockerfile.

~~Note: There are some TODOs in the code relating to required fields which are part of story 11446. These will be addressed when that story has been merged in.~~

### DIFF
--- a/addons/udes_delivery_control/models/stock_picking.py
+++ b/addons/udes_delivery_control/models/stock_picking.py
@@ -73,6 +73,17 @@ class StockPicking(models.Model):
     u_backload_end_date = fields.Datetime(related="u_extras_id.backload_end_date")
     u_backload_time_taken = fields.Float(related="u_extras_id.backload_time_taken")
 
+    u_delivery_control_picking_id = fields.Many2one(
+        "stock.picking",
+        string="Originating Delivery Control Picking",
+        readonly=True,
+        index=True,
+        copy=False,
+    )
+    u_goods_in_picking_id = fields.Many2one(
+        "stock.picking", string="Generated Goods In Picking", readonly=True, index=True, copy=False
+    )
+
     _sql_constraints = [
         (
             "extras_id_uniq",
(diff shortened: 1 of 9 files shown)


addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -0,0 +1,204 @@
...
+        for picking in self.filtered(lambda p: p.picking_type_id == delivery_control_type):
+            picking.write({"state": "done", "date_done": fields.Datetime.now()})
+
+            if not picking.u_goods_in_picking_id:
+                self.create_goods_in_from_delivery_control()
+            elif picking.u_goods_in_picking_id.state == "waiting":
+                picking.u_goods_in_picking_id.action_confirm()
+
+        return True
+
```
squah-u, line 150: Do we need an `action_done` override somewhere?

PeterC10: Maybe. I'm doing the call to `action_delivery_control_done` inside of `button_validate`, but I suppose it is safer/more logical to have an `action_done` override too. Will adjust. :)

PeterC10: I've added in an override for `action_done`. It will simply call `action_delivery_control_done` for Delivery Control pickings, and work as normal for other types.

I needed to keep in the override for `button_validate` as this checks Stock Moves, but I am now calling `action_done` in the unit tests rather than `action_delivery_control_done` directly.

addons/udes_delivery_control/tests/test_delivery_control_workflow.py
```diff
@@ -0,0 +1,223 @@
...
+        self.delivery_control_picking.button_validate()
+        self._assert_picking_state(self.delivery_control_picking, "done")
+
+    def test_assert_goods_in_generated_when_delivery_control_complete(self):
+        """
+        Assert that a Goods In picking is generated when a Delivery Control picking is completed,
+        and is correctly setup
+        """
+        # Assert that Delivery Control picking doesn't already have a Goods In picking linked
+        self.assertFalse(
```
davidmcgaw: IMHO you don't need to do an assert before your test

addons/udes_delivery_control/tests/test_related_pickings.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Set Pick C's move to have originated from Goods In and Pick A's move
+        pick_c_move.move_orig_ids = goods_in_move | pick_a_move
+
+        # Pick Relationship Diagram
+        #
+        # DC         A
+        # |          |
+        # GI____     |
+        # |     |    |
```
davidmcgaw, line 108: nice ascii art, but maybe could use \ and / above C.
It looks like GI comes from the top but A from the bottom

PeterC10: Art in general has never been my strong point!

But updated now with your suggestion. :)

davidmcgaw (approved):
LGTM!

PeterC10:
Unit test `test05_get_info_all` is failing from the `udes_stock` module, because as part of this story Goods In picking type was changed to not allow partial pickings (to prevent Goods In being completed while the Delivery Control is pending). This meant the `get_info` method on `stock.picking` was returning `u_pending` - which the unit test was not expecting for Goods In.

There are two potential solutins:

1. Update the unit test so it checks whether the picking type allows partial pickings and adjust the expected outcome depending on that
2. Undo the change setting Goods In to now allow partial pickings, and put in a manual check to see if the Delivery Control picking is complete

Change No. 2 has been added in commits marked: `To Be Squashed/Dropped`. This is to be reviewed.

PeterC10:
After speaking with @ws-mps I have now implemented a slightly different version of change no. 1 from https://github.com/unipartdigital/udes-open/pull/363#issuecomment-709353427

Goods In has been set to not handle partial pickings again. `TestGoodsInPicking` in `udes_stock` has been updated so that `u_handle_partials` for Goods in is explicitly set to `True` for the tests, which means it will pass CI with the `udes_delivery_control` enabled.

PeterC10:
Have now rebased on latest master version and made use of the `u_is_delivery_control` field which was added in story 11446.

squah-u (approved):

apatrick88 (approved):

PR #374, opened 2020-10-27
story/11448 - udes_delivery_control

kenvac:
Two new endpoints get and confirm picking is added with this commit
Those endpoints will be consumed by mobile client

User-story: 11448

Signed-off-by: Kinner Vachhani

### DIFF
--- a/addons/udes_delivery_control/controllers/main.py
+++ b/addons/udes_delivery_control/controllers/main.py
@@ -1,0 +1,66 @@
+from odoo import http, _
+from odoo.http import request
+from odoo.addons.udes_stock.controllers.main import UdesApi
+from odoo.exceptions import ValidationError
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+class PickingApi(UdesApi):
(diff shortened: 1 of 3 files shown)


williamsajr:
If we're going to make stock mobile depend on this module we may want to disable the delivery control picking type by default and enable manually for customers so that they don't all see it

addons/udes_delivery_control/controllers/main.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @http.route(
+        "/api/delivery-control/stock-picking/<int:ident>",
+        type="json",
+        methods=["POST"],
+        auth="user",
+    )
+    def confirm_picking(self, ident, **kwargs):
+        """ Update and confirm picking
+        """
```
williamsajr, line 47: Style nit: For docstring that can fit on one line we usually have the closing `"""` on the same line

williamsajr (approved):

apatrick88 (approved):

addons/udes_delivery_control/controllers/main.py
kenvac: done

PR #375, opened 2020-10-28
 User-story: 11547

ws-mps:
Extend stock.picking._backorder_movelines()

Added optional parameter `picking` so that move lines can be moved
into an existing picking instead of creating a new one.

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -907,41 +907,44 @@ class StockPicking(models.Model):
                 return True
         return False
 
-    def _backorder_movelines(self, mls=None):
-        """ Creates a backorder pick from self (expects a singleton)
+    def _backorder_movelines(self, mls=None, dest_picking=None):
+        """ Creates a backorder picking from self (expects a singleton)
             and a subset of stock.move.lines are then moved into it.
-            The move from which the move lines have been transferred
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -907,41 +907,44 @@ def _requires_backorder(self, mls):
                 return True
         return False
 
-    def _backorder_movelines(self, mls=None):
-        """ Creates a backorder pick from self (expects a singleton)
+    def _backorder_movelines(self, mls=None, picking=None):
```
squah-u, line 910: nit: Can we name it `dest_picking` or something that's not `picking`, to distinguish it from the `self` picking?

squah-u (approved):

nocturnalastro (approved):

addons/udes_stock/models/stock_picking.py
ws-mps: Done, please review.

squah-u: The final `return` needs updating

ws-mps: Changed too :)

squah-u: lgtm

williamsajr (approved):

apatrick88 (approved):

PR #379, opened 2020-11-04
story/11474:  adds endpoint got get products classification message

kenvac:


addons/udes_warehouse_classification/models/stock_picking.py
```diff
@@ -14,7 +14,8 @@ def _get_classification_messages_for_product_picking(self):
                 lambda c: self.picking_type_id in c.picking_type_ids
             ).sorted("sequence")
             product_classifications[product.barcode] = [
-                {"message": classification.alert_message} for classification in classifications
+                {"message": classification.alert_message, "name": classification.name}
```
apatrick88, line 17: I think we just need the alert message and not the name of the classification. So we display the `alert_message` in the mobile UI.

apatrick88 (changes requested):

kenvac:
These changes are not longer required as only message is used for for display. 
Closing this PR

PR #381, opened 2020-11-05
Story/11788 - Add Package Type support

PeterC10:
Package Type now displayed on Packages.

Pallet and Stillage Package Types added.

API endpoint added to retrieve Package Types.

Parent package type can now be set via `update_picking`.

### DIFF
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -379,6 +379,20 @@ class StockPickingType(models.Model):
         ),
     )
 
+    u_check_package_type = fields.Boolean(
+        string="Check Package Type",
+        default=False,
+        help="If set, the user may be asked to confirm the type of the package they are scanning"
+        " during the picking workflow",
+    )
+    
+    u_default_package_type_id = fields.Many2one(
+        comodel_name="product.packaging",
+        string="Default Package Type (For Check)",
+        help="Default value for when user is asked to confirm type of package they are scanning"
+        " (this only applies if 'Check Package Type' is enabled)",
+    )
+
     def do_refactor_action(self, action, moves):
         """Resolve and call the method to be executed on the moves.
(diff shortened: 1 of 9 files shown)


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -443,6 +457,8 @@ def _prepare_info(self):
             - u_reserve_pallet_per_picking: boolean
             - u_max_reservable_pallets: int
             - u_selection_method: string
+            - u_check_package_type: boolean
+            - u_default_package_type_id: id
```
squah-u, line 461: nit: should this say `int` like `default_location_dest_id`? I prefer `id` as it's more informative but for consistency `int` wins

addons/udes_stock/views/stock_picking_type.xml
```diff
@@ -60,6 +60,8 @@
                     <field name="u_reserve_pallet_per_picking" />
                     <field name="u_max_reservable_pallets" />
                     <field name="u_selection_method" />
+                    <field name="u_check_package_type" />
+                    <field name="u_default_package_type_id" options="{'no_create': True}"/>
```
squah-u, line 64: formatting nit: space before the `/>`

squah-u (approved):

addons/udes_stock/models/stock_picking_type.py
PeterC10: Yep, that wasn't intentional! Good spot. :)

PeterC10: Fixed. :)

addons/udes_stock/views/stock_picking_type.xml
PeterC10: Fixed. :)

apatrick88 (approved):

PR #384, opened 2020-11-10
[ADD] udes_stock: Add field to disable the `No backorder` button

peterpratt13:
Add `u_disable_no_backorder_button_picking_type_ids` to stock.warehouse, and
`u_disable_no_backorder_button` to stock.backorder.confirmation. This allows
easier configuration of the picking types that must always create a backorder
when validating in the UI when not everything is available/done.

User-story: 11495

### DIFF
--- a/addons/udes_stock/models/stock_warehouse.py
+++ b/addons/udes_stock/models/stock_warehouse.py
@@ -19,6 +19,15 @@ class StockWarehouse(models.Model):
 
         return [('id', 'not in', stock_children_locations)]
 
+    def _domain_warehouse_picking_types(self):
+        """
+        Domain for getting picking types in the current warehouse
+        """
+        User = self.env["res.users"]
+
+        warehouse = User.get_user_warehouse()
+        return [("warehouse_id", "=", warehouse.id)]
+
     u_handle_damages_picking_type_ids = fields.Many2many(
         comodel_name='stock.picking.type',
         relation='stock_warehouse_damages_picking_types_rel',
@@ -101,6 +110,15 @@ class StockWarehouse(models.Model):
         help="Allow users to inventory adjust reserved stock."
     )
 
+    u_disable_no_backorder_button_picking_type_ids = fields.Many2many(
+        comodel_name="stock.picking.type",
+        relation="stock_warehouse_disable_no_backorder_button_rel",
+        string="Which Picking Types to hide the `No Backorder`",
+        help="Which Picking Types have to always create backorders on validate when the pickings "
+             "are not fully available, so hide the `No backorder` button.",
+        domain=_domain_warehouse_picking_types
+    )
+
     @lazy_property
     def reserved_package_name(self):
         return list(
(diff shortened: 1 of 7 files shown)


addons/udes_stock/models/stock_warehouse.py
```diff
@@ -101,6 +101,14 @@ def _domain_u_damaged_location_id(self):
         help="Allow users to inventory adjust reserved stock."
     )
 
+    u_disable_no_backorder_button_picking_type_ids = fields.Many2many(
```
williamsajr, line 104: It may be worth adding a domain so you can only add picking types which are associated with the current warehouse

williamsajr: Could you also update the readme for the new field

williamsajr (approved):

addons/udes_stock/wizard/stock_backorder_confirmation.py
```diff
@@ -0,0 +1,22 @@
...
+        "validating a picking not fully available for that picking type.",
+    )
+
+    @api.depends("pick_ids")
+    def _compute_disable_no_backorder_button(self):
+        """Compute if the picking type must always create a backorder"""
+        Users = self.env["res.users"]
+        warehouse = Users.get_user_warehouse()
+        self.u_disable_no_backorder_button = (
+            self.pick_ids.picking_type_id
```
ws-mps, line 20: This is assuming there is only one picking in `pick_ids`. Would be better to use `mapped`. We could even ensure we only have one picking type:
```
picking_type = self.mapped('pick_ids.picking_type_id')
picking_type.ensure_one()
```

peterpratt13: Fixed. Out of curiosity, how do you create a backorder for multiple pickings at once, I've only ever done it per picking in the UI.

ws-mps: I don't know that, just raised it because it is is `pick_ids`.

peterpratt13: I'll leave it in though just so if they do find a way they can't break everything.

addons/udes_stock/wizard/stock_backorder_confirmation.xml
```diff
@@ -0,0 +1,24 @@
...
+        <!-- Change backorder confirmation -->
+        <record id="view_backorder_confirmation" model="ir.ui.view">
+            <field name="name">udes_stock.stock.backorder.confirmation</field>
+            <field name="model">stock.backorder.confirmation</field>
+            <field name="inherit_id" ref="stock.view_backorder_confirmation"/>
+            <field name="arch" type="xml">
+                <xpath expr="//form" position="inside">
+                    <field name="u_disable_no_backorder_button" invisible="1"/>
+                </xpath>
+                <!-- Hide the description as it is no longer ambiguos if teh `No backorder` button is disabled -->
```
unirho: teh

unirho: ambiguos should be ambiguous

unirho (approved):

PR #387, opened 2020-11-20
User-story: 12091

ws-mps:
Add recompute to split_out_move_lines()

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -195,15 +195,16 @@ class StockMove(models.Model):
                 }
             )
 
+            if self.move_orig_ids:
+                (bk_move | self).update_orig_ids(self.move_orig_ids)
+
             # When not complete, splitting a move may change its state,
             # so recompute
             incomplete_moves = (self | bk_move).filtered(lambda mv: mv.state != "done")
+            incomplete_moves.recompute()
             incomplete_moves._recompute_state()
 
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -195,16 +195,17 @@ def split_out_move_lines(self, move_lines):
...
+            if self.move_orig_ids:
+                (bk_move | self).update_orig_ids(self.move_orig_ids)
+
             # When not complete, splitting a move may change its state,
             # so recompute
             incomplete_moves = (self | bk_move).filtered(lambda mv: mv.state != "done")
+            incomplete_moves.recompute()
             incomplete_moves._recompute_state()
 
             move_lines.write({"state": bk_move.state})
```
nocturnalastro, line 207: Does this also need moving?

nocturnalastro (approved):

addons/udes_stock/models/stock_move.py
ws-mps: I don't think so, we first recompute the state of the stock.move and then update its move lines.

nocturnalastro: Is the state of the moves dependent on the move lines?

ws-mps: Yes, but not on their state.

apatrick88 (approved):

joe-gb (approved):

PR #390, opened 2020-11-25
Story 11973

kenvac:
- New button on the inventory dashboard shows the number of packages at a location for a picking type
- Existing "Transfers" button can be shown/ hidden based upon config for the picking type. This should be shown for all picking types

### DIFF
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -395,6 +395,36 @@ class StockPickingType(models.Model):
         " (this only applies if 'Check Package Type' is enabled)",
     )
 
+    # Package count options
+    u_show_package_count = fields.Boolean(
+        string="Show Package Count",
+        default=False,
+        help="Enables Package count button on kanban view",
+    )
+    u_package_count_button_name = fields.Char(
+        translate=True,
+        default=_("Package Count"),
+        string="Pack Count Text",
+        help="Custom text for Pack button on picking type tiles",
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -395,6 +395,42 @@ class StockPickingType(models.Model):
         " (this only applies if 'Check Package Type' is enabled)",
     )
 
+    # Package count options
+    u_show_package_count = fields.Boolean(
+        string="Show Package Count",
+        default=False,
+        help="Enables Package count button on kanban view",
+    )
+    u_package_count_location_id = fields.Many2one(
```
ws-mps, line 404: Have we considered using the default source location of the picking type instead of adding a new field?

apatrick88: I think we probably could for this use case, although might we want to option to specify a different location in the future?

apatrick88 (approved):

addons/udes_stock/models/stock_picking_type.py
ws-mps: In that case, we probably want multiple locations, since a picking type can be used from different source locations.
I would keep it simple for now, and iterate later.

kenvac: Thank you guys for your review and suggestions. 
I have remove the pack_count_location_id field and now using default_location_src_id

PR #392, opened 2020-11-25
[FIX] udes_priority: Fixes for priority propagation

joe-gb:
User-story: 11723

### DIFF
--- a/addons/udes_priorities/models/stock_move.py
+++ b/addons/udes_priorities/models/stock_move.py
@@ -1,4 +1,7 @@
 from odoo import fields, models, api, _
+import logging
+
+_logger = logging.getLogger(__name__)
 
 
 class StockMove(models.Model):
@@ -8,14 +11,15 @@ class StockMove(models.Model):
     def _default_priority(self):
         Priorities = self.env["udes_priorities.priority"]
         priorities = Priorities.search(self._priority_domain(), limit=1)
-        return priorities.reference if priorities else None
+        return (
+            priorities.reference if priorities else self.env.ref("udes_priorities.normal").reference
+        )
 
     priority = fields.Selection(selection="get_priorities_for_selection", default=_default_priority)
 
     def _priority_domain(self):
         domain = []
         picking_type_ids = []
-
         picking_types = self.mapped("picking_type_id")
         default_picking_type_id = self.env.context.get("default_picking_type_id", None)
 
@@ -39,4 +43,32 @@ class StockMove(models.Model):
     def get_priorities_for_selection(self):
         Priorities = self.env["udes_priorities.priority"]
         priorities = Priorities.search(self._priority_domain())
+
+        # hard coded default value means there is always a priority to set
+        normal = self.env.ref("udes_priorities.normal")
+        priorities |= normal
+
+        for move in self.filtered(lambda m: m.priority not in priorities.mapped("reference")):
+            # Theres some race conditions around where data is aviable to search on this means
+            # that sometimes an invalid default can be set
+            move.priority = normal.reference
+
         return priorities.get_selection_values()
+
+    @api.constrains("priority")
+    @api.onchange("priority")
+    def _priority_cant_be_empty(self):
+        for move in self:
+            if not move.priority:
+                move.priority = self.env.ref("udes_priorities.normal").reference
+
+    @api.model
+    def create(self, values):
+        context = {}
+        picking_type_id = values.get("picking_type_id", None)
+        if picking_type_id:
+            context = {"default_picking_type_id": picking_type_id}
+        res = super(
+            StockMove, self.with_context(**context),
+        ).create(values)
+        return res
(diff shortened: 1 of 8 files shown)


addons/udes_priorities/models/stock_move.py
```diff
@@ -39,4 +43,33 @@ def _priority_domain(self):
     def get_priorities_for_selection(self):
         Priorities = self.env["udes_priorities.priority"]
         priorities = Priorities.search(self._priority_domain())
+
+        # hard coded default value means there is always a priority to set
+        normal = self.env.ref("udes_priorities.normal")
+        priorities |= normal
```
ws-mps, line 49: Do we want to do this only if `priorities` is empty?

addons/udes_priorities/models/stock_move.py
```diff
@@ -39,4 +43,33 @@ def _priority_domain(self):
...
+
+        for move in self.filtered(lambda m: m.priority not in priorities.mapped("reference")):
+            # Theres some race conditions around where data is aviable to search on this means
```
ws-mps, line 52: Typo: aviable

addons/udes_priorities/models/stock_picking.py
```diff
@@ -37,31 +40,41 @@ def _priority_and_priority_group_domain(self):
...
             active_pick = self.browse(active_id).exists()
             if active_pick:
                 pick = active_pick
 
-        return Priorities.search(pick._priority_domain()).get_selection_values()
+        priorities = Priorities.search(pick._priority_domain())
+
+        # hard coded default value means there is always a priority to set
+        normal = self.env.ref("udes_priorities.normal")
+        priorities |= normal
```
ws-mps, line 67: Same as above.

addons/udes_priorities/models/stock_picking.py
```diff
@@ -81,13 +94,23 @@ def get_priorities(self):
...
-        for group in PriorityGroup.search(self._priority_group_domain()):
-            groups.append(
-                {
-                    "name": group.name,
-                    "priorities": [{"id": p.reference, "name": p.name} for p in group.priority_ids],
-                }
-            )
+        group_domain_kwargs = {}
+        if picking_type_id is None:
+            group_domain_kwargs["picking_type_ids"] = [picking_type_id]
```
ws-mps, line 99: `picking_type_id` is None here, so why don't just []?

unirho: should probably be not none but mick wasn't sure why he had put this

addons/udes_priorities/models/stock_picking.py
unirho: as we can default to normal we always want to provide that back as a choice, we could run into some nasties where move has defaulted to normal (say we make one without pick type) and the picking we try to give it to doesn't have normal as an acceptable value. This is mostly relevant when you try to set values to nothing, there has to be a supreme default, along the lines of why vanilla odoo always puts 1

addons/udes_priorities/tests/test_stock_picking.py
```diff
@@ -23,3 +23,31 @@ def test_correct_picking_type_or_no_picking_type_is_valid(self):
...
+        Move = self.env["stock.move"]
+        apple = Product.create(
+            {"name": "Apple", "barcode": "bar-apple", "default_code": "APL", "type": "product",}
+        )
+        self.not_urgent.picking_type_ids = self.picking_type_pick
+        self.assertEqual(self.not_urgent.picking_type_ids, self.picking_type_pick)
+        pick = Picking.create({"priority": self.not_urgent.reference})
+        self.assertEqual(pick.priority, self.not_urgent.reference)
+        move = Move.create(
+            {
```
unirho, line 41: as you see here we provide nothing for the priority for the move and so it is defaulted to normal, which means we must include that in the selection domain defined by the method "get_priorities_for_selection"

addons/udes_priorities/models/stock_picking.py
unirho: i will update this in the hht story so we can test it as that is what this function is intended for, as you say it doesn't block this

addons/udes_priorities/models/stock_move.py
unirho: resolving as per comments

unirho (approved):

addons/udes_priorities/models/stock_move.py
```diff
@@ -39,4 +43,33 @@ def _priority_domain(self):
...
+            # Theres some race conditions around where data is available to search on this means
+            # that sometimes an invalid default can be set
+            move.priority = normal.reference
+
         return priorities.get_selection_values()
+
+    @api.constrains("priority")
+    @api.depends("priority")
+    @api.onchange("priority")
+    def _priority_cant_be_empty(self):
```
apatrick88, line 61: Strictly speaking, do we need all of constrains, depends and onchange here? Would we get what we need with just constrains?

joe-gb: constrains and onchange are needed, I think depends is the one we can remove... but not 100%

joe-gb: removed :)

apatrick88 (approved):

PR #394, opened 2020-11-27
[11929] Delivery Control Changes

kenvac:


addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -13,11 +13,32 @@ class StockPickingExtras(models.Model):
     picking_ids = fields.One2many(
         comodel_name="stock.picking", inverse_name="u_extras_id", string="Pickings"
     )
-    # Unloading Fields
+    loading_type = fields.Selection(
+        [("unloading", "Unloading"), ("backloading", "Backloading"), ("both", "Both")],
```
PeterC10, line 17: Nit: `both` seems a bit wrong to use here for the code value (although in the context of the full list it makes perfect sense). Might be worth getting some opinions on this. I suppose we would want to avoid something long like `unloading_and_backloading`.

addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -114,6 +119,22 @@ def _compute_backload_time_taken(self):
                     record.backload_start_date, record.backload_end_date
                 )
 
+    @api.depends("loading_type")
+    # @api.onchange("u_loading_type")
```
PeterC10, line 123: Nit: Commentated out `onchange` can be removed.

addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -114,6 +119,22 @@ def _compute_backload_time_taken(self):
...
+    def _compute_loading_type(self):
+        """Compute is_backloading and is_unloading from loading_type"""
+        # TODO improve the code
```
PeterC10, line 126: Is this `TODO` still needed?

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -39,38 +39,43 @@ class StockPicking(models.Model):
...
     u_pallet_count = fields.Integer(related="u_extras_id.pallet_count")
     u_stillage_count = fields.Integer(related="u_extras_id.stillage_count")
     u_box_count = fields.Integer(related="u_extras_id.box_count")
-    u_is_planned = fields.Boolean(related="u_extras_id.is_planned",)
-    u_is_late = fields.Boolean(related="u_extras_id.is_late",)
-    u_is_fast_track = fields.Boolean(related="u_extras_id.is_fast_track",)
 
     # Backloading Fields
-    u_is_backload = fields.Boolean(related="u_extras_id.is_backload",)
+    u_is_backload = fields.Boolean(related="u_extras_id.is_backload")
```
PeterC10, line 68: Do we need `u_is_backload` and `u_is_unload`? Can't we just check `loading_type` for any checks?

PeterC10: I suppose it makes it easier to just check a boolean value rather than having to check if a selection value is in a list of values.

addons/udes_delivery_control/models/picking_extra_fields.py
apatrick88: Let's use `unload`, `backload` and `unload_backload`.

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -166,6 +171,21 @@ def _compute_show_mark_as_todo(self):
         for picking in self.filtered("u_is_delivery_control"):
             picking.show_mark_as_todo = picking.state == "draft"
 
+    @api.onchange("u_loading_type")
+    def _compute_loading_type(self):
```
PeterC10, line 175: Why is this method in both `stock.picking` and `stock.picking.extras`? I might be missing something but in the picking can't we just take the value from the extras record?

addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -114,6 +119,22 @@ def _compute_backload_time_taken(self):
...
+        for record in self:
```
PeterC10, line 127: Using Adam's values for `loading_type` we could simplify this to:
```
for record in self:
    record.is_unload = "unload" in record.loading_type
    record.is_backload = "backload" in record.loading_type
```

addons/udes_delivery_control/models/picking_extra_fields.py
kenvac: done!

addons/udes_delivery_control/models/picking_extra_fields.py
kenvac: done!

addons/udes_delivery_control/models/picking_extra_fields.py
kenvac: Removed

addons/udes_delivery_control/models/picking_extra_fields.py
kenvac: done!

addons/udes_delivery_control/models/stock_picking.py
kenvac: The onchange on loading_type field will hide/unhide fields on delivery control picking type. We can't fetch value from `stock.picking.extras` because at this point won't have been updated on the related field ( the write back doesn't happened until we save the record)

addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -27,44 +52,28 @@ class StockPickingExtras(models.Model):
...
-    )
 
     # Back Loading Fields
-    is_backload = fields.Boolean(
-        string="Backload Required ?",
-        help="Indicate whether a backload is required. \
-            (e.g. to take back cages used for past deliveries)",
-        default=False,
-    )
+    is_backload = fields.Boolean(compute="_compute_loading_type")
```
ws-mps, line 61: Do we want readonly?

PeterC10: Aren't computed fields readonly by default?

addons/udes_delivery_control/models/stock_picking.py
PeterC10: In that case could we refactor out the logic into a `get_loading_type` method and call this from both compute methods? Not sure if that's practical, but would be good to avoid duplicated code if possible. Will have a think and get back to you.

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -166,6 +252,13 @@ def _compute_show_mark_as_todo(self):
...
+        """Compute is_backloading and is_unloading from u_loading_type"""
+        for record in self:
+            record.is_unload = "unload" in record.loading_type
```
PeterC10, line 259: Shouldn't this be `u_loading_type`? If so I will resolve my comment about refactoring the logic as they are different field names.

kenvac: Errr! I made a mistake. Good eyes mate :+1: 
Changed

addons/udes_delivery_control/models/stock_picking.py
PeterC10: Resolving as no longer relevant (both entities use different field names).

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -267,7 +361,7 @@ def action_delivery_control_done(self):
         for picking in self.filtered("u_is_delivery_control"):
             picking.write({"state": "done", "date_done": fields.Datetime.now()})
 
-            if not picking.u_goods_in_picking_id:
+            if not picking.u_goods_in_picking_id and (picking.u_loading_type != "backload"):
```
PeterC10, line 364: Nit: Don't need brackets.

PeterC10 (approved):
LGTM

addons/udes_delivery_control/models/stock_picking.py
PeterC10: Maybe`picking.u_is_unload` would be neater here too?

kenvac: done

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -267,7 +360,7 @@ def action_delivery_control_done(self):
         for picking in self.filtered("u_is_delivery_control"):
             picking.write({"state": "done", "date_done": fields.Datetime.now()})
 
-            if not picking.u_goods_in_picking_id:
+            if not picking.u_goods_in_picking_id and picking.u_is_backload:
```
PeterC10, line 363: This needs to be `picking.u_is_unload` not `picking.u_is_backload`.

We only want to generate a Goods In if the Delivery Control is unload or both unload and backload. If it is backload only then a Goods In shouldn't be generated.

addons/udes_delivery_control/models/stock_picking.py
PeterC10: Wrong field was used here, have raised a separate comment: https://github.com/unipartdigital/udes-open/pull/394#discussion_r533487629

PeterC10:
I think you will need to set `u_loading_type` to `unload_backload` in `delivery_control_picking_vals` for the unit tests.

https://github.com/unipartdigital/udes-open/blob/master/addons/udes_delivery_control/tests/common.py#L18

It might be worth adding a test for a backload only picking in https://github.com/unipartdigital/udes-open/blob/master/addons/udes_delivery_control/tests/test_delivery_control_workflow.py to make sure it doesn't generate a goods in picking.

addons/udes_delivery_control/models/stock_picking.py
kenvac: done

addons/udes_delivery_control/models/stock_picking.py
kenvac: The tests were failing because of this :) 
Done

PeterC10 (approved):
LGTM

addons/udes_delivery_control/models/picking_extra_fields.py
apatrick88: Yep computed fields are automatically readonly by default.

apatrick88 (approved):

PR #395, opened 2020-11-27
Story/11767 - Lock pickings once confirmed/completed

PeterC10:
Unlocked pickings (e.g. Goods In Pickings generated from completed Delivery Control pickings) are now locked once marked as todo or as an extra safety check, when the picking is completed.

Fixed issue with Stock Picking form override that was only hiding the Unlock button and not Lock.

Added Unit Test to ensure unlocked pickings are locked once confirmed or completed.


kenvac (approved):
LGTM

apatrick88 (approved):

PR #403, opened 2020-12-07
[11930] Adds support for multiple backload

kenvac:
Story: 11930


apatrick88 (approved):

addons/udes_delivery_control/controllers/main.py
```diff
@@ -19,12 +19,21 @@ def _get_picking_info(self, pickings, fields_to_fetch=None):
...
+            if "u_backload_ids" in fields_to_fetch:
+                picking_dict["u_backload_ids"] = [
+                    {"id": x.id, "supplier_name": x.supplier_id.name}
+                    for x in picking.u_backload_ids
+                ]
+            vals.append(picking_dict)
+        return vals
 
-    @http.route("/api/delivery-control/stock-picking/", type="json", methods=["GET"], auth="user")
+    @http.route("/api/delivery-control/", type="json", methods=["GET"], auth="user")
```
ws-mps, line 36: Can we also rename the class please? `PickingApi`

addons/udes_delivery_control/controllers/main.py
```diff
@@ -19,12 +19,21 @@ def _get_picking_info(self, pickings, fields_to_fetch=None):
...
     def get_pickings_delivery_control(self, fields_to_fetch=None, **kwargs):
```
ws-mps, line 37: I would also rename the function name.

addons/udes_delivery_control/controllers/main.py
```diff
@@ -37,7 +46,7 @@ def get_pickings_delivery_control(self, fields_to_fetch=None, **kwargs):
         return self._get_picking_info(pickings, fields_to_fetch)
 
     @http.route(
-        "/api/delivery-control/stock-picking/<int:ident>",
+        "/api/delivery-control/<int:ident>",
```
ws-mps, line 49: I would also rename the function name.

addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -284,6 +246,14 @@ def write(self, values):
 
     @api.model
     def create(self, values):
+        picking_type_dc = self.env.ref("udes_delivery_control.picking_type_delivery_control")
+        picking_type_id = values.get("picking_type_id")
+        if picking_type_id and picking_type_id == picking_type_dc.id:
+            # Set loading type to unload if picking is of delivery control type
+            # Note: u_loading_type is of related type field and compute attribute
+            # is not compatible/called. Hence, we are setting it's default value in create method
+            values["u_loading_type"] = "unload"
```
ws-mps, line 255: Could we just do this as default value for DC ?

addons/udes_delivery_control/controllers/main.py
kenvac: How does `DeliveryControlApi` sound?

addons/udes_delivery_control/models/stock_picking.py
kenvac: I tried all approaches but none seems to work
1. set default on fields - Using this method we have no way to get the value of picking type unless we set the context from where we create picking. 
2. Override default_get method - When the values are set in this method it gets changed by odoo somewhere as the field u_loading_type is a related field
3. compute on-field - Does not get called at all. Might be something to do with a related type of field

addons/udes_delivery_control/controllers/main.py
kenvac: done

addons/udes_delivery_control/controllers/main.py
kenvac: done

addons/udes_delivery_control/models/picking_extra_fields.py
```diff
@@ -134,10 +95,11 @@ def _check_unloading_dates(self):
...
 
-    @api.constrains("backload_start_date", "backload_end_date")
-    def _check_backload_dates(self):
-        msg = _("Backloading start time cannot be greater than end time")
-        validate_dates(self.backload_start_date, self.backload_end_date, msg)
+    @api.constrains("loading_type", "backload_ids")
+    def _check_backload_ids(self):
+        if self.is_backload and not self.backload_ids:
+            msg = _("Please add backloads.")
+            raise ValidationError((msg) % self.backload_ids)
```
ws-mps, line 102: Why do we need this ` % self.backload_ids` ?

addons/udes_delivery_control/models/stock_picking.py
ws-mps: Fair enough, worthed asking though :)

addons/udes_delivery_control/models/picking_extra_fields.py
kenvac: Removed

ws-mps (approved):

PR #405, opened 2020-12-08
Story/12000

williamsajr:
Add ability to move reserve pacakges

### DIFF
--- a/addons/udes_stock/tests/test_change_quant_location_wizard.py
+++ b/addons/udes_stock/tests/test_change_quant_location_wizard.py
@@ -14,10 +14,7 @@ class TestChangeQuantLocationWizard(common.BaseUDES):
         cls.product_quantity = 10
 
         cls.quant1 = cls.create_quant(
-            cls.apple.id,
-            cls.test_location_01.id,
-            cls.product_quantity,
-            package_id=cls.package1.id
+            cls.apple.id, cls.test_location_01.id, cls.product_quantity, package_id=cls.package1.id
         )
 
         cls.products_info = [{"product": cls.apple, "qty": cls.product_quantity}]
@@ -27,19 +24,23 @@ class TestChangeQuantLocationWizard(common.BaseUDES):
         due to permissions/settings"""
         with self.assertRaises(ValidationError) as e:
             cql.create_picking()
-        self.assertTrue(e.exception.name.startswith(
-            "Items are reserved and cannot be moved. "
-            "Please speak to a team leader to resolve the issue."
-        ))
+        self.assertTrue(
+            e.exception.name.startswith(
+                "Items are reserved and cannot be moved. "
+                "Please speak to a team leader to resolve the issue."
+            )
+        )
 
     def assert_cannot_create_partially_reserved(self, cql):
         """Assert we get a validation error when we try to move a reserved package
         as it is partially reserved"""
         with self.assertRaises(ValidationError) as e:
             cql.create_picking()
-        self.assertTrue(e.exception.name.startswith(
-            "Packages/Pallets must be either entirely reserved or unreserved."
-        ))
+        self.assertTrue(
+            e.exception.name.startswith(
+                "Packages/Pallets must be either entirely reserved or unreserved."
+            )
+        )
 
     def give_permission_to_move_reserved(self, user):
         """Setup the user and their warehouse to allow them to move reserved packages"""
@@ -52,9 +53,9 @@ class TestChangeQuantLocationWizard(common.BaseUDES):
         """Check that we can create a picking for a package"""
         self.assertFalse(self.quant1.reserved_quantity)
 
-        cql = self.change_quant_location.with_context(
-            active_ids=[self.package1.id]
-        ).create({"picking_type_id": self.picking_type_pick.id})
+        cql = self.change_quant_location.with_context(active_ids=[self.package1.id]).create(
+            {"picking_type_id": self.picking_type_pick.id}
+        )
         cql.create_picking()
 
         new_picking = self.package1.move_line_ids.picking_id
@@ -65,10 +66,7 @@ class TestChangeQuantLocationWizard(common.BaseUDES):
         """Check that we can create a picking for multiple packages"""
         package2 = self.create_package()
         quant2 = self.create_quant(
-            self.apple.id,
-            self.test_location_01.id,
-            self.product_quantity,
-            package_id=package2.id
+            self.apple.id, self.test_location_01.id, self.product_quantity, package_id=package2.id
         )
 
         self.assertFalse(self.quant1.reserved_quantity)
@@ -99,9 +97,9 @@ class TestChangeQuantLocationWizard(common.BaseUDES):
         test_picking.action_assign()
         self.assertEqual(self.quant1.quantity, self.quant1.reserved_quantity)
 
-        cql = self.change_quant_location.with_context(
-            active_ids=[self.package1.id]
-        ).create({"picking_type_id": self.picking_type_out.id})
+        cql = self.change_quant_location.with_context(active_ids=[self.package1.id]).create(
+            {"picking_type_id": self.picking_type_out.id}
+        )
 
         # We cannot create a picking without user permission for either value
         # of u_allow_create_picking_reserved_package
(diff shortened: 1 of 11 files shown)


addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
...
+        return self.create_picking()
+
+    def _check_reserved_packages_permitted(self):
+        """Check if a user is allowed to create a picking for reserved packages.
+        Returns a boolean.
+        """
+        User = self.env['res.users']
+
+        group_manage_reserved_packages = self.env.ref("udes_stock.group_manage_reserved_packages")
+        user_allowed = group_manage_reserved_packages in self.env.user.groups_id
```
apatrick88, line 65: Could also use `user_allowed = self.env.user.has_group("udes_stock.group_manage_reserved_packages")`

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
...
+                    "views": [
+                        (self.env.ref("udes_stock.view_move_reserved_package_check").id, "form")
+                    ],
+                    "view_id": self.env.ref("udes_stock.view_move_reserved_package_check").id,
+                    "target": "new",
+                    "res_id": self.id,
+                    "context": self.env.context,
+                }
+            else:
+                # Otherwise Use existing functionality to show full details on error
```
apatrick88, line 52: Super mega nit: `use` instead of `Use`.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -43,6 +88,24 @@ def create_picking(self):
...
+        # If we allow users to create these pickings for reserved quants, delete
+        # moves/movelines on other pickings.
+        # If we don't allow it the validation in create_picking will error
+        if self._check_reserved_packages_permitted():
+            # Remove moves/move lines for quants in other pickings
+            movelines = packages.get_move_lines()
+            # First split out the movelines from the original pickings
+            # to preserve the original picking and any moves that are for quants
+            # in different pacakges.
+            new_moves = Move.browse()
```
apatrick88, line 100: Can we give the `new_moves` variable a more descriptive name? This reads as though we're cancelling the new moves created for the new picking.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -43,6 +88,24 @@ def create_picking(self):
...
+            for move, movelines in movelines.groupby(lambda ml: ml.move_id):
```
ws-mps, line 101: Have we considered using backorder move lines so we don't cancel and re-reserve the stock ? This will imply that we have to remove quant_ids from params and create the picking before.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
     )
     reference = fields.Char('Reference')
 
+    @api.multi
+    def check_reserved(self):
+        """Check if we are moving reserved pacakges and if that is permitted"""
```
ws-mps, line 26: Typo: pacakges

addons/udes_stock/tests/test_change_quant_location_wizard.py
```diff
@@ -0,0 +1,108 @@
...
+        package2_picking = package2.move_line_ids.picking_id
+        self.assertEqual(package1_picking, package2_picking)
+        self.assertEqual(package1_picking.picking_type_id, self.picking_type_pick)
+        self.assertEqual(self.quant1.quantity, self.quant1.reserved_quantity)
+        self.assertEqual(quant2.quantity, quant2.reserved_quantity)
+
+    def test_create_picking_reserved_permissions(self):
+        """Check that a picking for a reserved package can only be created
+        with the correct settings and permissions."""
+        User = self.env['res.users']
```
PeterC10, line 73: Nit: Single quotes.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
...
+        self.ensure_one()
+        Package = self.env['stock.quant.package']
+
+        packages = Package.browse(self.env.context['active_ids'])
```
PeterC10, line 30: Nit: Mix of single/double quotes here and in `_check_reserved_packages_permitted`. Although as this is an existing file can understand we might not want to run black.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
     )
     reference = fields.Char('Reference')
 
+    @api.multi
+    def check_reserved(self):
+        """Check if we are moving reserved pacakges and if that is permitted"""
+        self.ensure_one()
+        Package = self.env['stock.quant.package']
```
PeterC10, line 28: (Feel free to ignore/resolve) Nit: I think `Package` should be placed above `self.ensure_one()` to follow convention. But looking around other methods it seems to be a mix over what should take precedent.

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -22,11 +21,57 @@ def _default_stock(self):
...
+                quants.assert_not_reserved()
+
+        # If nothing is reserved we can directly create the picking
+        return self.create_picking()
+
+    def _check_reserved_packages_permitted(self):
+        """Check if a user is allowed to create a picking for reserved packages.
+        Returns a boolean.
+        """
+        User = self.env['res.users']
```
PeterC10, line 62: Do you need `User`? Can you just do `self.env.user.get_user_warehouse()`?

addons/udes_stock/tests/test_change_quant_location_wizard.py
```diff
@@ -0,0 +1,108 @@
...
+        """Setup basic data"""
+        super(TestChangeQuantLocationWizard, cls).setUpClass()
+        Picking = cls.env["stock.picking"]
+        cls.change_quant_location = cls.env["change_quant_location"].sudo(cls.outbound_user)
+
+        cls.package1 = cls.create_package()
+        cls.quant1 = cls.create_quant(
+            cls.apple.id,
+            cls.test_location_01.id,
+            10,
```
PeterC10, line 17: (Feel free to ignore/resolve) Nit: Quantity of 10 is set here for `quant_1` and also for `quant_2` in `test_create_picking_multiple_packages`, is it worth adding in a class variable? e.g. `cls.product_quantity = 10`

addons/udes_stock/wizard/change_quant_location_view.xml
```diff
@@ -26,5 +26,24 @@
             </field>
         </record>
 
+        <record id="view_move_reserved_package_check" model="ir.ui.view">
+            <field name="name">Create Picking for Reserved Package(s)?</field>
+            <field name="model">change_quant_location</field>
```
PeterC10, line 31: Both `view_move_reserved_package_check` and `create_picking_for_packages_form` are form views for `change_quant_location`, but neither one is set explicitly as a default.

Could you set a priority on the least important one? As long as the priority is above the default 16 it should be fine.

e.g. `<field name="priority">20</field>`

addons/udes_stock/wizard/change_quant_location.py
williamsajr: I did consider and try that. The problem then was that we were adding quants at multiple points which caused issues with various validations. I did it this way so that all quants were added at once to avoid those issues

ws-mps: Is this when is not all the package reserved? I would restrict this to be only used in full reserved packages.

williamsajr: Or if we were trying to move a pallet when some packages were reserved and some weren't. I don't know if restricting it to just fully reserved packages is drawing an arbitrary line

apatrick88 (approved):

davidmcgaw (approved):
Tests LGTM!

ws-mps (approved):

PeterC10 (approved):
LGTM

addons/udes_stock/wizard/change_quant_location.py
```diff
@@ -1,55 +1,120 @@
...
-        # after location_dest_id is updated to the specific
-        # location set at self.location_dest_id
-        if self.location_dest_id:
-            new_picking.location_dest_id = self.location_dest_id
+        # If quants are reserved move them to a new picking
+        if total_quantity_reserved:
+            # Create a new picking
+            new_picking = Picking.create(params)
+            movelines = packages.get_move_lines()
+            movelines = movelines.filtered(lambda ml: ml.picking_id.state == "assigned")
```
ws-mps, line 104: `get_move_lines()` supports extra domain, so you can filter them out in the previous call.

williamsajr: Added this in

PR #418, opened 2020-12-21
[FIX] udes_stock: Sort location results in alphanumeric order

samclk:
Signed-off-by: Sam Clark <sam.clark@unipart.io>

### DIFF
--- a/addons/udes_stock/controllers/stock_move_line.py
+++ b/addons/udes_stock/controllers/stock_move_line.py
@@ -37,7 +37,7 @@ class StockMoveLineApi(UdesApi):
 
         if locations:
             response.append({"title": "",
-                             "locations": locations.sorted(lambda l: l.name).get_info()})
+                             "locations": locations.get_info()})
 
             if empty_locations:
                 # Ensure we don't have duplicates in the two recordsets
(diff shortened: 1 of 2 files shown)


joe-gb (approved):

nocturnalastro (approved):

addons/udes_stock/controllers/stock_move_line.py
```diff
@@ -37,14 +37,14 @@ def suggested_locations(self, move_line_ids):
 
         if locations:
             response.append({"title": "",
-                             "locations": locations.get_info()})
+                             "locations": locations.sorted(lambda l: l.name).get_info()})
```
ws-mps, line 40: For consistency, can we sort them in the function just before returning them?  This way in case it is used other places we will use them in the same order, example https://github.com/unipartdigital/udes-open/blob/master/addons/udes_stock/models/stock_picking.py#L1925

addons/udes_stock/controllers/stock_move_line.py
```diff
@@ -37,14 +37,14 @@ def suggested_locations(self, move_line_ids):
...
 
             if empty_locations:
                 # Ensure we don't have duplicates in the two recordsets
                 empty_locations = empty_locations - locations
 
         if empty_locations:
             response.append({"title": "empty",
-                             "locations": empty_locations.get_info()})
+                             "locations": empty_locations.sorted(lambda l: l.name).get_info()})
```
ws-mps, line 48: Same as above.

ws-mps (approved):

PR #419, opened 2020-12-21
[ADD] udes_stock: Add ability to change parent package for stock check

williamsajr:
User-story: 12266

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -21,7 +21,7 @@ VALID_SERIAL_TRACKING_QUANTITIES = [1, 0]
 ## Auxiliary types
 #
 
-StockInfoPI = namedtuple('StockInfoPI', ['product_id', 'package_id', 'lot_id'])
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -428,6 +435,7 @@ def _process_inventory_adjustments(self, adjustments_request):
               'product_qty':  quantity,
               'location_id':  self.id,
               'package_id':   stock_info.package_id,
+              'u_result_parent_package_id': stock_info.parent_package_id,
```
ws-mps, line 438: Do we have to set `u_package_parent_package_id` in case `stock_info.package_id` has it?

williamsajr: Now that stock inventory lines include this field we have to pass it on creation otherwise it would remove it from the pallet. Also, if a stock check reveals the package is on a different pallet (in the same location), this will create an inventory adjustment to change it in the system

PeterC10 (approved):
LGTM

ws-mps (approved):

PR #426, opened 2020-12-30
Story/11962

kdwyer:


ws-mps (approved):

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -513,6 +518,7 @@ def _prepare_info(self):
             - u_check_package_type: boolean
             - u_default_package_type_id: int
             - u_show_all_tasks: boolean
+            - u_enable_printing: boolean
```
nocturnalastro, line 521: Do all the print strategies respect this if not this could cause a lot of confusion.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -435,6 +435,11 @@ class StockPickingType(models.Model):
         help="Enable/Disable Transfer count button on kanban view",
     )
 
+    # Enable printing
+    u_enable_printing = fields.Boolean(
```
nocturnalastro, line 439: Do all the print strategies respect this if not this could cause a lot of confusion.

ws-mps: Agree, the name is confusing, because it is only to enable printing of delivery note on mobile, therefore normal strategies ignore this setting.

kdwyer: @nocturnalastro @ws-mps I've changed the name, please review.

nocturnalastro: I think it's clear enough now but it should probably be revisited allow the front end to query if a type of printing is enabled. Something like having `printable_report_ids` or something.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -435,6 +435,13 @@ class StockPickingType(models.Model):
         help="Enable/Disable Transfer count button on kanban view",
     )
 
+    # Enable printing of delivery notes
+    u_enable_delivery_note_printing = fields.Boolean(
+        string="Enable printing of delivery notes from the front end",
```
ws-mps, line 440: Enable front end DN printing? to shorten it a bit. We have the help for longer description

kdwyer: Text changed

nocturnalastro (approved):

apatrick88 (approved):

PR #429, opened 2021-01-05
[FIX] udes_priorities: Store picking_type_ids on priority groupds

nocturnalastro:


ws-mps (approved):

addons/udes_priorities/models/priority_groups.py
```diff
@@ -1,3 +1,4 @@
+from typing import Tuple
```
williamsajr, line 1: Why the new import?

joe-gb: beat me to it, it's not used? :O

nocturnalastro: I removed it :P I also noticed that there was a missing depends on the compute function

williamsajr (approved):

PR #433, opened 2021-01-07
[FIX] udes_priorities: Fix logic bug which caused get_piorities to re…

nocturnalastro:
…turn an empty list

### DIFF
--- a/addons/udes_priorities/models/stock_picking.py
+++ b/addons/udes_priorities/models/stock_picking.py
@@ -87,7 +87,7 @@ class StockPicking(models.Model):
         groups = []
 
         group_domain_kwargs = {}
-        if picking_type_id is None:
+        if picking_type_id is not None:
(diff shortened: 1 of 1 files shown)


addons/udes_priorities/models/stock_picking.py
```diff
@@ -1,3 +1,4 @@
+import ipdb
```
williamsajr, line 1: Adding more random imports again? :smile:

williamsajr: Also need to make the commit message under 80 characters

nocturnalastro: Done :)

williamsajr (approved):

ws-mps (approved):

unirho (approved):

PR #446, opened 2021-01-19
Story/12774 - User Password Management

PeterC10:
User Settings top level menu added.

Password Manager user group added.

Added Password Manager user group which allows users to edit the password of other users. Password Managers however are unable to edit the password of admin/debug users unless they are one themselves. In addition any user with Access Rights permissions will automatically get the new Password Manager group.

Password Management menu items added to User Settings menu. Simplified user views focussed on user password added for these menu items only. This avoids loading computed fields which require the user has edit rights for other users which will error for users that do not have the Access Rights group.

ChangePasswordUser overridden to check that the user has permission to update the selected user's password and to log the change or if permission was denied.

User Groups moved from data to security folder.

Added unit tests for Password Manager permissions.


addons/udes_security/tests/test_password_manager.py
```diff
@@ -0,0 +1,194 @@
...
+            - groups: Recordset of `res.groups` to apply to the user,
+                      if None only the default groups will be used
+        """
+        User = self.env["res.users"]
+
+        login = name.lower().replace(" ", "_")
+        user_vals = {"name": name, "login": login}
+
+        if groups:
+            group_ids = [group.id for group in groups]
```
kenvac, line 59: have you tried groups.ids work here?
It may or may not work just want to check

kenvac (commented):
Comments left

addons/udes_security/models/res_users.py
```diff
@@ -119,3 +119,71 @@ def _auth_timeout_check(self):
...
+            debug_group = "udes_security.group_debug_user"
+
+            pwd_manager = self.env.user.has_group(pwd_manager_group)
+            debug_user = self.env.user.has_group(debug_group)
+
+            if not pwd_manager:
+                # Non-password manager is trying to change the password of another user
+                self._log_and_raise_access_error(users)
+
+            if not debug_user:
```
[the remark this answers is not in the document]
PeterC10: Yep, debug user must be password manager in order to change other users passwords.

However a debug user will most likely have the "Access Rights" permission, which automatically inherits password manager anyway.

addons/udes_security/tests/test_password_manager.py
PeterC10: Good spot, thanks, have changed this. :)

kenvac (approved):

addons/udes_security/models/res_users.py
apatrick88, line 183: Who has permissions to add the password manager role to users? I think we want debug users to be able to do this and by extension it makes sense that they have the password manager group themselves.

PeterC10: Password Manager is a standard group so any user with the "Access Rights" group would be able to do this. Debug users would be setup with this anyway (as presently it is the only way to change other users' passwords as non-admin).

So debug users can make non-debug users Password Managers, as long as their permissions are setup appropriately.

apatrick88 (approved):

PR #449, opened 2021-01-29
[ADD] udes_stock: Add function to copy `active` from products to templates

squah-u:


addons/udes_stock/models/product_product.py
```diff
@@ -81,17 +79,47 @@ def get_product(self, product_identifier):
...
+        for product in self:
+            if product.active != product.product_tmpl_id.active:
+                if product.active:
+                    templates_to_activate |= product.product_tmpl_id
+                else:
+                    templates_to_deactivate |= product.product_tmpl_id
+
+        if activate_templates:
+            templates_to_activate.write({"active": True})
+        if deactivate_templates:
```
ws-mps, line 124: I would deactivate first, just in case we ever get in a point where there is no one2one relation.

squah-u: Updated the code

apatrick88 (approved):

ws-mps (approved):

PR #452, opened 2021-02-09
[ADD] udes_stock: Add option to only reserve full lines

williamsajr:
- When partial reservation of a picking is enabled there is
now the option to only allow this by entire move lines rather
than partial move lines

User-story: 12840

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -2409,8 +2409,18 @@ class StockPicking(models.Model):
         """
         # Assign at the move level because refactoring may change
         # the pickings.
+        Move = self.env["stock.move"]
+
         moves = self.mapped("move_lines")
         moves.with_context(lock_batch_state=True)._action_assign()
+
+        # Unreserve any partially reserved lines if not allowed by the picking type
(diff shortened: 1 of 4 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -2412,6 +2412,14 @@ def _reserve_stock_assign(self):
         moves = self.mapped("move_lines")
         moves.with_context(lock_batch_state=True)._action_assign()
 
+        # Unreserve any partially reserved lines if not allowed by the picking type
+        partial_moves = moves.filtered(lambda m: m.state == "partially_available")
+        moves_to_unreserve = partial_moves.filtered(
+            lambda m: m.picking_type_id.u_handle_partials
+            and not m.picking_type_id.u_handle_partial_lines
```
kenvac, line 2419: Would it not read `u_handle_partial_lines` config on every line. Can we not save it before and then compare.

williamsajr: Good point: I've changed this to group by picking type to avoid so many comparisions

kenvac (approved):

PR #454, opened 2021-02-17
[ADD] udes_stock: Inventory check for changed theoretical quantity

williamsajr:
- Check to see if theoretical quantity has changed since inventory was
started. This prevents incorrect adjustments being made after something
has been moved into/out of a location before adjustment

User-story: 10433

### DIFF
--- a/addons/udes_stock/models/stock_inventory.py
+++ b/addons/udes_stock/models/stock_inventory.py
@@ -20,6 +20,13 @@ class StockInventory(models.Model):
                                           readonly=True,
                                           index=True)
 
+    u_conflicting_theoretical_qty_message = fields.Html(
+        string="Conflicting Theoretical Quantities",
+        readonly=True,
+        store=False,
+        compute="_compute_conflicting_theoretical_qty_message",
+    )
+
     @api.multi
     def action_done(self):
         """
@@ -50,8 +57,24 @@ class StockInventory(models.Model):
 
     @api.multi
     def button_done(self):
-        """Add a popup to inform a user that they are adjusting reserved stock."""
+        """Add a popup to inform a user when
+           - The theoretical quantity has changed
+           - They are adjusting reserved stock.
+        """
         self.ensure_one()
+        if self._has_theoretical_quantity_changed():
+            return {
+                'name': _('Theoretical Quantity has Changed'),
+                'type': 'ir.actions.act_window',
+                'view_type': 'form',
+                'view_mode': 'form',
+                'res_model': 'stock.inventory',
+                'views': [(self.env.ref('udes_stock.view_theoretical_changed').id, 'form')],
+                'view_id': self.env.ref('udes_stock.view_theoretical_changed').id,
+                'target': 'new',
+                'res_id': self.id,
+                'context': self.env.context,
+            }
 
         if self._is_adjusting_reserved():
             return {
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_inventory.py
```diff
@@ -84,12 +104,41 @@ def _is_adjusting_reserved(self):
...
 
+    def _has_theoretical_quantity_changed(self):
+        """Check if any theoretical quantity in the location has changed.
+        """
+        self.ensure_one()
+        for line in self.line_ids:
+            current_qty = sum(line._get_quants().mapped("quantity"))
+            if (line.theoretical_qty or current_qty) and current_qty != line.theoretical_qty:
+                return True
+        return False
```
armandcela, line 115:
```suggestion
    def _has_theoretical_quantity_changed(self):
        """Check if any theoretical quantity in the location has changed.
                """
        self.ensure_one()
        return any(sum(line._get_quants().mapped("quantity")) != line.theoretical_qty for line in self.line_ids)
```

addons/udes_stock/models/stock_inventory.py
```diff
@@ -84,12 +104,41 @@ def _is_adjusting_reserved(self):
...
+        message = "<ul>"
+        for line in self.line_ids:
+            current_qty = sum(line._get_quants().mapped("quantity"))
+            if (line.theoretical_qty or current_qty) and current_qty != line.theoretical_qty:
+                message += "<li>" + line.location_id.name + ": " + line.product_id.name
+                message += " " + line.prod_lot_id.name if line.prod_lot_id else ""
+                message += " " + line.package_id.name if line.package_id else ""
+                message += "; Old Theoretical Quantity: " + str(line.theoretical_qty)
+                message += ", <b>New Theoretical Quantity: " + str(current_qty) + "</b>"
+                message += "</li>"
```
armandcela, line 131: I would suggest to use .format(), example 
message += "; Old Theoretical Quantity: " + str(line.theoretical_qty) == >
message += "; Old Theoretical Quantity: {}".format(line.theoretical_qty)

addons/udes_stock/models/stock_inventory.py
```diff
@@ -84,12 +104,41 @@ def _is_adjusting_reserved(self):
                 return True
         return False
 
+    def _has_theoretical_quantity_changed(self):
+        """Check if any theoretical quantity in the location has changed.
+        """
+        self.ensure_one()
+        for line in self.line_ids:
```
apatrick88, line 111: May need to override this in some customer repos e.g. to cover off quants with GRNs and ODNs

addons/udes_stock/models/stock_inventory.py
PeterC10: Might be worth also putting this logic into a `get_line_info` or similar function? That way might be a bit easier to deal with overrides for customers like @apatrick88  mentioned.

addons/udes_stock/models/stock_inventory.py
PeterC10: Yeah have added a comment here: https://github.com/unipartdigital/udes-open/pull/454#discussion_r577669955

Think it is worth refactoring into its own function to make it easier to add in/take out fields to display as needed.

williamsajr: This function should be ok as McLaren overwrites the function `_get_quants()` to return quants matching grn/odn

addons/udes_stock/models/stock_inventory.py
williamsajr: Not quite as we only want to do this check if either of the theoretical values is > 0 (we dont want to return true if we are adding a new quant). But I can definitely use some of this

addons/udes_stock/models/stock_inventory.py
PeterC10: Yep sorry got a bit tripped up there. We refactored the stock module to make it more easily overridable so that shouldn't be an issue.

I thought this comment was also referring to the message on the popup view, my bad. :)

addons/udes_stock/models/stock_inventory.py
armandcela: I think its sufficient != because if it is a new quant both will be 0 , theoritical and current_qty. It still will be false.
First control will be False only if line.theoretical_qty = 0 and current_qty = 0, which even on != control it will be false.

williamsajr: True, I think I originally got cufused with the many different types of quantity on inventory lines. I'll change to the suggested

williamsajr: Changed

addons/udes_stock/models/stock_inventory.py
williamsajr: Done both of these suggestions. The identifying information is now in a separate function that can be overwritten when needed

williamsajr: Also added translation of those strings at Peter's suggestion

armandcela (approved):

PeterC10 (approved):
LGTM

addons/udes_stock/models/stock_inventory.py
```diff
@@ -52,6 +59,19 @@ def action_done(self):
     def button_done(self):
         """Add a popup to inform a user that they are adjusting reserved stock."""
```
unirho, line 60: can we change this as this is no longer what it's doing?

unirho (changes requested):

unirho (approved):

PR #455, opened 2021-02-18
Story/10778 - Change Blocked File Type entity to Allowed File Type

PeterC10:
Files downloaded or uploaded via attachments or the binary field will now be blocked unless their extension is on the list of Allowed File Types.

Code updated for the required change of logic along with unit tests.


PeterAlabaster (approved):

addons/udes_security/tests/test_allowed_file_type.py
```diff
@@ -0,0 +1,148 @@
...
+    def create_allowed_file_type(self, file_type, **kwargs):
+        """Create allowed file type record from supplied file type"""
+        allowed_file_type_vals = {"name": file_type}
+        allowed_file_type_vals.update(kwargs)
+
+        allowed_file_type = self.AllowedFileType.create(allowed_file_type_vals)
+
+        return allowed_file_type
+
+    def test_allowed_file_active_status(self):
```
kdwyer, line 93: Nit: the comments are perhaps redundant?  a descriptive method name and docstring ought to be sufficient to document what is being tested.

addons/udes_security/tests/test_allowed_file_type.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_assert_allowed_file_type_names_formatted(self):
+        """
+        Assert that file type name is properly formatted, 
+        i.e. lowercase with full stop character removed
+        """
+        json_file_type = "json"
+        json_unformatted_file_types = ("JSON", ".json", ".JSON")
+
+        # Ensure that _format_name properly formats each unformatted file type
```
kdwyer, line 134: Consider using a [subtest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) here.

addons/udes_security/tests/test_allowed_file_type.py
PeterC10: Have renamed this test, although still not convinced about it. Have removed the comments also and added a more descriptive docstring instead.

I've also fixed an issue where it was checking if the TXT file was active rather than the CSV file, and also added in a check for RTF files (which are allowed), to show that both active/inactive attachments are unaffected if a different allowed file type record is made inactive.

addons/udes_security/tests/test_allowed_file_type.py
PeterC10: Thanks for the tip! Am using subtest now and looks much better. Also much nicer to debug if it fails.

Will have to keep that in mind for future. :)

kdwyer: Most of my code reviews of test code are "You could use subtest here" - for some reason no-one seems to know about it :shrug:

addons/udes_security/tests/test_allowed_file_type.py
kdwyer: Maybe split the test into two methods, as it's testing two things?  That should make it easier to name.

addons/udes_security/models/allowed_file_type.py
```diff
@@ -4,13 +4,13 @@
 from odoo.exceptions import UserError
 
 
-class BlockedFileType(models.Model):
```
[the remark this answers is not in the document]
PeterC10: Yep will address this at the end, once the changes made are final. :)

addons/udes_security/tests/test_allowed_file_type.py
PeterC10: I've now split the test into `test_assert_allowed_attachment_set_active` and `test_assert_blocked_attachment_set_inactive`. 
I also removed `test_allowed_file_type_deleted` as this is now covered in `test_assert_blocked_attachment_set_inactive`.

PeterC10:
I've added a function `_set_blocked_attachments_to_inactive` which runs after `udes_security` is installed.

This is needed because attachments with a file type that is not allowed may exist before the module installed, so these will need to be identified and then set inactive.

This wasn't a problem before when we had blocked file types, because as soon as the blocked file type records were added it would then identify any attachments that needed to be set inactive.

addons/udes_security/tests/test_allowed_file_type.py
```diff
@@ -93,36 +95,45 @@ def create_allowed_file_type(self, file_type, **kwargs):
...
 
-    def test_assert_attachment_active_status_set_from_allowed_file_type(self):
+    def test_assert_allowed_attachment_set_active(self):
         """
-        Assert that an attachment is active if its file type is allowed, otherwise inactive.
+        Assert that an attachment with an allowed file type is set to active.
 
-        Also assert that when the active field on allowed file type is updated,
-        the active field for attachments of that file type only is also updated.
+        Also assert that the attachment is also set to active on either of the following conditions:
```
kdwyer, line 102: Supernit: too many alsos.

addons/udes_security/tests/test_allowed_file_type.py
```diff
@@ -93,36 +95,45 @@ def create_allowed_file_type(self, file_type, **kwargs):
...
-        self.assertNotEqual(self.atch_csv_file.active, csv_active)
+        # Create HTML allowed file type record, unblocking HTML files
+        self.create_allowed_file_type(("html"))
+        self.assertTrue(self.atch_html_file.active)
 
-        self.assertEqual(self.atch_rtf_file.active, rtf_active)
-        self.assertEqual(self.atch_txt_file.active, txt_active)
-
-    def test_allowed_file_type_deleted(self):
+    def test_assert_blocked_attachment_set_inactive(self):
```
kdwyer, line 117: Nit: "assert" is redundant in the method name, since the purpose of test methods is to assert things. 

I tend to prefer something like:  `test_sets_blocked_attachment_inactive`, where the verb comes first, and describes what the feature under test is doing.  It is a matter of style though, so no obligation to do it that way.

addons/udes_security/tests/test_allowed_file_type.py
PeterC10: Have now reduced the also usage. :)

addons/udes_security/tests/test_allowed_file_type.py
PeterC10: Yeah fair point, often find the naming tricky to be honest. I just got into the habit of using "assert" in the name a lot of the time. Have updated the names of the tests. :)

kdwyer: I found the [GOOS book](http://www.growing-object-oriented-software.com/) really helpful for thinking about tests, if you don't mind Java.

kdwyer (approved):

addons/udes_security/models/allowed_file_type.py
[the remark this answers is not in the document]
PeterC10: I've now rebased the commits. The first commit now moves the files from `blocked_xyz` to `allowed_xyz` and updates the manifest/init files where needed.

No changes are made to the files until the later commits (with the previous to be squashed commits not squashed).

PR #457, opened 2021-02-23
Story/12253

BlommaertsEdwin:


addons/edi_notifier/models/email_notifier.py
```diff
@@ -35,16 +39,24 @@ def _notify(self, notifier, event_type, recs):
                 template = template.with_context(attachments=attachments)
             email_values = None
             if attachments:
-                email_values = {'attachment_ids': attachments.mapped('id')}
+                email_values = {"attachment_ids": attachments.mapped("id")}
             template.send_mail(rec.id, force_send=True, email_values=email_values)
 
+    def check_disbaled_edi_notifier(self):
```
kenvac, line 45: NIT: spell check `check_disabled_edi_notifier` ?

addons/edi_notifier/models/email_notifier.py
```diff
@@ -35,16 +39,24 @@ def _notify(self, notifier, event_type, recs):
...
+
     @api.multi
     def notify(self, notifier, event_type, recs):
-        """Filter records and send them for notification"""
-        if not recs:
-            recs = notifier.doc_type_ids
-        recs = self.filter_records(notifier, event_type, recs)
-        self._notify(notifier, event_type, recs)
+
+        IrConfig = self.env["ir.config_parameter"].sudo()
```
kenvac, line 51: NIT: Unused variable `IrConfig`

PeterAlabaster (approved):

addons/edi_notifier/models/email_notifier.py
```diff
@@ -17,8 +17,12 @@ def _should_notify(self, notifier, event_type, rec):
 
     @api.multi
     def _notify(self, notifier, event_type, recs):
+        instance_url = self.env['ir.config_parameter'].sudo().get_param('web.base.url')
```
PeterC10, line 20: Nit: `self.env['ir.config_parameter']` should be put into its own variable at the top of the method, e.g.:

`ConfigParameter = self.env["ir.config_parameter"]`

addons/edi_notifier/tests/test_edi_notifier.py
```diff
@@ -51,25 +56,70 @@ def create_transfer(self, doc):
...
+        cls.email_template = cls.env.ref(
+            "edi_notifier.email_template_edi_document_not_received"
+        )
+        cls.notifier = EdiNotifier.create(
+            {
+                "name": "Email on test doc missing",
+                "model_id": IrModel._get_id("edi.notifier.email.missing.in.range"),
+                "doc_type_ids": [(6, False, cls.doc_type.ids)],
+                "template_id": cls.email_template.id,
+                "lookback_hours": 3,
```
kenvac, line 107: Just a suggestion - Can we make this configurable rather than static value?

addons/edi_notifier/models/email_notifier.py
```diff
@@ -17,8 +17,12 @@ def _should_notify(self, notifier, event_type, rec):
...
+        context = self.env.context.copy()
+        context.update({'instance_url': instance_url})
```
ws-mps, line 22: Formatting: still a few ' instead of "

addons/edi_notifier/tests/test_edi_notifier.py
```diff
@@ -51,25 +56,70 @@ def create_transfer(self, doc):
...
+                "active": True,
+            }
+        )
+
+    @mock.patch(
+        "odoo.addons.edi_notifier.models.email_notifier.EdiEmailNotifier.check_disbaled_edi_notifier",
+        return_value=True,
+    )
+    def test_no_email_sent_on_success_when_safety_set_to_true(self, mock_config):
+        self.setup_cron(datetime.now() - timedelta(hours=1))
```
kenvac, line 117: Q: Any reason we are setting time in the past and then directly triggering it ?

kenvac (changes requested):
Comments added

addons/edi_notifier/models/email_notifier.py
```diff
@@ -17,8 +17,12 @@ def _should_notify(self, notifier, event_type, rec):
...
         for rec in recs:
-            template = notifier.template_id
+            
+            template = notifier.template_id.with_context(context)
```
ws-mps, line 25: Probably you don't need the copy of the context of line 21, instead you can do:
```
template = notifier.template_id.with_context(instance_url=instance_url)
```
Which should create a new context.

addons/edi_notifier/models/email_notifier.py
```diff
@@ -35,16 +39,24 @@ def _notify(self, notifier, event_type, recs):
                 template = template.with_context(attachments=attachments)
             email_values = None
             if attachments:
-                email_values = {'attachment_ids': attachments.mapped('id')}
+                email_values = {"attachment_ids": attachments.mapped("id")}
             template.send_mail(rec.id, force_send=True, email_values=email_values)
 
+    def check_disbaled_edi_notifier(self):
+        return config.get_misc("edi", "disable_edi_notifications", False)
```
squah-u, line 46: Does the story want notifications disabled or enabled by default?

PeterC10: It is a bit ambiguous but the story says "I don't want notifiers to be enabled by default on my UAT instance" so I guess it just means they have to be enabled manually on the live instance (Rather than having to manually turn them off for UAT).

ws-mps, line 46: For the other safety catches (edi, print) there is a setting where the safety catch name can be changed. Would be good to follow that approach instead of hardcoding the safety catch name.

joe-gb: I think here we can add notifiers on customer instances, we will make it based on a config field against the email_notifier object

joe-gb:
can you please squash and tidy up the commits? we probably only need 1 for auto formatting, 1 for the code change, 1 for the test

addons/edi_notifier/models/email_notifier.py
ws-mps: Yes, please. And follow the same behaviour as the other safety catches.

addons/edi_notifier/tests/test_edi_notifier.py
ws-mps: This is a test, do we need config?

addons/edi_notifier/models/notifier.py
```diff
@@ -167,13 +178,28 @@ def action_view_cron(self):
         }
         return action
 
+    @api.multi
+    def check_disbale_edi_notifications(self):
```
ws-mps, line 182: Typo: disbale --> disable

addons/edi_notifier/models/notifier.py
```diff
@@ -167,13 +178,28 @@ def action_view_cron(self):
...
+
     @api.multi
     def notify(self, event_type, recs=None):
-        for notifier in self:
-            if notifier.active:
-                self.env[notifier.model_id.model].notify(
-                    notifier, event_type, recs
-                )
+        """Check for edi notification safety"""
+        if not self.check_disbale_edi_notifications():
```
ws-mps, line 197: Typo: disbale --> disable

addons/edi_notifier/models/notifier.py
```diff
@@ -167,13 +178,28 @@ def action_view_cron(self):
         }
         return action
 
+    @api.multi
+    def check_disbale_edi_notifications(self):
+        """Check safety config parameter is present and return True if it's enable, otherwise False."""
+        self.ensure_one()
+        if self.safety:
+            section, _sep, key = self.safety.rpartition('.')
+            if config.get_misc(section or self._name, key):
```
ws-mps, line 187: This will disable if the safety catch is in the config file, shouldn't it be the other way around ?

addons/edi_notifier/models/notifier.py
ws-mps: This check should be inside the loop, otherwise this function allows self to be nothing or more than one element in the recordset, and `check_disbale_edi_notifications()`  ensures one.

addons/edi_notifier/models/notifier.py
BlommaertsEdwin: I'm not sure I follow what you mean here.
If the safety catch is set in the config file and it is set to True, this will disable the notifications, vice versa if set to False.
It will only disable if the safety catch is set through the interface.

addons/edi_notifier/tests/test_edi_notifier.py
joe-gb: this wasn't part of this PR anyway, just formatting I think!

addons/edi_notifier/models/notifier.py
```diff
@@ -168,12 +169,28 @@ def action_view_cron(self):
...
+                    self.name,
+                    ".".join((section or self._name, key)),
                 )
+                return False
+        return True
+
+    @api.multi
+    def notify(self, event_type, recs=None):
+        """Check for edi notification safety"""
+        if self.check_edi_notifications_enabled():
```
ws-mps, line 190: his check should be inside the loop, otherwise this function allows self to be nothing or more than one element in the recordset, and `check_edi_notifications_enabled()` ensures one.

PeterC10 (approved):
LGTM

ws-mps (approved):

joe-gb (approved):

kenvac (approved):

PR #459, opened 2021-03-17
Story/13267

peterpratt13:


addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1032,6 +1032,157 @@ def test31_backorder_generated_unpickable_item_product_ok_in_two_picks_and_two_p
...
+
+        # Create a list of packages used for the quants, and create a picking for each product
+        # that takes everything created
+        apple_packages = [self.create_package() for i in range(4)]
+        banana_packages = [self.create_package() for i in range(4)]
+        for idx, (apple_package, banana_package) in enumerate(
+            zip(apple_packages, banana_packages), start=1
+        ):
+            self.create_quant(
+                self.apple.id, self.test_location_01.id, idx, package_id=apple_package.id
```
ws-mps, line 1083: I'm guessing we are using idx as quantity. For clarity of the test I suggest to create a variable quantity and use it when calling `create_quant()`.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1032,6 +1032,157 @@ def test31_backorder_generated_unpickable_item_product_ok_in_two_picks_and_two_p
...
+        """
+        Picking = self.env["stock.picking"]
+        Batch = self.env["stock.picking.batch"]
+        Group = self.env["procurement.group"]
+        # Cron job that deletes empty draft pickings
+        unlink_cron = self.env.ref("udes_stock.unlink_empty_pickings_action")
+        unlink_cron.active = False
+
+        # Create a list of packages used for the quants, and create a picking for each product
+        # that takes everything created
```
ws-mps, line 1076: Extend the comment to clarify that the packages contain 1, 2, 3 and 4 units of each product, for a total of 10.

addons/udes_stock/tests/test_picking_batch.py
```diff
@@ -1032,6 +1032,157 @@ def test31_backorder_generated_unpickable_item_product_ok_in_two_picks_and_two_p
...
+            inv_pickings, 8, "assigned", [1, 2, 3, 4], package_list=apple_packages + banana_packages
+        )
+        # Check the pickings are correct - two separate ones grouped by product, it merges
+        # everything into the created picking with the group.
+        # Trigger the cron job to clean up the draft pickings, and remove the backorder id
+        unlink_cron.method_direct_trigger()
+        pick_pickings = Picking.search([("picking_type_id", "=", self.picking_type_pick.id)])
+        self._assert_apple_banana_pickings(pick_pickings, 2, "confirmed", [1, 2, 3, 4])
+        for p in pick_pickings:
+            with self.subTest(p=p):
```
PeterC10, line 1182: Suggestion (feel free to resolve):

Might be worth refactoring the check for `backorder_id` and `group_id` into a method, as this is similar to the checks done on line 1156.

e.g. something like:

`confirmed_pickings.assert_backorder_and_group(expected_backorder=True, expected_group=True)`

and

`pick_pickings.assert_backorder_and_group(expected_backorder=False, expected_group=True)`

PeterC10 (approved):
LGTM

ws-mps (approved):

addons/udes_stock/tests/test_picking_batch.py
peterpratt13: Thanks, have left this as it was only used twice and having the branches to filter down the assertTrue/False route seemed more work than :)

PR #463, opened 2021-03-29
[FIX] udes_open: Changing the error message

armandcela:
with scanned package instead of expected package

Story: 12656

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1775,8 +1775,8 @@ class StockPicking(models.Model):
 
         if not scanned_package.has_same_content(expected_package):
             raise ValidationError(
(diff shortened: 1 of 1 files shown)


apatrick88 (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -1776,7 +1776,7 @@ def _get_mls_for_entire_package_swap(self, scanned_package, expected_package):
         if not scanned_package.has_same_content(expected_package):
             raise ValidationError(
                 _("The contents of %s does not match what you have been asked " "to pick.")
```
PeterC10, line 1778: Nit: Not related to your change but while you're in here could we change ```" "to pick."``` to ```to pick."```?

PeterC10 (approved):

PR #465, opened 2021-04-06
Story/10434

kdwyer:


ws-mps (approved):

PeterAlabaster (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -267,6 +274,12 @@ def _compute_state(self):
 
         super()._compute_state()
 
+    @api.depends('state')
+    @api.one
+    def _compute_is_locked(self):
+        """Derive is_locked from the picking's state."""
+        self.is_locked = self.state != 'draft'
```
PeterC10, line 281: Nit: Evil single quotes.

addons/udes_stock/models/stock_picking.py
```diff
@@ -267,6 +274,12 @@ def _compute_state(self):
 
         super()._compute_state()
 
+    @api.depends('state')
+    @api.one
```
PeterC10, line 278: Q: Is there any chance this would error? Don't computed fields usually loop through a recordset in self?

kdwyer: I was concerned that it might error, or recurse endlessly, so I used the same structure as the immediately preceding `_compute_state` method.

addons/udes_stock/models/stock_picking.py
ws-mps: Also in `compute='_compute_is_locked',`

ws-mps (approved):

PeterC10 (approved):
LGTM

PR #468, opened 2021-04-09
[IMP] udes_open: Adding last reserved pallet name on batch info

armandcela:
Field is needed for an improvement to guide the user if has interrrupted
the process and picked some already with a pallet , so that user
goes and scanns last pallet
Story: 13659

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -575,6 +575,7 @@ class StockPickingBatch(models.Model):
             "result_package_names": pickings.get_result_packages_names(),
             "u_original_name": self.u_original_name,
             "has_done_pickings": has_done_pickings,
+            "u_last_reserved_pallet_name": pickings.get_result_packages_names() and self.u_last_reserved_pallet_name or False,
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -575,6 +575,7 @@ def _prepare_info(self, allowed_picking_states):
             "result_package_names": pickings.get_result_packages_names(),
             "u_original_name": self.u_original_name,
             "has_done_pickings": has_done_pickings,
+            "u_last_reserved_pallet_name": pickings.get_result_packages_names() and self.u_last_reserved_pallet_name or False,
```
squah-u, line 578: Since this is going to be JSON'd, do we want to use `or None` so that we emit a JSON `null`?

squah-u (approved):

apatrick88 (approved):

armandcela:
I am closing this MR as we found an issue while signing off , and last pallet cannot be populated like i have done on this MR.
Changes to populate i did in udes_workflows with calling a super.

PR #471, opened 2021-05-17
[FIX] udes_stock; potential MissingError when _action_refactor is called

PeterAlabaster:
User-story: 13752


PeterAlabaster:
Closing as this fix breaks other tests due to `invalidate_cache()` not always being scoped into `Environment`

PR #479, opened 2021-06-10
[FIX] udes_stock: Add optional limit and sort to suggested locations

joe-gb:
Allowing sorting and limit optionally in suggested_locations, sorting
is for all policies, limit only for those that accept it (story/14394)

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -2162,14 +2162,14 @@ class StockPicking(models.Model):
 
         return location
 
-    def _get_suggested_location_empty_location(self, limit=None):
+    def _get_suggested_location_empty_location(self, limit=None, sort=False):
         """Get suggested locations by finding locations which are empty.
         """
-        return self.get_empty_locations(limit=limit)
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -2113,18 +2113,36 @@ def get_suggested_locations(self, move_line_ids):
...
-                        result = func(move_line_ids)
-
-        return result.sorted(lambda l: l.name)
+                        # The signature of each "get suggested location" method may not
+                        # match, so inspect the method and only pass in the expected args
+                        expected_args = inspect.getfullargspec(func)[0]
+                        kwargs = {}
+                        for var in ["move_line_ids", "limit", "sort"]:
+                            if var in expected_args:
+                                kwargs[var] = eval(var)
```
unirho, line 2122: O________________________________________o

unirho: how about using locals? and get these values off the dictionary?

ws-mps (approved):

PeterAlabaster (approved):

PR #487, opened 2021-07-20
[ADD] udes_stock: Can refactor pickings by maximum quantity post assign

williamsajr:
User-story: 13472

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from odoo import api, models, fields, _
-from odoo.exceptions import UserError
+from odoo.exceptions import UserError, ValidationError
 
 import logging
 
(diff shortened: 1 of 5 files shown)


apatrick88 (approved):

addons/udes_stock/models/stock_move.py
```diff
@@ -654,3 +654,40 @@ def _refactor_action_batch_pickings_by(self, by_key):
...
             # Associate picking to the batch.
             picking.write({"batch_id": batch.id})
+
+    def refactor_action_by_maximum_quantity(self):
+        """Split moves to have a maximum quantity"""
+        picking_type = self.mapped("picking_type_id")
+        picking_type.ensure_one()
+
+        if not picking_type.u_assign_refactor_constraint_value:
+            return
```
peterpratt13, line 664: Perhaps we should return self here if we don't do anything.

williamsajr: This called on line 271. The expectation in this function is that if one of the `refactor_action` methods returns nothing then nothing has been refactored and moves won't be updated. I could return `self` here and it would still work but there would be some extra logic on line 273-275 triggered; removing self from a recordset and then readding it

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1136,3 +1136,62 @@ def _check_resultant_package_level(self):
...
+                    new_ml = ml._split_by_qty(maximum_qty)
+                    grouped_mls.append(new_ml)
+                # If there is not a remainder, ml has already been grouped
+                # so remove from self
+                if not remainder:
+                    self -= ml
+
+        # Finally split and group the remainder to meet the maximum
+        quantity = 0
+        movelines = self.browse()
```
peterpratt13, line 1172: Did anyone ever decide on a convention for this? MoveLine=self? If we did not, ignore

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -762,3 +762,118 @@ def test01_auto_unlink_empty_pickings(self):
...
+            ) for pick in all_pickings
+        ]
+        expected_quantities_ordered_and_reserved = [
+            (2, 2), (2, 2), (2, 2), (1, 1), (3, 0)
+        ]
+        self.assertCountEqual(
+            quantities_ordered_and_reserved, expected_quantities_ordered_and_reserved
+        )
+
+    def test_split_mixed_products(self):
```
peterpratt13, line 846: Can we get a similar unit test but the moves are for more just to cover that extra case

addons/udes_stock/models/stock_move_line.py
williamsajr: I think the poll ended up split with no clear decision. So I've assumed either is fine. I personally don't like `MoveLine=self` as that may not be clear whether `MoveLine` is empty or not

peterpratt13 (approved):

PR #491, opened 2021-08-03
[ADD] udes_open: Changes on task type

armandcela:
Changes on task type when field u_user_scans is not set on picking type
Adding a new type general in this case
Story: 10456

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -865,7 +865,12 @@ class StockMoveLine(models.Model):
         task = {
             "picking_id": picking.id,
         }
-        user_scans = picking.picking_type_id.u_user_scans
+
+        # Check if user_scans is manually set in context first
+        user_scans = self.env.context.get("user_scans")
+
+        if not user_scans:
+            user_scans = picking.picking_type_id.u_user_scans
 
         if user_scans == "product":
             task["type"] = "product"
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -42,11 +42,16 @@ class StockMoveLine(models.Model):
         return self.filtered(lambda ml: ml.qty_done == ml.product_uom_qty)
 
     def _prepare_result_packages(
-        self, package, result_package, result_parent_package, products_info
+        self,
+        package,
+        result_package,
+        result_parent_package,
+        products_info,
+        target_storage_format=None,
     ):
-        """ Compute result_package and result_parent based on the
-            u_target_Storage_format of the picking type + the input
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -42,21 +42,27 @@ def get_lines_done(self):
...
+        result_package,
+        result_parent_package,
+        products_info,
+        target_storage_format=None,
     ):
-        """ Compute result_package and result_parent based on the
-            u_target_Storage_format of the picking type + the input
-            parameters.
+        """
+        Compute result_package and result_parent based on the u_target_Storage_format of the
```
kdwyer, line 53:
```suggestion (typo in field name)
        Compute result_package and result_parent based on the u_target_storage_format of the 
```

PeterC10: Good spot, fixed. :)

addons/udes_stock/models/stock_move_line.py
```diff
@@ -861,11 +871,7 @@ def _prepare_task_info(self):
         }
         user_scans = picking.picking_type_id.u_user_scans
 
-        if user_scans == "product":
-            task["type"] = "product"
-            task["pick_quantity"] = sum(self.mapped("product_qty"))
-            task["quant_ids"] = self.get_quants().get_info()
-        else:
+        if user_scans in ["package", "pallet"]:
```
kdwyer, line 874: I may be over-thinking this, but the way we are using `user_scans` doesn't feel quite right to me: we seem to be using it as a proxy for some unknown abstraction?  I'm not sure what it might be, but it feels (to me) like the task type ought to be computed by the picking type  rather than derived from `user_scans`.  Possibly also the guards on the old behaviour.

PeterC10: Yep I get what you're saying. Although not sure what you mean by computed by the picking type? Do you mean we should just be looking at the move line. e.g. if the ML has a product in a package, we should be scanning a package?

kdwyer: I mean something like

```python
class StockPickingType(Model):

    @property
    def task_type(self):
        if self.user_scans in ['package', 'pallet']:
            return 'package'
        if self.user_scans:
            return 'package'
        return 'general'
```

assuming that wouldn't break anything...

kenvac (approved):
LGTM

addons/udes_stock/models/stock_move_line.py
PeterC10: Ah gotcha, let me try it out. :)

williamsajr (approved):

PR #500, opened 2021-09-07
Story/14868

kenvac:
Adds new config on stock.picking.type to not allow pickings with destination location block on batch

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1547,12 +1547,7 @@ class StockPicking(models.Model):
         _logger.info(msg)
 
     def _swap_moveline_info_at_pack_level(self, pack, values):
-        values.update(
-            {
-                "package_id": pack.id,
-                "result_package_id": pack.id,
-            }
-        )
+        values.update({"package_id": pack.id, "result_package_id": pack.id})
         return values
 
     def _swap_moveline_info_below_pack_level(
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -2463,13 +2463,16 @@ def search_for_pickings(self, picking_type_id, picking_priorities, limit=1, doma
...
                 search_domain.append(("u_location_category_id", "child_of", categories.ids))
 
-        search_domain.extend(
-            [
-                ("picking_type_id", "=", picking_type_id),
-                ("state", "=", "assigned"),
-                ("batch_id", "=", False),
-            ]
-        )
+        extra_filter = [
```
armandcela, line 2466: To be consistent in the way search_domain have been extended i would suggest not to use a new variable extra_filter, but directly extend search_domain if picking_type.u_batch_dest_loc_not_allowed

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -429,6 +429,12 @@ class StockPickingType(models.Model):
         default=False,
         help="If set, the user will be shown the list of all tasks during the picking workflow",
     )
+    u_batch_dest_loc_not_allowed = fields.Boolean(
+        string="Do not allow pickings with block destination location",
```
PeterC10, line 433: This string is too long, won't display nicely on UI. Could you simplify the title? We always have the help string for a more in-depth explanation.

Also please remember to capitalise the string, capitalizemytitle.com is a useful tool for this. :)

PeterC10:
General nit: Could you shorten the title of the first commit? It should be no more than 72 characters based on our CR checklist.

I think you could also fixup the second commit into the first.

addons/udes_stock/models/stock_picking_type.py
kenvac: How about "No blocked location pickings" sounds to you?

PeterC10: I think we need to include the fact it's for destination locations, how about `No Blocked Dest. Location Pickings`? Still a bit long but cut down a fair bit from before.

Might be worth getting another opinion if unsure.

armandcela (approved):

PeterC10 (approved):
LGTM

PR #503, opened 2021-09-16
[ADD] udes_stock: Adding Mobile Type field in stock inventory model

armandcela:
Story: 14975

### DIFF
--- a/addons/udes_stock/models/stock_inventory.py
+++ b/addons/udes_stock/models/stock_inventory.py
@@ -26,6 +26,18 @@ class StockInventory(models.Model):
         store=False,
         compute="_compute_conflicting_theoretical_qty_message",
     )
+    u_stock_check_origin = fields.Selection(
+        string="Stock Check Origin",
+        selection=[
+            ("desktop", "Desktop"),
+        ],
+        default="desktop",
+        help="""Stock check origin"""
+    )
+    u_date_start_scanned = fields.Datetime(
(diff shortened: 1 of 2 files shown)


apatrick88 (approved):

addons/udes_stock/models/stock_inventory.py
```diff
@@ -26,6 +26,22 @@ class StockInventory(models.Model):
         store=False,
         compute="_compute_conflicting_theoretical_qty_message",
     )
+    u_mobile_type = fields.Selection(
+        string="Mobile Type",
+        selection=[
+            ("mobile", "Mobile"),
+            ("workflow", "Workflow"),
```
williamsajr, line 33: Would it be better to add the `workflow` selection in the workflows module?

armandcela: For sure it will be needed like a field when udes_workflows is not installed, as we will filter with this field stock checks have been created from old mobile as well and udes_workflows may not be even installed.
Yes it can be done to add a selection in udes_workflows, but i think its just adding complexity as we already know the reason of using the field. We agree to be added in udes_stock

adamrpatrick (changes requested):
I think we can put these fields into a different module.

williamsajr (approved):

adamrpatrick (approved):

PR #505, opened 2021-09-17
[IMP] udes_stock: Compute _compute_state improved for checking quants

PeterC10:
On databases with a large amount with locations with a high number quants the "location.quant_ids" can slow down the initial computation significantly. As we only care about whether a location has any quants, doing a search count will speed this check up significantly.

Story/15029

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -204,7 +204,7 @@ class StockLocation(models.Model):
     )
 
     @api.multi
-    @api.depends("quant_ids", "u_blocked")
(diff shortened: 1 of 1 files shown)


kenvac (approved):

kenvac:
Good spot Pete.
Search is always faster than browse in my experience.

PeterC10:
> Good spot Pete.
> Search is always faster than browse in my experience.

Wish I had spotted it! Joe found it on Friday. :) Usually we can get away with it but some locations have thousands of quants, something to try and keep in mind in future. :)

williamsajr (approved):

PR #507, opened 2021-09-21
[ADD] udes_stock: Adding user_id on stock check model

armandcela:
1. Adding new field user_id (Commented on existing module)
2. Creating stock check even when it is counted correctly
3. Setting date start of stock check on location that is in process

Story: 15398


addons/udes_stock/controllers/stock_location.py
```diff
@@ -41,6 +41,9 @@ def get_location(self, location_id=None, location_name=None,
         if check_blocked:
             location.check_blocked()
 
+        # Updating scanning start date on location
```
kenvac, line 44: Super NIT: Update instead of Updating

addons/udes_stock/models/stock_inventory.py
```diff
@@ -10,6 +10,13 @@ class StockInventory(models.Model):
     _name = 'stock.inventory'
     _inherit = 'stock.inventory'
 
+    @api.model
+    def _get_stock_check_users(self):
+
```
kenvac, line 15: Missing doc string

addons/udes_stock/models/stock_inventory.py
```diff
@@ -38,6 +45,14 @@ class StockInventory(models.Model):
         "Start Date Scanning",
         help="The date operator started scanning."
     )
+    user_id = fields.Many2one(
+        "res.users",
+        string="Checked Owner",
+        readonly=True,
+        states={"draft": [("readonly", False)], "waiting": [("readonly", False)]},
+        help="Specify Owner to focus your stock check on a particular user.",
```
kenvac, line 53: Super NIT: owner and not Owner

kenvac, line 53: Nit: we can make the help string simpler e.g "Stock check owner"

addons/udes_stock/models/stock_inventory.py
```diff
@@ -260,6 +275,9 @@ def write(self, values):
             raise UserError(_("Cannot write to an adjustment which has already been validated"))
         return super(StockInventory, self).write(values)
 
+    def get_stock_check_event_states(self):
+        return ["confirm", "done"]
```
kenvac, line 279: doc string please

addons/udes_stock/models/stock_location.py
```diff
@@ -588,6 +593,12 @@ def _process_inventory_adjustments(self, adjustments_request):
                     "prod_lot_id": stock_info.lot_id,
                 }
             )
+            inventory_line_ids.append(inventory_line.id)
+        inventory_data = self.get_inventory_adjustment_data()
+        inventory_data.update({
```
kenvac, line 598: `inventory_data["line_ids"] = [(6, 0, inventory_line_ids)]` is faster than {}.update()

addons/udes_stock/models/stock_location.py
```diff
@@ -597,6 +608,18 @@ def get_inventory_adjustment_data(self):
             "location_id": self.id,
             "filter": "none",
             "state": "confirm",
+            "user_id": self.env.user.id,
+            "u_date_start_scanned": self.u_date_start_scanned,
+        }
+
+    def get_inventory_validated_data(self):
+        return {
```
kenvac, line 616: doc string please

addons/udes_stock/models/stock_location.py
```diff
@@ -801,6 +833,11 @@ def button_view_child_locations(self):
             "context": {"default_location_id": self.id},
         }
 
+    def update_scanning_date_start(self):
+        """Method to be override in other modules"""
+        self.ensure_one()
+        return None
```
kenvac, line 839: Question: Do we really need to return?

kenvac (changes requested):
Few minor comments

addons/udes_stock/models/stock_inventory.py
armandcela: This field is a copy paste from udes_stock_count and commented on udes_stock_count :).
Happy to make the change though :)

armandcela: As i said previously its a copy paste from udes_stock_count

armandcela: Last chance: Do i need to change it as in the same way is used even in another help :)?

addons/udes_stock/models/stock_location.py
armandcela: There are some dummy methods which are merged that return None. Me personally, prefer to return smth

addons/udes_stock/models/stock_inventory.py
kenvac: Keep it consistent

addons/udes_stock/models/stock_location.py
kenvac: We can remove the return None statement but it can't leave it

kenvac (approved):

addons/udes_stock/models/stock_inventory.py
```diff
@@ -38,6 +49,14 @@ class StockInventory(models.Model):
         "Start Date Scanning",
         help="The date operator started scanning."
     )
+    user_id = fields.Many2one(
+        "res.users",
+        string="Checked Owner",
+        readonly=True,
+        states={"draft": [("readonly", False)], "waiting": [("readonly", False)]},
+        help="Specify Owner to focus your stock check on a particular user.",
+        domain=_get_stock_check_users,
```
PeterC10, line 58: Q: Does this method get called each time we check for a user, or only when the Odoo process starts?

For example if Bob isn't a stock user, but then I make him one through the UI, will he appear in this dropdown without needing to restart Odoo?

Probably fine just something I want to double check because I don't use a method on domain that often. :)

addons/udes_stock/models/stock_inventory.py
```diff
@@ -260,6 +279,14 @@ def write(self, values):
             raise UserError(_("Cannot write to an adjustment which has already been validated"))
         return super(StockInventory, self).write(values)
 
+    def get_stock_check_event_states(self):
+        """
+        States that trigger creating events if advanced stock check is not installed
```
PeterC10, line 284: We should make this a bit more generic.

addons/udes_stock/models/stock_location.py
```diff
@@ -597,6 +606,19 @@ def get_inventory_adjustment_data(self):
             "location_id": self.id,
             "filter": "none",
             "state": "confirm",
+            "user_id": self.env.user.id,
+            "u_date_start_scanned": self.u_date_start_scanned,
+        }
+
+    def get_inventory_validated_data(self):
```
ws-mps, line 613: This function is the same as the previous one besides the state, can we have a common function to get default data?
Otherwise, if we want to change the name  (`"PI inventory validated " + self.name,`) to something else we will have to change it twice or we might forget one of them.

addons/udes_stock/models/stock_location.py
```diff
@@ -703,6 +725,15 @@ def _get_stock_drift(self, adjustments_request):
...
         return stock_drift
 
+    def _process_inventory_check_validation(self):
+        """Method will be override if advanced stock check is installed"""
+        Inventory = self.env["stock.inventory"]
+
+        inventory_data = self.get_inventory_validated_data()
+        inventory = Inventory.create(inventory_data)
+        inventory.action_start()
+        inventory.sudo().button_done()
```
ws-mps, line 735: Any reason we use button_done() instead of action_done()? Button done can open a pop-up, which we probably don't want in this case.

addons/udes_stock/models/stock_location.py
armandcela: Yes i was aware tbh :), just thought it would look like more complex to send name as parameter :(

ws-mps (approved):

PeterC10 (approved):
LGTM

PR #508, opened 2021-09-22
[IMP] udes_stock Adds location size and update date. (story/15054)

BlommaertsEdwin:


armandcela (approved):

armandcela (approved):

addons/udes_stock/models/stock_location.py
```diff
@@ -202,6 +202,28 @@ def _compute_storage_format(self):
...
+    
+    u_size = fields.Float(
+        string="Location size",
+        help="""
+            The size of the location in square feet.
+        """,
+        digit=2,
+    )
+
+    u_size_write_date = fields.Datetime(
```
ws-mps, line 214: Since this is not only about size changed, can we find a better name?

BlommaertsEdwin: Changed the name to be more specific.

PR #510, opened 2021-09-29
[15497][FIX] udes_stock: Super calling wrong class

kenvac:


addons/udes_stock/models/stock_quant.py
```diff
@@ -16,7 +16,7 @@ def _get_removal_strategy_order(self, removal_strategy=None):
         # NOTE: Preserve `NULLS FIRST` for `in_date` and use for `package_id` as well
         if removal_strategy == "fifo":
             return "in_date ASC NULLS FIRST, package_id ASC NULLS FIRST"
-        return super(Quant, self)._get_removal_strategy_order(removal_strategy)
+        return super(StockQuant, self)._get_removal_strategy_order(removal_strategy)
```
peterpratt13, line 19: Can we just use super()?

peterpratt13:
lol, that was me. We must only ever use fifo...

armandcela (approved):

kdwyer:
Do we need  a unit test?  Evidently there isn't one for this method.

peterpratt13:
> Do we need a unit test? Evidently there isn't one for this method.

Good point, add one to here
https://github.com/unipartdigital/udes-open/commit/c360bacb62442a91445f83930a6ed4761ff4cc46

Only checks the Fifo method

PeterAlabaster (approved):

ws-mps (approved):

PR #514, opened 2021-10-08
Story/15280

BlommaertsEdwin:


peterpratt13 (approved):

adamrpatrick (approved):

addons/edi_notifier/tests/test_edi_notifier.py
```diff
@@ -143,10 +144,15 @@ def setUpClass(cls):
                 "doc_type_ids": [(6, False, cls.doc_type.ids)],
                 "template_id": cls.email_template.id,
                 "active": True,
+                "safety": "edi.disable_edi_notifications",
             }
         )
 
-    def test_send_mail_on_success(self):
+    @mock.patch(
```
kdwyer, line 151: We could apply `@mock.patch` as a class decorator, which would reduce the noise in the code (assuming that applying it to _all_ the test methods won't have any adverse side effects).   It's a matter of taste though, I think.

ws-mps: We might need another test for the change or change the mock, because with the current mock we return true for check_edi_notifications_enabled but we don't have a test that test the function itself. Could we mock `config.get_misc(section or self._name, key)` instead?

addons/edi_notifier/README.md
```diff
@@ -38,3 +38,16 @@ Checks if a transfer with doc type exists between the start of the day and the t
 
 Checks if a transfer with doc type exists within the number of `lookback_hours` to
 the trigger `time` of the crons attached and that it has not already been reported on
+
+# Safety Catch
+
+In order to allow the edi notifier to send out emails:
+- A safety catch needs to be set on the edi notifier
```
ws-mps, line 45: Do we want . at the end of the sentences? We have it in other READMEs

addons/edi_notifier/tests/test_edi_notifier.py
PeterC10: So I looked into this a bit more last night. We could apply the `@mock.patch' at the class level but actually there's an even less noisier way we can do it.

I have removed all the mocks, and now set the safety catch into the config in `EdiNotifierCase`. This has the added benefit of being sure that `check_edi_notifications_enabled` does pick up config changes, rather then mocking it to just return what we want for the test. So I think avoids the need for a test on that method.

`TestDisabledNotifier` also ensures that it isn't being sent when the safety isn't set.

Let me know what you think.

kdwyer: That's much cleaner :+1:

kdwyer (approved):

PR #516, opened 2021-10-11
[ADD] udes_security: Add View Only Desktop User setting for Users

PeterC10:
If set, the user cannot create/edit/delete any records in the desktop UI.

Story/15330


addons/udes_security/models/models.py
```diff
@@ -0,0 +1,46 @@
+import simplejson
```
kenvac: This is not used anywhere as far as I can see

addons/udes_security/models/ir_actions.py
```diff
@@ -0,0 +1,16 @@
...
+class IrActions(models.Model):
+    _inherit = "ir.actions.actions"
+
+    @api.model
+    @tools.ormcache("frozenset(self.env.user.groups_id.ids)", "model_name")
+    def get_bindings(self, model_name):
+        """Override to not add any actions if the user is set to view only on desktop"""
+        if self.env.user.u_desktop_readonly:
+            return defaultdict(list)
+        else:
```
kenvac, line 15: Do we really need 'else' there?

addons/udes_security/models/ir_actions.py
```diff
@@ -0,0 +1,16 @@
+from collections import defaultdict
+
+from odoo import api, models, tools
+
+
+class IrActions(models.Model):
+    _inherit = "ir.actions.actions"
+
+    @api.model
+    @tools.ormcache("frozenset(self.env.user.groups_id.ids)", "model_name")
```
kenvac, line 10: This could cause issue. It may not get updated when user permissions are changed from frontend.
So test the following scenario
* Start the server
* Change the permissions
* Test the permissions
* Change the permission and then test the permissions

PeterC10: This was taken from the original get_bindings method in `base/ir/ir_actions.py` so kept it in for safety, can try with/without to see if it makes a difference though. This permission is a boolean flag on the user so should be fine either way but will look into it. :)

PeterC10: All good with and without this being set, but since I'm not using `groups_id` there is no point in keeping it, so have removed.

addons/udes_security/models/models.py
```diff
@@ -0,0 +1,69 @@
...
+                # Make all fields readonly
+                # Readonly user shouldn't be able to get a form into edit mode, but it is possible
+                # via the URL to bring up new record screen and quick create for relational fields
+                _set_fields_readonly(doc, res)
+
+            elif doc.tag == "tree":
+                _remove_buttons(doc)
+                _set_fields_readonly(doc, res)
+                # Sequence handle allows records to be re-ordered
+                for node in doc.xpath("//field[@widget='handle']"):
```
ws-mps, line 50: We could generalise `_remove_buttons(doc)` into `_remove_nodes(doc, xpath)` and reuse it for this loop and the next one.

PeterC10: Yep good idea :) 

I kept `_remove_buttons` (needed also to remove from Kanban views, as buttons can lead to changes, e.g. Install on the module cards...) but this just makes use of `_remove_nodes`. :)

PR #522, opened 2021-10-29
[FIX] udes-open: restrict ability to add picking types to priorities

LoKKeR96:
When adding a new stock transfer type to a priority if there was a
picking that is not yet completed then UDES would complain of this
and not let any user modify the list of assigned stock transfer types.

This change adds the ability for sysadmin, debug and trusted users
to modify the list of stock transfer types for priorities and also
see the outstanding pickings for the priority in question.

User-story: 20

### DIFF
--- a/addons/udes_priorities/models/priority.py
+++ b/addons/udes_priorities/models/priority.py
@@ -1,15 +1,19 @@
 # -*- coding: utf-8 -*-
 
-from odoo import models, fields, api, _
+from odoo import models, fields, api, SUPERUSER_ID, _
 from odoo.exceptions import UserError
 from uuid import uuid4
 
+
+""" Cannot be edited by anyone, including admin"""
+FORBIDDEN_FIELDS = {"reference"}
+
+""" Cannot be edited by anyone else than trusted or admin"""
 PROTECTED_FIELDS = {
     "name",
-    "reference",
     "sequence",
     "picking_type_ids",
-    "group_ids",
+    "priority_group_ids",
     "active",
 }
 
@@ -26,16 +30,37 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _compute_pickings_count(self):
+        """
+        Count the number of transfers assigned to each priority.
+        """
+        for priority in self:
+            Picking = self.env["stock.picking"]
+            priority.picking_count = Picking.search_count(
+                [
+                    ("priority", "=", priority.reference),
+                ]
+            )
+
     def _default_sequence(self):
         max_rec = self.search([], order="sequence desc, id desc", limit=1)
         return max_rec.sequence + 1 if max_rec else 1
 
-    sequence = fields.Integer(default=_default_sequence, help="Used to order the priorities")
+    sequence = fields.Integer(
+        default=_default_sequence, help="Used to order the priorities"
+    )
     description = fields.Text(help="A description about the purpose of the priority")
     picking_type_ids = fields.Many2many(
         "stock.picking.type",
         string="Stock Transfer Types",
-        help="The stock picking types this priority is visible on. No picking types means all picking types",
+        help="The stock transfer types this priority is visible on. No transfer types means all transfer types",
+    )
+    picking_count = fields.Integer(
+        string="Affected Stock Transfers",
+        compute="_compute_pickings_count",
+        help="Stock transfers that would be affected by changes in the current priority.",
     )
     priority_group_ids = fields.Many2many(
         "udes_priorities.priority_group",
@@ -84,45 +109,86 @@ class UdesPriority(models.Model):
     @api.one
     def _check_for_duplicate_names_for_same_pick_type(self):
         if self._has_duplicate_names_for_same_pick_type() > 0:
-            raise UserError("Name must be unique for the picking types it is shown on")
+            raise UserError("Name must be unique for the transfer types it is shown on")
 
-    @api.multi
-    def _number_of_oustanding_pickings(self):
-        Picking = self.env["stock.picking"]
-        self.ensure_one()
-        return Picking.search_count(
-            [("priority", "=", self.reference), ("state", "not in", ("cancel", "done")),]
+    @api.constrains("active")
+    def _check_existing_transfers(self):
+        """Constrain active field
+
+        Allow only priorities with no existing stock transfers set to this priority to be archived.
+        """
+        inactive_priorities = self.filtered(lambda p: not p.active)
+        for priority in inactive_priorities:
+            if priority.picking_count:
+                raise UserError(
+                    "Cannot archive a priority that is set on existing transfers."
+                )
+
+    def _is_user_trusted_or_admin(self):
+        return self.env.uid == SUPERUSER_ID or self.env.user.has_group(
+            "udes_security.group_trusted_user"
         )
 
+    def _is_install_mode(self):
+        """Return true if Odoo is installing the module"""
+        return self.env.context.get("install_mode", False)
+
     @api.multi
-    def _check_for_oustanding_pickings(self):
+    def _validate_transfer_types_changes(self, values):
+        """Validate changes to transfer types
 
-        install_mode = self.env.context.get("install_mode", False)
-
-        if install_mode:
+        Allow addition of any transfer type but no removal of transfer types when
+        there are existing transfers of this type with the current priority set.
+        """
+        if self._is_install_mode():
             return
 
-        priorities_with_outstanding_picks = self.browse()
-        for priority in self:
-            if priority._number_of_oustanding_pickings() > 0:
-                priorities_with_outstanding_picks |= priority
-
-        if priorities_with_outstanding_picks:
-            raise UserError(
-                _("There are outstanding pickings for priorities: {}").format(
-                    ", ".join(priorities_with_outstanding_picks.mapped("name"))
-                )
+        if "picking_type_ids" in values:
+            Picking = self.env["stock.picking"]
+            message = (
+                "Invalid changes, you cannot remove the following transfer types as there are existing"
+                " transfers from this type assigned to the current priority:\n"
             )
+            invalid = False
+            new_ids = values["picking_type_ids"][0][2]
+
+            for old_pick_type in self.picking_type_ids:
+                if old_pick_type.id not in new_ids:
+                    assigned_pickings = Picking.search_count(
+                        [
+                            ("priority", "=", self.reference),
+                            ("picking_type_id", "=", old_pick_type.id),
+                        ]
+                    )
+                    if assigned_pickings:
+                        invalid = True
+                        message += "\n" + old_pick_type.name
+            if invalid:
+                raise UserError(message)
 
     @api.multi
     def write(self, values):
-        if values and any(k in PROTECTED_FIELDS for k in values):
-            self._check_for_oustanding_pickings()
+        if any(k in FORBIDDEN_FIELDS for k in values):
+            if self._is_install_mode():
+                return
+            raise UserError("Cannot change protected fields.")
+
+        if self._is_user_trusted_or_admin():
+            self._validate_transfer_types_changes(values)
+        else:
+            if any(k in PROTECTED_FIELDS for k in values):
+                raise UserError("You do not have the rights for making changes.")
         return super().write(values)
 
     @api.multi
     def unlink(self):
-        self._check_for_oustanding_pickings()
+        if self._is_user_trusted_or_admin() is False:
+            raise UserError("You do not have the rights for making changes.")
+        for priority in self:
+            if priority.picking_count:
+                raise UserError(
+                    "Changes cannot be applied as there are transfers assigned to selected priorities."
+                )
         return super().unlink()
 
     @api.multi
--- a/addons/udes_priorities/tests/common.py
+++ b/addons/udes_priorities/tests/common.py
@@ -9,7 +9,9 @@ class BasePriorities(common.BaseUDES):
         super().setUpClass()
 
         Priorities = cls.env["udes_priorities.priority"]
-        (Priorities.search([]) - cls.env.ref("udes_priorities.normal")).write({"active": False})
+        (Priorities.search([]) - cls.env.ref("udes_priorities.normal")).write(
+            {"active": False}
+        )
 
         cls.urgent = Priorities.create(
             {
@@ -39,3 +41,28 @@ class BasePriorities(common.BaseUDES):
         )
 
         cls.test_priorities = cls.urgent | cls.normal | cls.not_urgent
+
+        User = cls.env["res.users"]
+
+        cls.group_stock_user = cls.env.ref(
+            "stock.group_stock_user"
+        )  # Necessary to access stock.picking
+        cls.group_trusted_user = cls.env.ref("udes_security.group_trusted_user")
+
+        cls.trusted_usr = User.create(
+            {
+                "name": "test_priority_usr_1",
+                "login": "test_priority_usr_1",
+                "groups_id": [
+                    (6, 0, [cls.group_stock_user.id, cls.group_trusted_user.id])
+                ],
+            }
+        )
+
+        cls.simple_stock_usr = User.create(
+            {
+                "name": "test_priority_usr_2",
+                "login": "test_priority_usr_2",
+                "groups_id": [(6, 0, [cls.group_stock_user.id])],
+            }
+        )
--- a/addons/udes_priorities/views/priorities.xml
+++ b/addons/udes_priorities/views/priorities.xml
@@ -37,6 +37,7 @@
                 <field name="reference"/>
                 <field name="description"/>
                 <field name="picking_type_ids"/>
+                <field name="picking_count"/>
                 <field name="priority_group_ids"/>
               </group>
             </group>
--- a/addons/udes_priorities/__manifest__.py
+++ b/addons/udes_priorities/__manifest__.py
@@ -7,7 +7,7 @@
     "website": "http://unipart.io",
     "category": "Warehouse",
     "version": "0.1",
-    "depends": ["udes_stock"],
+    "depends": ["udes_stock", "udes_security"],
     "data": [
         "data/default_priority_groups.xml",
         "data/default_priorities.xml",
--- a/addons/udes_priorities/tests/test_priority.py
+++ b/addons/udes_priorities/tests/test_priority.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
-
 from . import common
-from odoo.exceptions import UserError
+from odoo.exceptions import AccessError, UserError, ValidationError
 
 
 class TestPriorities(common.BasePriorities):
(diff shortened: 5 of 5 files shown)


addons/udes_priorities/models/priority.py
```diff
@@ -107,7 +121,11 @@ def _check_for_oustanding_pickings(self):
             if priority._number_of_oustanding_pickings() > 0:
                 priorities_with_outstanding_picks |= priority
 
-        if priorities_with_outstanding_picks:
+        privileged_user = self.env.uid == SUPERUSER_ID or \
+            self.env.user.has_group("udes_security.group_debug_user") or \
+            self.env.user.has_group('udes_security.group_trusted_user')
+
+        if priorities_with_outstanding_picks and not privileged_user:
```
PeterC10, line 128: Is `priorities_with_outstanding_picks` the same as `affected_pickings`?

addons/udes_priorities/tests/common.py
```diff
@@ -39,3 +39,21 @@ def setUpClass(cls):
...
+            "login": "test_priority_usr_1",
+            "groups_id": [(6, 0, [cls.group_stock_user.id,
+                                  cls.group_trusted_user.id])],
+        })
+
+        cls.simple_stock_usr = User.create({
+            "name": "test_priority_usr_2",
+            "login": "test_priority_usr_2",
+            "groups_id": [(6, 0, [cls.group_stock_user.id])],
+        })
```
PeterC10, line 59: Nit: Missing newline at the end of the file.

addons/udes_priorities/views/priorities.xml
```diff
@@ -37,6 +37,7 @@
                 <field name="reference"/>
                 <field name="description"/>
                 <field name="picking_type_ids"/>
+                <field name="affected_pickings"/>
```
PeterC10, line 40: Adding this to the view means we have to compute the field each time a priority is opened, which depending on the number of pickings could be quite expensive.

Maybe have a button instead to redirect to a view of a list of pickings that match the domain used in the compute field?

Might be worth getting a second opinion on this. :)

addons/udes_priorities/models/priority.py
```diff
@@ -37,6 +45,12 @@ def _default_sequence(self):
         string="Stock Transfer Types",
         help="The stock picking types this priority is visible on. No picking types means all picking types",
     )
+    affected_pickings = fields.Many2many(
+        "stock.picking",
+        string="Affected Stock Transfers",
+        compute="_get_default_affected_pickings",
```
ws-mps, line 51: Suggest to rename the compute method to '_compute_affected_pickings`, that is the usual convention for the methods of computed fields.

addons/udes_priorities/models/priority.py
```diff
@@ -26,6 +26,14 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _get_default_affected_pickings(self):
+        Picking = self.env["stock.picking"]
+        self.affected_pickings = Picking.search(
+            [("priority", "=", self.reference), ("state", "not in", ("cancel", "done")), ]
```
PeterC10, line 34: Nit: Extra space before the closing square bracket, has the file been formatted with black?

LoKKeR96: No, I have switched to a different IDE (PyCharm) and haven't set up black formatting yet. Will definitely spend some time making it the default formatting of my IDE to avoid this.

addons/udes_priorities/models/priority.py
LoKKeR96: I haven't given too much thought on this because it's simply avoiding users without access rights to modify priorities but the priorities edit form cannot be seen by anyone else apart from trusted users. Since we are now allowing trusted users to edit priorities completely we could avoid entirely checking whether there are outstanding priorities since we are showing these on the form view. We could simply check for user privileges as a validation step. Does that sound clear?

addons/udes_priorities/models/priority.py
ws-mps, line 34: We are using the same domain in `_number_of_oustanding_pickings` and they should be aligned. Would be good to have a function that generates this domain and use it in both methods.

addons/udes_priorities/models/priority.py
ws-mps: Just realised there is a typo in `_number_of_oustanding_pickings`, we might want to fix this too :)

addons/udes_priorities/models/priority.py
```diff
@@ -37,6 +45,12 @@ def _default_sequence(self):
         string="Stock Transfer Types",
         help="The stock picking types this priority is visible on. No picking types means all picking types",
     )
+    affected_pickings = fields.Many2many(
```
ws-mps, line 48: We might want this field as readonly.

addons/udes_priorities/models/priority.py
PeterC10: We could also call this method for the domain of the picking view if we do the change suggested in: https://github.com/unipartdigital/udes-open/pull/522#discussion_r741020847

armandcela: I think to remove at all _number_of_oustanding_pickings method, and to change with the field affected_pickings.
When first was done , it was used i think search_count for performance but now that we are adding affected_pickings we dont need at all _number_of_oustanding_pickings. Even if the number is needed can use length but as far as i saw is not needed at the moment, just True or False we need

addons/udes_priorities/views/priorities.xml
PeterC10: So I think something like this should work (not tested at all, and just a suggestion so you can change as needed of course):

Add this to the top of the priority view:

```
<div class="oe_button_box" name="button_box">
    <button string="Outstanding Transfers"
        class="oe_stat_button"
        icon="fa-building-o"
        name="button_view_outstanding_pickings"
        type="object"
</div>
```

Add the method to the priority model:

```
def button_view_outstanding_pickings(self):
    self.ensure_one()
    return {
        "name": _("%s Outstanding Pickings") % self.name,
        "type": "ir.actions.act_window",
        "view_type": "tree",
        "view_mode": "tree",
        "res_model": "stock.picking",
        "target": "current",
        "domain": self._get_outstanding_pickings_domain, # use the method with the domain set, as per Miquel's suggestion
    }
```

addons/udes_priorities/models/priority.py
ws-mps: It is fine as it is, we only have to consolidate the compute/usage of number of outstanding pickings / affected pickings.

addons/udes_priorities/views/priorities.xml
PeterC10: Resolving this as not needed as talking with Miquel.

If a user is told there are outstanding pickings then they can go to the picking view and find the pickings themselves.

So we don't need the list of pickings and can remove that many2many field.

addons/udes_priorities/models/priority.py
PeterC10, line 48: We can get rid of this field. It's only important for the user to know if there are any outstanding pickings when they try to change a priority. If they want to find the actual pickings then they can manually search themselves in the picking view.

addons/udes_priorities/models/priority.py
armandcela: I totally agree :)

addons/udes_priorities/models/priority.py
```diff
@@ -26,17 +26,35 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _compute_outstanding_pickings(self):
+        Picking = self.env["stock.picking"]
+        self.outstanding_pickings = Picking.search_count(
```
ws-mps, line 33: We might still want to keep `self.ensure_one()`.

addons/udes_priorities/models/priority.py
```diff
@@ -87,42 +105,33 @@ def _check_for_duplicate_names_for_same_pick_type(self):
...
-            if priority._number_of_oustanding_pickings() > 0:
-                priorities_with_outstanding_picks |= priority
+        privileged_user = (
+            self.env.uid == SUPERUSER_ID
+            or self.env.user.has_group("udes_security.group_debug_user")
+            or self.env.user.has_group("udes_security.group_trusted_user")
+        )
 
-        if priorities_with_outstanding_picks:
+        if self.affected_pickings > 0 and not privileged_user:
```
ws-mps, line 121: Typo: outstanding_pickings instead of affected_pickings

ws-mps: I think we still need `priorities_with_outstanding_picks`, since `_check_for_outstanding_pickings()` is called from write, which can contain more than one record in self. Although we can simplify the computation to something like:
```priorities_with_outstanding_picks = self.filtered(lambda p: p.outstanding_pickings > 0```

Comment (LoKKeR96) [reply addons/udes_priorities/models/priority.py:121]: I substituted "priorities_with_outstanding_picks" with "self.affected_pickings > 0" which is basically still checking for outstanding pickings but instead of checking all priorities it only checks the current one that is being visualised/modified which to me makes more sense since the changes to a priority would only affect pickings with that priority assigned. Is there a reason why we want to check all pickings regardless of the priority? Just trying to be specific and clear about the goal.

File: addons/udes_priorities/models/priority.py (hunk 7 above)
Comment (LoKKeR96) [reply addons/udes_priorities/models/priority.py:33]: I am still learning about when it is appropriate to use `ensure.one()` and because of that I'd like to ask, what would be the purpose of having this called here in a compute method? In which occasion would this be a problem? I assume that if the method is only used by a compute field then there is no need for calling `ensure_one()`?

Comment (ws-mps) [reply addons/udes_priorities/models/priority.py:33]: In this case it might not be needed, but in general @api.multi means we can have more than one record in self, which means if we do self.outstanding_pickings = XXX (when self contains more than one record) it will fail, instead we should use self.write(YYY). Usually we add a loop just in case:
```
for record in self:
      record.outstanding_pickings = XXX
```
At the end _compute_outstanding_pickings() is a function and can be called somewhere else (it shouldn't though).

File: addons/udes_priorities/models/priority.py (hunk 8 above)
Comment (ws-mps) [reply addons/udes_priorities/models/priority.py:121]: This function is called from write() which is used to update multiple records, in that case self could contain more than one record and sefl.XXXX will fail for the same reason I have explained above.
Although we are updating only one priority using the UI we could have some code that calls write() and will trigger the error.

File: addons/udes_priorities/models/priority.py [hunk 9]
```diff
@@ -26,17 +26,36 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _compute_outstanding_pickings(self):
+        self.ensure_one()
+        Picking = self.env["stock.picking"]
```
ws-mps, line 33: By convention we put this at the start of the method.

addons/udes_priorities/models/priority.py
```diff
@@ -87,42 +106,37 @@ def _check_for_duplicate_names_for_same_pick_type(self):
...
-        for priority in self:
-            if priority._number_of_oustanding_pickings() > 0:
-                priorities_with_outstanding_picks |= priority
+        privileged_user = (
+            self.env.uid == SUPERUSER_ID
+            or self.env.user.has_group("udes_security.group_debug_user")
+            or self.env.user.has_group("udes_security.group_trusted_user")
+        )
+
+        priorities_with_outstanding_picks = len(
```
ws-mps, line 122: len is not needed, empty recordset evaluates to false, and non empty recordset evaluates to true. So we don't need the > 0 in line 126.

addons/udes_priorities/models/priority.py
```diff
@@ -26,17 +26,36 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _compute_outstanding_pickings(self):
+        Picking = self.env["stock.picking"]
+        for record in self:
```
PeterC10, line 33: Nit: Sorry super super minor one. But can we call this `priority` instead of `record`? As we are looping through a recordset of priorities.

Not a big deal but we try to be as explicit as possible with the variable names. :)

addons/udes_priorities/models/priority.py
```diff
@@ -26,17 +26,36 @@ class UdesPriority(models.Model):
         default=lambda self: uuid4(),
     )
 
+    @api.multi
+    @api.depends("reference")
+    def _compute_outstanding_pickings(self):
```
PeterC10, line 31: Nit: Can we add a doc-string? Doesn't need to be super detailed, but is good practice.

addons/udes_priorities/models/priority.py
```diff
@@ -26,17 +26,36 @@ class UdesPriority(models.Model):
...
     description = fields.Text(help="A description about the purpose of the priority")
     picking_type_ids = fields.Many2many(
         "stock.picking.type",
         string="Stock Transfer Types",
         help="The stock picking types this priority is visible on. No picking types means all picking types",
     )
+    outstanding_pickings = fields.Integer(
+        string="Affected Stock Transfers",
+        compute="_compute_outstanding_pickings",
+        help="Active stock pickings that would be affected by changes in the current priority.",
```
PeterC10, line 57: Super nit: Can we change this to `Active stock transfers?`. Otherwise it's a bit inconsistent as the string says transfers but the help says pickings.

addons/udes_priorities/models/priority.py
```diff
@@ -26,17 +26,36 @@ class UdesPriority(models.Model):
...
+    sequence = fields.Integer(
+        default=_default_sequence, help="Used to order the priorities"
+    )
     description = fields.Text(help="A description about the purpose of the priority")
     picking_type_ids = fields.Many2many(
         "stock.picking.type",
         string="Stock Transfer Types",
         help="The stock picking types this priority is visible on. No picking types means all picking types",
     )
+    outstanding_pickings = fields.Integer(
```
PeterC10, line 54: Suggestion (feel free to ignore or get a 2nd opinion): To me the field name `outstanding_pickings` isn't clear as to whether it is a recordset of pickings or a count.

Suggestion may `num_outstanding_pickings` or `outstanding_picking_count` or similar.

addons/udes_priorities/models/priority.py
LoKKeR96: I was considering doing that but then I noticed on `picking_type_ids` it mentions both pickings and transfer terms so I copied the style over. Probably need to fix the one above too to be consistent.

PeterC10 (approved):
LGTM

addons/udes_priorities/models/priority.py
```diff
@@ -84,45 +107,40 @@ def _has_duplicate_names_for_same_pick_type(self):
...
             return
 
-        priorities_with_outstanding_picks = self.browse()
-        for priority in self:
-            if priority._number_of_oustanding_pickings() > 0:
-                priorities_with_outstanding_picks |= priority
+        privileged_user = (
+            self.env.uid == SUPERUSER_ID
+            or self.env.user.has_group("udes_security.group_debug_user")
+            or self.env.user.has_group("udes_security.group_trusted_user")
```
adamrpatrick, line 123: Now that we're using some user groups from the udes_security module, we'll need to add the udes_security module as a dependency in the manifest.

PeterC10: Don't mind adding it explicitly in this module, but we have a dependency on `udes_security` in `udes_stock`, which is a dependency for this one. :)

ws-mps: I like explicit dependencies when ids are used.

adamrpatrick (approved):

williamsajr (approved):

addons/udes_priorities/models/priority.py
```diff
@@ -4,12 +4,14 @@
 from odoo.exceptions import UserError
 from uuid import uuid4
 
+
+MAIN_PROTECTED_FIELDS = {"reference"}
```
PeterC10, line 8: `MAIN_PROTECTED_FIELDS` is a bit confusing. I can't think of a nicer way to link it with `PROTECTED_FIELDS` though...

addons/udes_priorities/models/priority.py
```diff
@@ -84,45 +105,86 @@ def _has_duplicate_names_for_same_pick_type(self):
...
-            [("priority", "=", self.reference), ("state", "not in", ("cancel", "done")),]
+    def _is_trusted_user(self):
+        return self.env.uid == SUPERUSER_ID or self.env.user.has_group(
+            "udes_security.group_trusted_user"
         )
 
-    @api.multi
-    def _check_for_oustanding_pickings(self):
+    def _get_outstanding_pickings(self, reference_id, pick_type_id=None):
+        Picking = self.env["stock.picking"]
```
armandcela, line 116: At the import `from odoo.osv import expression`
`search_domain = [("priority", "=", reference_id),("state", "not in", ("cancel", "done"))]`
`if pick_type_id:
    search_domain = expression.AND([search_domain, [("picking_type_id", "=", pick_type_id)]])`
`return Picking.search_count(search_domain)`

addons/udes_priorities/models/priority.py
LoKKeR96: Agreed, maybe we could call this one BASIC_PROTECTED_FIELDS and the other one EXTENDED_PROTECTED_FIELDS?

addons/udes_priorities/models/priority.py
```diff
@@ -84,45 +105,86 @@ def _has_duplicate_names_for_same_pick_type(self):
...
+                raise UserError("You do not have the rights for making changes.")
         return super().write(values)
 
     @api.multi
     def unlink(self):
-        self._check_for_oustanding_pickings()
+        if self._is_trusted_user() is False:
+            raise UserError("You do not have the rights for making changes.")
+        for priority in self:
+            if priority.outstanding_picking_count > 0:
```
armandcela, line 184: Dont need > 0 i think should be enough:
`if priority.outstanding_picking_count:`

armandcela (changes requested):

addons/udes_priorities/models/priority.py
```diff
@@ -84,45 +105,86 @@ def _has_duplicate_names_for_same_pick_type(self):
...
+                "Invalid changes, you cannot remove the following transfer types as there are outstanding "
+                "transfers\n "
             )
+            invalid = False
+            old_ids = self.picking_type_ids.ids
+            new_ids = values["picking_type_ids"][0][2]
+
+            for old_pick_type_id in old_ids:
+                if old_pick_type_id not in new_ids:
+                    outstanding_pickings = self._get_outstanding_pickings(
```
PeterC10, line 153: We shouldn't be able to remove a priority if it has been set on any previous picking, not just outstanding pickings.

addons/udes_priorities/models/priority.py
```diff
@@ -84,45 +105,86 @@ def _has_duplicate_names_for_same_pick_type(self):
...
+                        self.reference, pick_type_id=old_pick_type_id
+                    )
+                    if outstanding_pickings:
+                        invalid = True
+                        picking_type = PickingType.search(
+                            [("id", "=", old_pick_type_id)]
```
adamrpatrick, line 159: If we know the `id` of the picking type we can just do `PickingType.browse(old_pick_type_id)` instead of `.search`

PeterC10:
Not sure if this was mentioned earlier but the commit should start `[FIX] udes_priorities` i.e. the module, not the repo.

addons/udes_priorities/models/priority.py
```diff
@@ -1,15 +1,17 @@
...
+MAIN_PROTECTED_FIELDS = {"reference"}
+
 PROTECTED_FIELDS = {
     "name",
-    "reference",
     "sequence",
     "picking_type_ids",
-    "group_ids",
+    "priority_group_ids",
     "active",
```
adamrpatrick, line 15: I think we should only have `name` and `picking_type_ids` in `PROTECTED_FIELDS` and the rest should be in `MAIN_PROTECTED_FIELDS`

addons/udes_priorities/models/priority.py
LoKKeR96: Thanks! I was wondering how this was done in a simpler way.

addons/udes_priorities/models/priority.py
PeterC10: Yeah good point, if we make a priority inactive I can see that breaking a lot of things... but if it is protected then why even have an active field?

Maybe we need a constraint on active field, that it can't be set inactive if it is used on other pickings. Otherwise we'll risk getting an error when accessing a picking that is using an inactive priority.

addons/udes_priorities/models/priority.py
PeterC10: After much debate...

`MAIN_PROTECTED_FIELDS` -> `FORBIDDEN_FIELDS` and leave `PROTECTED_FIELDS` as they are.

I'd suggest adding a comment above each one to be clear that `FORBIDDEN_FIELDS` cannot be edited by anyone (not even admin), and `PROTECTED_FIELDS` can only be edited by trusted users.

addons/udes_priorities/models/priority.py
armandcela: I agree with @adamrpatrick , but believe it can be improved at the beginning of for loop.
Loop through records not through ids, and you dont have even to browse if you loop through recrods

armandcela (approved):

addons/udes_priorities/models/priority.py
LoKKeR96: Done.

addons/udes_priorities/models/priority.py
LoKKeR96: Done.

addons/udes_priorities/models/priority.py
LoKKeR96: After some discussions, agreed and done in latest commit.

PeterC10 (approved):
LGTM :)

PR #523, opened 2021-11-02
[ADD] udes_stock: Add tool to check for multi-location packages/pallets

williamsajr:
User-story: x12

### DIFF
--- a/addons/udes_stock/tools/find_multi_location_packages_and_pallets.sql
+++ b/addons/udes_stock/tools/find_multi_location_packages_and_pallets.sql
@@ -1,0 +1,46 @@
+\timing
+
+-- Find packages/pallets of products in multiple locations
+SELECT
+    p.package_id,
+    COUNT(location_id)
+FROM (
+    SELECT
+        package_id,
+        location_id
+    FROM
+        stock_quant
+    WHERE
+        package_id IS NOT NULL
+    GROUP BY
+        package_id,
+        location_id
+) AS p
+GROUP BY
+    p.package_id
+HAVING
+    COUNT(location_id) > 1
+;
+
+-- Find pallets of packages in mutliple locations
+SELECT
+    p.pallet_id,
+    COUNT(location_id)
+FROM (
+    SELECT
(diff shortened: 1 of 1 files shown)


addons/udes_stock/tools/find_multi_location_packages_and_pallets.sql
```diff
@@ -0,0 +1,44 @@
...
+
+-- Find packages in multiple locations
+SELECT
+    p.package_id,
+    COUNT(location_id)
+FROM (
+    SELECT
+        package_id,
+        location_id
+    FROM
```
ws-mps, line 11: We can add a WHERE here to filter out quants without package_id, so we won't need it in the having.

williamsajr: Changed

addons/udes_stock/tools/find_multi_location_packages_and_pallets.sql
```diff
@@ -0,0 +1,46 @@
...
+    LEFT JOIN
+        stock_quant_package pallet ON pallet.id = pack.package_id
+    GROUP BY
+        pallet.id,
+        q.location_id
+) AS p
+GROUP BY
+    p.pallet_id
+HAVING
+    p.pallet_id IS NOT NULL AND COUNT(location_id) > 1
```
peterpratt13, line 45: I guess miquel wanted this in the where as well?

peterpratt13: actually can pallet.id be null?

williamsajr: I've moved the where inside the select

williamsajr: Changed my mind and used the inner join. Good call

ws-mps (approved):

peterpratt13 (approved):

PR #527, opened 2021-11-15
[ADD] udes_stock: Removing logic dependencies with couple of modules

armandcela:
on specific modules

Making Workflows module more generic
Story: 4

### DIFF
--- a/addons/udes_delivery_control/models/stock_picking.py
+++ b/addons/udes_delivery_control/models/stock_picking.py
@@ -1,4 +1,5 @@
 from odoo import fields, models, api
+from odoo.addons.udes_stock import utils
 
 EXTRA_FIELDS = {
     "u_user_id",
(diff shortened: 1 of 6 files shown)


addons/udes_delivery_control/models/stock_picking.py
```diff
@@ -374,3 +374,9 @@ def create_goods_in_from_delivery_control(self, **kwargs):
             self.write({"u_goods_in_picking_id": goods_in_picking.id})
 
         return goods_in_picking
+
+    def _get_delivery_control_picking_id(self):
+        delivery_control_picking_id = super()._get_delivery_control_picking_id()
+        if not delivery_control_picking_id:
+            delivery_control_picking_id = self.u_delivery_control_picking_id
+        return delivery_control_picking_id
```
adamrpatrick, line 382: See comment on the other MR

adamrpatrick (approved):

addons/udes_stock/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+def sorted_group_by(iterable, key):
+    """Sorts and then groups an iterable by a key"""
+    return groupby(sorted(iterable, key=key), key=key)
+
+
+def product_image_urls(base_url, product):
+    """Produces a product image urls object"""
+    if product.image:
+        base_url = urljoin(base_url, "/web/image/product.product/%i" % product.id)
```
LoKKeR96, line 71: Might be good idea to import this path from BASE_PRODUCT_IMAGE_URL in "addons/udes_stock/models/product_product.py" so that it's not a magic value as in appearing from nowhere.

armandcela: Yep good suggestion, made the changes and pushed them

LoKKeR96 (approved):
Looks good to me. I cannot see anything wrong with it for what I know.

PR #528, opened 2021-11-19
[FIX] udes_stock: Don't propagate done dates to incomplete pickings

montymagic9:
When refactoring, only keep done dates for complete pickings

User-story: 19

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -468,9 +468,9 @@ class StockMove(models.Model):
         if len(partners) == 1:
             values["partner_id"] = partners.id
 
-        dates_done = list(set(moves.mapped("date")))
-        if len(dates_done) == 1:
-            values["date_done"] = dates_done[0]
+        dates_done = set(moves.mapped("date"))
+        if set(moves.mapped("state")).issubset({"done", "cancel"}):
+            values["date_done"] = max(dates_done)
 
         return values
 
--- a/addons/udes_stock/tests/test_refactoring_splitting.py
+++ b/addons/udes_stock/tests/test_refactoring_splitting.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from . import common
+import datetime
 
 
 class TestRefactoringAssignSplitting(common.BaseUDES):
@@ -379,8 +380,7 @@ class TestRefactoringAssignSplitting(common.BaseUDES):
         self.assertEqual(new_picking.move_lines.quantity_done, 3)
 
     def test10_backorder_partially_picked_mls_with_update_picking(self):
-        """Test splitting for backordering when everything is available via update_picking
-        """
+        """Test splitting for backordering when everything is available via update_picking"""
         Picking = self.env["stock.picking"]
         self.create_quant(self.elderberry.id, self.test_location_01.id, 5)
         self.create_quant(self.banana.id, self.test_location_02.id, 10)
@@ -428,9 +428,13 @@ class TestRefactoringAssignSplitting(common.BaseUDES):
         variable recompute=False"""
         Picking = self.env["stock.picking"]
         apple_pallet_01 = self.create_package()
-        self.create_quant(self.apple.id, self.test_location_01.id, 10, package_id=apple_pallet_01.id)
+        self.create_quant(
+            self.apple.id, self.test_location_01.id, 10, package_id=apple_pallet_01.id
+        )
         apple_pallet_02 = self.create_package()
-        self.create_quant(self.apple.id, self.test_location_01.id, 10, package_id=apple_pallet_02.id)
+        self.create_quant(
+            self.apple.id, self.test_location_01.id, 10, package_id=apple_pallet_02.id
+        )
 
         self.create_move(self.apple, 20, self.picking)
         self.picking.with_context(recompute=False).action_assign()
@@ -498,17 +502,24 @@ class TestRefactoringValidateSplitting(common.BaseUDES):
         # apple and banana moves are now in different pickings
         # and the original picking has been reused.
         self.assertNotEqual(apple_move.picking_id, banana_move.picking_id)
+        apple_picking = apple_move.picking_id
+        banana_picking = banana_move.picking_id
         self.assertTrue(
             self.picking == apple_move.picking_id or self.picking == banana_move.picking_id
         )
 
+        self.assertTrue(apple_picking.date_done)
+        self.assertTrue(banana_picking.date_done)
+        self.assertGreaterEqual(apple_picking.date_done, apple_move.date)
+        self.assertGreaterEqual(banana_picking.date_done, banana_move.date)
+
     def test02_maintain_single_pick_extra_info(self):
-        """ Check that when a move is split the picking's extra info is copied
-            to the new pick.
-            Extra info:
-            - origin
-            - partner_id
-            - date_done (comes from move.date when not reusing picking)
+        """Check that when a move is split the picking's extra info is copied
+        to the new pick.
+        Extra info:
+        - origin
+        - partner_id
+        - date_done (comes from move.date when not reusing picking)
         """
         apple_pallet = self.create_package()
         self.create_quant(self.apple.id, self.test_location_01.id, 10, package_id=apple_pallet.id)
@@ -564,14 +575,14 @@ class TestRefactoringValidateSplitting(common.BaseUDES):
             self.assertEqual(apple_move.picking_id.date_done, apple_move.date)
             self.assertGreaterEqual(banana_move.picking_id.date_done, banana_move.date)
 
-    def test02_maintain_two_picks_extra_info(self):
-        """ Check that when a moves from different picks are split the pickings
-            extra info is copied to the new pick and maintained when two picks
-            share the same info.
-            Extra info:
-            - origin
-            - partner_id
-            - date_done (comes from move.date)
+    def test03_maintain_two_picks_extra_info(self):
+        """Check that when a moves from different picks are split the pickings
+        extra info is copied to the new pick and maintained when two picks
+        share the same info.
+        Extra info:
+        - origin
+        - partner_id
+        - date_done (comes from move.date)
         """
 
         # Setup pick 1
@@ -672,7 +683,7 @@ class TestRefactoringConfirmSplitting(common.BaseUDES):
 
     def test01_simple(self):
         """Reserve self.picking with one pallet of each product and check it
-           splits correctly when confirmed.
+        splits correctly when confirmed.
         """
         apple_pallet = self.create_package()
         self.create_quant(self.apple.id, self.test_location_01.id, 5, package_id=apple_pallet.id)
@@ -700,8 +711,7 @@ class TestRefactoringConfirmSplitting(common.BaseUDES):
 
 class TestRefactoringAutoUnlinkEmpty(common.BaseUDES):
     def setUp(self):
-        """ Setup picking type config
-        """
+        """Setup picking type config"""
         super(TestRefactoringAutoUnlinkEmpty, self).setUp()
 
         # group by product post confirm at goods-out
@@ -718,13 +728,13 @@ class TestRefactoringAutoUnlinkEmpty(common.BaseUDES):
         return Picking.search_count([("picking_type_id", "=", self.picking_type_out.id)])
 
     def test01_auto_unlink_empty_pickings(self):
-        """ Check that unlink_empty finds any picking in the system marked as
-            empty and that when auto unlink empty is disabled for goods-out any
-            empty picking is not deleted when searching for any picking.
+        """Check that unlink_empty finds any picking in the system marked as
+        empty and that when auto unlink empty is disabled for goods-out any
+        empty picking is not deleted when searching for any picking.
 
-            Create two different picks for the same product, confirm them one
-            by one so the goods-out picking is reused leaving one empty picking
-            for the second picking.
+        Create two different picks for the same product, confirm them one
+        by one so the goods-out picking is reused leaving one empty picking
+        for the second picking.
         """
         Picking = self.env["stock.picking"]
 
@@ -762,6 +772,7 @@ class TestRefactoringAutoUnlinkEmpty(common.BaseUDES):
         Picking.unlink_empty()
         # Empty picking has disappeared
         self.assertEqual(n_out_pickings, self._count_out_pickings())
+
 
 class TestRefactoringAssignSplittingQuantity(common.BaseUDES):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -923,3 +923,75 @@ def test_split_mixed_products_partial_reserve(self):
...
+        """If we complete a pick with some lines partially completed, generated ones don't have a done date"""
+        Picking = self.env["stock.picking"]
+
+        self.pick_2.action_assign()
+        move_line_id_1, move_line_id_2 = self.pick_2.move_line_ids
+        move_line_id_1.write({"location_dest_id": self.received_location.id, "qty_done": 10})
+        move_line_id_2.write({"location_dest_id": self.received_location.id, "qty_done": 1})
+        self.pick_2.action_done()
+
+        backorder = Picking.search([("backorder_id", "=", self.pick_2.id)])
```
peterpratt13, line 986: can we not use .backorder_id here?

addons/udes_stock/models/stock_move.py
```diff
@@ -468,9 +468,9 @@ def _prepare_extra_info_for_new_picking_for_group(self, pickings, moves):
         if len(partners) == 1:
             values["partner_id"] = partners.id
 
-        dates_done = list(set(moves.mapped("date")))
-        if len(dates_done) == 1:
-            values["date_done"] = dates_done[0]
+        dates_done = set(moves.mapped("date"))
+        if set(moves.mapped("state")).issubset({"done", "cancel"}):
+            values["date_done"] = max(dates_done) # do we want the max value?
```
peterpratt13, line 473: I presume so due to post_validate refactor

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -923,3 +923,75 @@ def test_split_mixed_products_partial_reserve(self):
...
+        product_info_1 = [{"product": cls.apple, "qty": 5}]
+        product_info_2 = [{"product": cls.apple, "qty": 10}, {"product": cls.banana, "qty": 10}]
+
+        cls.pick_1 = cls.create_picking(cls.picking_type_in, products_info=product_info_1)
+        cls.pick_2 = cls.create_picking(cls.picking_type_in, products_info=product_info_2)
+
+    def test_does_not_propagate_date_done_to_incomplete_picking(self):
+        """Check the done date is not carried over to incomplete pickings"""
+        # Complete pick_1 and assert put away is generated and then assert that date done is set on the Goods In but not on the put away
+        self.pick_1.action_assign()
```
peterpratt13, line 956: You can probably assign in the create function

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -923,3 +923,75 @@ def test_split_mixed_products_partial_reserve(self):
...
+        self.pick_2.action_assign()
+        move_line_id_1, move_line_id_2 = self.pick_2.move_line_ids
+        move_line_id_1.write({"location_dest_id": self.received_location.id, "qty_done": 10})
+        move_line_id_2.write({"location_dest_id": self.received_location.id, "qty_done": 1})
+        self.pick_2.action_done()
+
+        backorder = Picking.search([("backorder_id", "=", self.pick_2.id)])
+
+        # check that pick_2 is done and backorder is not done and backorder does not have a date done
+        self.assertEqual(self.pick_2.state, "done")
```
peterpratt13, line 989: Want to check the date_done is set - like in test 1

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -923,3 +923,75 @@ def test_split_mixed_products_partial_reserve(self):
...
+        self.assertNotEqual(backorder.state, "done")
+        self.assertFalse(backorder.date_done)
+
+        putaway_picking = self.pick_2.u_next_picking_ids
+
+        # check that putaway is not in state done and does not have date done set
+        self.assertNotEqual(putaway_picking.state, "done")
+        self.assertFalse(putaway_picking.date_done)
```
peterpratt13, line 997: Do we want a post validate test?

peterpratt13: Can we update this test as well please
    def test02_maintain_single_pick_extra_info(self):
    def test02_maintain_two_picks_extra_info(self):

Also, these names were copy and pasta, so can we bump the number up for it please.

peterpratt13:
LGTM - few nits

addons/udes_stock/tests/test_refactoring_splitting.py
montymagic9: Just to clarify, do you mean the setup function?

peterpratt13: you can pass assign=True as a kwarg in the create_picking function so we don't need to call this seperately.

addons/udes_stock/models/stock_move.py
```diff
@@ -468,9 +468,9 @@ def _prepare_extra_info_for_new_picking_for_group(self, pickings, moves):
         if len(partners) == 1:
             values["partner_id"] = partners.id
 
-        dates_done = list(set(moves.mapped("date")))
-        if len(dates_done) == 1:
-            values["date_done"] = dates_done[0]
+        dates_done = set(moves.mapped("date"))
+        if set(moves.mapped("state")).issubset({"done", "cancel"}):
```
PeterC10, line 472: Q: Why do we check for cancelled moves here?

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -923,3 +923,75 @@ def test_split_mixed_products_partial_reserve(self):
...
+                "location_dest_id": self.received_location.id,
+                "qty_done": self.pick_1.move_line_ids.product_uom_qty,
+            }
+        )
+        self.pick_1.action_done()
+
+        self.assertEqual(self.pick_1.state, "done")
+        self.assertEqual(
+            self.pick_1.date_done, self.pick_1.move_lines.date
+        )  # do we care about the specific date done or just whether the value exists?
```
PeterC10, line 968: A: We care about the specific date done.

peterpratt13: This fails in CI - probably because the pick date is not the same as the move lines date if it is done as normal. Once all the moves are done, then the date_done gets written (check i think action-done or something similar). So these are not necessarily equal - because the operations of the write occur at different points

peterpratt13: The time they are the same is in post_validate_refactor

PeterC10: Will need to look a bit more into it, feels like we should do some sort of check above just seeing if anything is set.

peterpratt13: FTI: action_done in stock_picking:
        todo_moves._action_done()
        self.write({'date_done': fields.Datetime.now()}) 
so the times are different

peterpratt13: Maybe we could call action_done on the moves, check the picking info, then call action_done on the picking. Or do some mocking and check when the date_done write occurs

montymagic9: What is wrong with just checking the date done exists?

peterpratt13: I think the point Boss Pete is making, is that due to refactoring, the write of the date_done on a picking happens lots. Which we don't want - in this case i think checking the date done before action-done is false then true and >= the last move date.

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -977,16 +973,16 @@ def test_does_not_propagate_done_date_to_back_order(self):
...
         Picking = self.env["stock.picking"]
 
-        self.pick_2.action_assign()
         move_line_id_1, move_line_id_2 = self.pick_2.move_line_ids
         move_line_id_1.write({"location_dest_id": self.received_location.id, "qty_done": 10})
         move_line_id_2.write({"location_dest_id": self.received_location.id, "qty_done": 1})
         self.pick_2.action_done()
 
-        backorder = Picking.search([("backorder_id", "=", self.pick_2.id)])
+        backorder = Picking.backorder_id
```
peterpratt13, line 981: backorder = self.pick_2.backorder_id

addons/udes_stock/models/stock_move.py
peterpratt13: It is because if a picking has 2 moves, they cancel one then do the second - then it should have the date_done copied across. But i have no idea if somehow you can copy all cancelled moves into a new picking - i hope that doesn't happen in Odoo :)

PeterC10: Ah okay, so if that means if we have a picking with 1 move cancelled and the other done, then `issubset({"done"})` would filter it out, hence why we need to allow cancel too?

PeterC10:
Nit: Black formatting please

addons/udes_stock/tests/test_refactoring_splitting.py
PeterC10: We have `test02_maintain_single_pick_extra_info` in another test class which checks for date done, would that suffice do you think? (So long as it is passing)

addons/udes_stock/tests/test_refactoring_splitting.py
montymagic9: What would you like me to update for them tests other than the name?

peterpratt13: Check that self.picking | self.picking_2 have the date_done as expected

addons/udes_stock/tests/test_refactoring_splitting.py
peterpratt13: I think those tests split on the move and have grouping on post confirm so the refactor only happens on the split

addons/udes_stock/tests/test_refactoring_splitting.py
peterpratt13: But not much actually- thought they split not on the move. Can probably get away without it :)

peterpratt13:
Are we removing the fi in udes_kafka_events as well? Or is this out of scope?
https://taiga.unipart.digital/project/udes-x/us/19?milestone=997

PeterC10:
> Are we removing the fi in udes_kafka_events as well? Or is this out of scope? https://taiga.unipart.digital/project/udes-x/us/19?milestone=997

Out of scope for this, will handle that separately.

addons/udes_stock/tests/test_refactoring_splitting.py
PeterC10: Have asked Caleb to look into the following test, with post validation refactoring only:

1. Do a picking for an Apple and a Banana and complete it
2. On post validation, refactor by product so we end up with 2 completed pickings instead of 1
3. Make sure both have the date done set (the new one should have a date done greater or equal to the original)

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -918,8 +926,129 @@ def test_split_mixed_products_partial_reserve(self):
...
+        move_format = [(m.product_id, m.product_uom_qty, m.reserved_availability) for m in moves]
         expected_move_format = [(self.apple, 2, 0), (self.banana, 3, 0)]
         self.assertCountEqual(move_format, expected_move_format)
+
+
+class TestRefactoringDateDone(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        """
+        Create stock: pallet with apples, pallet with bananas
```
peterpratt13, line 938: nit - delete this line as we don't do this in the setup

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -918,8 +926,129 @@ def test_split_mixed_products_partial_reserve(self):
...
+        # check that putaway is not in state done and does not have date done set
+        self.assertNotEqual(putaway_picking.state, "done")
+        self.assertFalse(putaway_picking.date_done)
+
+    def test_post_validate_refactor(self):
+        """Test123"""
+
+        mv = self.create_move(self.banana, 1, self.pick_1)
+        pickings = self.pick_1 | self.pick_2
+        self.pick_1.action_assign()
```
peterpratt13, line 1011: nit: move this up one line

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -918,8 +926,129 @@ def test_split_mixed_products_partial_reserve(self):
...
+        self.pick_1.date_done = time_shift
+        self.pick_1.move_lines.filtered(lambda mv: mv.product_id == self.apple).write({"date": time_shift})
+
+        self.assertGreater(self.pick_1.date_done, self.pick_2.date_done)
+
+        # Call refactor
+        pickings.mapped("move_lines")._action_refactor(stage="validate")
+
+        all_pickings = self.env["stock.picking"].search([("picking_type_id", "=", self.picking_type_in.id)])
+        draft_pickings = all_pickings.filtered(lambda p: p.state == "draft")
```
peterpratt13, line 1040: Might want to add something like 
self.assertTrue(set(pickings).issubset(draft_pickings)) 
to show that the refactoring that is being done is moving everything into nnew pickings
Or just add a comment

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -918,8 +926,129 @@ def test_split_mixed_products_partial_reserve(self):
...
+        banana_picking = refactored_pickings - apple_picking
+        self.assertEqual(len(apple_picking), 1)
+        self.assertEqual(len(banana_picking), 1)
+        # Due to refactoring, the pickings get moved out of a done picking
+        # So empty pickings have a date done. This should be ok provided these
+        # draft pickings are never reused - but unsure if other refactoring takes
+        # use of draft pickings. (Presumably unlink resolves this issues quickly in refactoring)
+        # self.assertEqual(draft_pickings.mapped("date_done"), [False])
+        self.assertTrue(banana_picking.date_done)
+        for move in apple_picking.move_lines:
```
peterpratt13, line 1052: Acutally lookign back at this might want to just comapre the move dates directly

apple_mv_1, apple_mv_2 = apple_picking.move_lines.sorted(key=lambda mv: mv.date)
self.assertGreater(apple_mv_2.date, apple_mv_1.date)
self.assertEqual(apple_picking.date_done, apple_mv_2.date)

Similar for banana - but that they are all equal

addons/udes_stock/tests/test_refactoring_splitting.py
```diff
@@ -918,8 +926,129 @@ def test_split_mixed_products_partial_reserve(self):
...
+        draft_pickings = all_pickings.filtered(lambda p: p.state == "draft")
+        refactored_pickings = all_pickings - draft_pickings
+        apple_picking = refactored_pickings.filtered(lambda p: p.mapped("move_lines.product_id") == self.apple)
+        banana_picking = refactored_pickings - apple_picking
+        self.assertEqual(len(apple_picking), 1)
+        self.assertEqual(len(banana_picking), 1)
+        # Due to refactoring, the pickings get moved out of a done picking
+        # So empty pickings have a date done. This should be ok provided these
+        # draft pickings are never reused - but unsure if other refactoring takes
+        # use of draft pickings. (Presumably unlink resolves this issues quickly in refactoring)
```
peterpratt13, line 1049: Probably need to check with miquel or boss pete on this @PeterC10

PeterC10 (approved):

PR #530, opened 2021-12-01
[ADD] udes_stock, udes_common modules

PeterAlabaster:
- Change `track_visibility="onchange"` on fields with `tracking=True`
- Replaced calls to `pick.action_done()` with `pick._action_done()`
- Use
```
@tagged("post_install")
```
in place of
```
@common.at_install(False)
@common.post_install(True)
```
- Patch MSM_STR_DOMAIN instead of writing to readonly object property
- Remove unused imports
- Use f-strings where possible
- Tweak create_product() in tests/common.py to set a different default_code to barcode
- Fix tests to reflect updated PROCUREMENT_PRIORITIES in odoo stock module
- Prevent deprecationwarning by using `e.exception.args[0]` instead of `e.exception.name`

User-story: 420


addons/udes_stock/data/company.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <record id="base.main_company" model="res.company">
+      <field name="name">UDES</field>
+    </record>
+
+    <!-- Details for partner --> 
+    <record id="base.main_partner" model="res.partner">
+      <field name="email">noreply@unipart.com</field>
```
PeterC10, line 10: Note: Not sure about using this email... just make a note here for my own benefit.

addons/udes_stock/data/locations.xml
```diff
@@ -0,0 +1,56 @@
...
+    <record id="stock.stock_location_stock" model="stock.location">
+      <field name="usage">view</field>
+    </record>
+
+
+    <!-- Construct "Warehouse/Input" view location -->
+    <record id="stock_location_input" model="stock.location">
+      <field name="name">Input</field>
+      <field name="location_id" ref="stock.warehouse0_view_location"/>
+      <field name="comment">Input zone</field>
```
PeterC10, line 15: Nit: Can we uppercase `zone`? Also for output and received for `location_input_received`.

Thinks it looks nicer, i.e. like how we have `Goods Out Lane` on line 40.

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,70 @@
...
+
+    <!-- Create "Pick, Goods Out" route -->
+    <record id="route_out" model="stock.location.route">
+      <field name="name">Outbound</field>
+      <field name="sequence" eval="10"/>
+      <field name="warehouse_selectable" eval="True"/>
+      <field name="warehouse_ids" eval="[(6, False, [ref('stock.warehouse0')])]"/>
+    </record>
+
+    <record id="procurement_in_putaway" model="stock.rule">
```
PeterC10, line 21: Super nit: Could we add a comment for this rule, it's the only one without.

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,70 @@
...
+      <field name="route_id" eval="route_out"/>
+      <field name="sequence" eval="10"/>
+      <field name="location_src_id" ref="stock.stock_location_stock"/>
+      <field name="location_id" ref="location_goods_out_lane"/>
+      <field name="procure_method">make_to_stock</field>
+      <field name="action">pull</field>
+      <field name="picking_type_id" ref="picking_type_pick"/>
+    </record>
+
+    <!-- Create "Goods Out" rules -->
```
PeterC10, line 44: Super Nit: Singular `rule` instead of `rules`

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,70 @@
...
+      <field name="route_id" eval="route_out"/>
+      <field name="sequence" eval="30"/>
+      <field name="location_src_id" ref="location_goods_out_lane"/>
+      <field name="location_id" ref="location_trailer"/>
+      <field name="procure_method">make_to_order</field>
+      <field name="action">pull</field>
+      <field name="picking_type_id" ref="stock.picking_type_out"/>
+    </record>
+
+    <!-- Create "Trailer Dispatch" procurement rules -->
```
PeterC10, line 56: Super nit: Remove `procurement`. The model was previously called `procurement.rule` but that is soooo Odoo 11. 😅

addons/udes_stock/data/stock_config.xml
```diff
@@ -0,0 +1,29 @@
...
+      <field name="group_stock_tracking_lot" eval="1"/>
+      <!-- Enable picking batches -->
+      <field name="module_stock_picking_batch" eval="1"/>
+      <!-- Enable product variants -->
+      <field name="group_product_variant" eval="1"/>
+      <!-- Enable products tracked by lot/serial numbers -->
+      <field name="group_stock_production_lot" eval="1"/>
+      <!-- Enable Display Lots & Serial Numbers on Delivery Slips -->
+      <field name="group_lot_on_delivery_slip" eval="1"/>
+      <!--<field name="module_stock_picking_batch">True</field>-->
```
PeterC10, line 17: Nit: Commented out code, already enabled on line 10.

addons/udes_stock/models/mixin_stock_model.py
```diff
@@ -0,0 +1,69 @@
...
+class MixinStockModel(models.AbstractModel):
+    _name = "mixin.stock.model"
+    _description = "Inherited Model"
+
+    # Whether the model is allowed to create a record when using get_or_create
+    MSM_CREATE = False
+    # Default fields to search with a str on get_or_create
+    MSM_STR_DOMAIN = ("name",)
+
+    def _get_msm_domain(self, identifier):
```
PeterC10, line 16: Nit: Could we add a doc-string to this method?

addons/udes_stock/models/res_users.py
```diff
@@ -0,0 +1,40 @@
...
+            :returns: Warehouse(s), or a singular warehouse if aux_domain not None
+        """
+        Warehouse = self.env["stock.warehouse"]
+
+        user_id = self.env.uid
+
+        if user_id != SUPERUSER_ID:
+            user = self.search([('id', '=', user_id)])
+            if not user:
+                raise ValidationError(_('Cannot find user'))
```
PeterC10: Nit: Evil single quotes.

addons/udes_stock/models/res_users.py
```diff
@@ -0,0 +1,40 @@
...
+        domain = [("company_id", "=", self.env.user.company_id.id)]
+        if aux_domain is not None:
+            domain += aux_domain
+        warehouse = Warehouse.search(domain)
+        if not warehouse:
+            raise ValidationError(_("Cannot find a warehouse for user"))
+        if len(warehouse) > 1 and aux_domain is not None:
+            raise ValidationError(
+                _(
+                    "Found multiple warehouses for user still, "
```
PeterC10, line 36: Nit: `Found multiple warehouses for user still` reads a bit odd, I think we should just ditch the word `still`.

addons/udes_stock/models/res_users.py
PeterAlabaster: That's weird, i black formatted the entire dir. Wonder why it didn't change this..

addons/udes_stock/models/mixin_stock_model.py
PeterAlabaster: I don't know the context behind this function (other than the variables it used causing some issues in tests). To help write the docstring context on why it exists would be useful

addons/udes_stock/models/res_users.py
PeterAlabaster: ah.. just checked locally. you must be looking at the initial commit. commit after black formats the entire thing which in turn resolves this

addons/udes_stock/models/stock_quant_package.py
```diff
@@ -0,0 +1,105 @@
...
+            raise ValidationError(
+                _(
+                    "Cannot move packages with reserved contents. "
+                    "Please speak to a team leader to resolve the issue.\n"
+                    "Affected packages: %s"
+                )
+                % (", ".join(reserved_packages.mapped("name")))
+            )
+
+    def get_quantities_by_key(self, get_key=lambda q: q.product_id):
```
PeterC10, line 30: Suggestion: Feels a bit jarring having a lambda as default argument. Personally I think the following would be neater:

```
def get_quantities_by_key(self, get_key=None):
    if get_key is None:
        get_key = lambda q: q.product_id
```

Might be worth getting a second opinion on that though... or if you prefer how it is now feel free to resolve this one. :D

addons/udes_stock/models/mixin_stock_model.py
PeterAlabaster: am thinking something like
```python
     def _get_msm_domain(self, identifier):
        """Return a domain based on MSM_STR_DOMAIN attribute
        (which can be changed on the model this MixinStockModel is mixed in to"""
```

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class StockWarehouse(models.Model):
+    _name = "stock.warehouse"
+    _inherit = ["stock.warehouse", "mail.thread"]
+
+    # Add tracking for archiving.
+    active = fields.Boolean(track_visibility="onchange")
+    u_pallet_barcode_regex = fields.Char("Pallet Barcode Format", default="^UDES(\\d)+$")
+    u_package_barcode_regex = fields.Char("Package Barcode Format", default="^UDES(\\d)+$")
```
PeterC10, line 13: Default for both pallets and packages are `UDES` could that trip us up?

addons/udes_stock/models/stock_quant_package.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+from collections import defaultdict
```
PeterC10, line 4: Nit: Unused import.

addons/udes_stock/models/stock_warehouse.py
PeterAlabaster: In 11.0 these are all set in most customers `warehouse.xml` data file anyway. If we want to change the default, am happy to

addons/udes_stock/models/stock_quant.py
```diff
@@ -0,0 +1,75 @@
...
+    _inherit = "stock.quant"
+    _description = "Quant"
+
+    @api.model
+    def _get_removal_strategy_order(self, removal_strategy=None):
+        """Change fifo removal strategy"""
+        # NOTE: Preserve `NULLS FIRST` for `in_date` and use for `package_id` as well
+        if removal_strategy == "fifo":
+            return "in_date ASC NULLS FIRST, package_id ASC NULLS FIRST"
+        return super(Quant, self)._get_removal_strategy_order(removal_strategy)
```
PeterC10, line 16: `Quant` should be `StockQuant`.

I think this was a long standing bug we recently sorted on 11. Luckily we had only been using first in first out!

addons/udes_stock/models/res_users.py
PeterC10: Whoops, yep sorry I have a bad habit of that!

addons/udes_stock/models/stock_warehouse.py
PeterC10: The idea with most systems we do going forwards though is that we will avoid have a customer module unless really needed. So most likely we would configure this anyway, it just seems having both the same as default seems dangerous. I'll ask around though.

addons/udes_stock/models/stock_quant.py
PeterAlabaster: i remember! i came across it :D will get it fixed

addons/udes_stock/data/company.xml
PeterC10: Closing, will look more into it and handle separately if needed.

addons/udes_stock/models/mixin_stock_model.py
PeterC10: That looks good to me :) Only super nit is that Odoo should be capitalised, sorry 😅

addons/udes_stock/models/stock_quant_package.py
nickcosta85: I don't mind seeing a lambda function as default augment but that's me, having said that if you do decide to change it Peter maybe also change the has_same_content() function on line 59 to match for consistency?

addons/udes_stock/models/stock_move.py
```diff
@@ -0,0 +1,148 @@
...
+        of move_lines.
+        If self is completely covered by move_lines, it will be removed from
+        its picking and returned.
+        Included Micky's code for splitting out partially done ones.
+        Preconditions: self is a single move,
+                       all moves_line are attached to self
+        :return: The (possibly new) move that covers all of move_lines,
+                 not currently attached to any picking.
+
+            Note: not using stock.move._split() since we want better handling of
```
peterpratt13, line 92: Nit: Could also add now that _split no longer returns a move like this function

addons/udes_stock/models/stock_move_line.py
```diff
@@ -0,0 +1,183 @@
...
+        to do is equal to parameter quantity.
+        In case that a move line needs to be split, the new move line is
+        also returned (this happens when total quantity in the move lines is
+        greater than quantity parameter).
+        If there is not enough quantity to do in the move lines,
+        also return the remaining quantity.
+        """
+        new_ml = None
+        result = self.browse()
+        if quantity == 0:
```
peterpratt13, line 43: I can't remember - what would the use case of quantity==0 be here? Is it just due to iteration over something possibly?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -0,0 +1,183 @@
...
+        """Return a subset of move lines from self where their sum of quantity
+        to do is equal to parameter quantity.
+        In case that a move line needs to be split, the new move line is
+        also returned (this happens when total quantity in the move lines is
+        greater than quantity parameter).
+        If there is not enough quantity to do in the move lines,
+        also return the remaining quantity.
+        """
+        new_ml = None
+        result = self.browse()
```
peterpratt13, line 42: Age old question - MoveLine = self; MoveLine.browse(), MoveLine = self.env["stock.move.line"]; MoveLine.browse(),  self.browse()? Time to define 14 convention :)

PeterC10: Hah good point. I think this is preferred:

```
MoveLine = self.env["stock.move.line"]
result = self.browse()
```

@ws-mps can you confirm? 😁

addons/udes_stock/models/stock_move_line.py
PeterAlabaster: I have no idea

addons/udes_stock/models/stock_move_line.py
PeterAlabaster: personally i like:
```python
result = self.browse() # if self is the target model

MoveLine = self.env["stock.move.line"]
result = MoveLine.browse()  # if self is not the target model
```

addons/udes_stock/models/stock_move_line.py
peterpratt13: no worries - i guess must be iteration. Else why would you call that function to return an empty move line :)

addons/udes_stock/models/stock_quant_package.py
PeterAlabaster: If it's OK i will leave this one

addons/udes_stock/models/stock_warehouse.py
PeterC10: Let's do the same as 11 for now, so pallet default `PAL` and package default `PACK`. We can always adjust later if needed.

addons/udes_stock/models/stock_move_line.py
PeterAlabaster: as discussed official way is second (explicitly defining the model ref), have changed.

armandcela: I agree with Peter if self is the target model :)

PeterC10 (approved):
LGTM

peterpratt13 (approved):

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+
+    <!-- Create "Goods In, Putaway" route -->
+    <record id="route_in" model="stock.location.route">
+      <field name="name">Inbound</field>
+      <field name="sequence" eval="10"/>
+      <field name="warehouse_selectable" eval="True"/>
+      <field name="warehouse_ids" eval="[(6, False, [ref('stock.warehouse0')])]"/>
```
armandcela, line 10: How about adding a warehouse not resetting it to only stock.warehouse0, maybe can be done with 4, ref('stock.warehouse0') not sure about the proper syntax. 
I am saying this just in case that if in future we want to apply multi company and we may add another warehouse on fly for the same route. 
I have a concern in general with data that we dont use noupdate=1 even if we use we ran the service as -i which it will think that is a new instalment and will override values changed from users.
Maybe thinking to much :)

addons/udes_stock/models/res_users.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from odoo import api, fields, models, SUPERUSER_ID, _
+from odoo.exceptions import ValidationError
+
+
+class ResUser(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def get_user_warehouse(self, aux_domain=None):
```
armandcela, line 10: Maybe the name of method as we may expect more than one warehouse and inside when we search it is warehouse = self.search .. should be plural :)
```suggestion
    def get_user_warehouses(self, aux_domain=None):
```

addons/udes_stock/models/res_users.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if user_id != SUPERUSER_ID:
+            user = self.search([("id", "=", user_id)])
+            if not user:
+                raise ValidationError(_("Cannot find user"))
+
+        domain = [("company_id", "=", self.env.user.company_id.id)]
+        if aux_domain is not None:
+            domain += aux_domain
+        warehouse = Warehouse.search(domain)
```
armandcela, line 30:
```suggestion
        warehouses = Warehouse.search(domain)
```

addons/udes_stock/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import fields, models
+
+
+class StockLocation(models.Model):
+    _name = "stock.location"
+    # Add messages and abstract model to locations
+    _inherit = ["stock.location", "mail.thread", "mixin.stock.model"]
+
+    # Disable translation instead of renaming.
+    name = fields.Char(translate=False)
```
armandcela, line 12: I think there is a way of overriding a method from ir_model_fields i think that can add/modify attribute for existing fields instead of redefining the field again. In that way can have smth more generic helper name of the field, model, attribute , value example and when overriding the method to pick all changes from helpers.
Just imagining now :)

Solving the comment though...

addons/udes_stock/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+    def get_common_ancestor(self):
+        """
+        Returns the smallest location containing all the locations in self.
+        Locations are considered to contain themselves.
+
+        :returns:
+            The stock.location containing all the locations in self,
+            or an empty stock.location recordset if there is no such location.
+        """
+        Location = self.env["stock.location"]
```
armandcela, line 25: Added this comment as you preferred  to use self when target model is self :)
```suggestion
        
```

addons/udes_stock/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if len(self) <= 1:
+            return self
+
+        # Each location's parent_path is a "/"-delimited string of parent ids
+        # including itself
+        common_path = os.path.commonpath(self.mapped("parent_path"))
+        id = common_path.split("/")[-1]
+        if id == "":
+            return Location.browse()
```
armandcela, line 35:
```suggestion
            return self.browse()
```

addons/udes_stock/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+        else:
+            return Location.browse(int(id))
```
armandcela, line 37:
```suggestion
            return self.browse(int(id))
```

addons/udes_stock/models/stock_location.py
```diff
@@ -0,0 +1,37 @@
...
+            or an empty stock.location recordset if there is no such location.
+        """
+        Location = self.env["stock.location"]
+
+        if len(self) <= 1:
+            return self
+
+        # Each location's parent_path is a "/"-delimited string of parent ids
+        # including itself
+        common_path = os.path.commonpath(self.mapped("parent_path"))
```
armandcela, line 32: I think from odoo13 no need to mapp if i am not wrong :)
```suggestion
        common_path = os.path.commonpath(self.parent_path)
```

addons/udes_stock/models/stock_move.py
```diff
@@ -0,0 +1,150 @@
...
+        MoveLine = self.env["stock.move.line"]
+
+        return MoveLine.create(move_line_values)
+
+    def _unreserve_initial_demand(self, new_move):
+        """Override stock default function to keep the old move lines,
+        so there is no need to create them again
+        """
+        self.ensure_one()
+        self.mapped("move_line_ids").filtered(lambda x: x.qty_done == 0.0).write(
```
armandcela, line 76:
```suggestion
        self.move_line_ids.filtered(lambda x: x.qty_done == 0.0).write(
```

peterpratt13: can we change x -> ml if we do thee change :)

addons/udes_stock/models/stock_move.py
```diff
@@ -0,0 +1,150 @@
...
+            {"move_id": new_move.id, "product_uom_qty": 0}
+        )
+
+    def split_out_move_lines(self, move_lines, **kwargs):
+        """Split sufficient quantity from self to cover move_lines, and
+        attach move_lines to the new move. Return the move that now holds all
+        of move_lines.
+        If self is completely covered by move_lines, it will be removed from
+        its picking and returned.
+        Included Micky's code for splitting out partially done ones.
```
armandcela, line 86: Not fan of using specific names on comments :)

addons/udes_stock/models/stock_move_line.py
```diff
@@ -0,0 +1,184 @@
...
+    def move_lines_for_qty(self, quantity, sort=True):
+        """Return a subset of move lines from self where their sum of quantity
+        to do is equal to parameter quantity.
+        In case that a move line needs to be split, the new move line is
+        also returned (this happens when total quantity in the move lines is
+        greater than quantity parameter).
+        If there is not enough quantity to do in the move lines,
+        also return the remaining quantity.
+        """
+        MoveLine = self.env["stock.move.line"]
```
armandcela, line 41:
```suggestion
        
```

addons/udes_stock/models/stock_move_line.py
```diff
@@ -0,0 +1,184 @@
...
+        new_ml = None
+        result = MoveLine.browse()
```
armandcela, line 43:
```suggestion
        result = self.browse()
```

armandcela (commented):
General comment

addons/udes_stock/data/routes.xml
PeterAlabaster: I don't believe there are any plans for multi-warehouse systems in the future

addons/udes_stock/models/res_users.py
PeterAlabaster: I don't believe there are any plans for multi-warehouse systems in the future

addons/udes_stock/models/stock_location.py
PeterAlabaster: Discussed in other thread, we should not use self.browse() and instead should explicitly declare the self.env["target.model"]

addons/udes_stock/models/stock_location.py
PeterC10: Our practice is to use the name of the object over self in this cases, to make it more explicit.

addons/udes_stock/models/res_users.py
armandcela: On the code there is a control if len > 1 and aux_domain 
To my understanding it means that we may expect some warehouses, its not related to multi company.
Event the docstring says warehouse-s smth like that

addons/udes_stock/models/stock_move_line.py
PeterC10: As per https://github.com/unipartdigital/udes-open/pull/530#discussion_r761163318 - we prefer to use the name of the object to make it more explicit.

Understand this is a contentious issue and will take a while until we're doing it consistently.

addons/udes_stock/models/stock_location.py
PeterC10: As per https://github.com/unipartdigital/udes-open/pull/530#discussion_r761163318 - we prefer to use the name of the object to make it more explicit.

Understand this is a contentious issue and will take a while until we're doing it consistently.

addons/udes_stock/models/stock_location.py
PeterC10: As per https://github.com/unipartdigital/udes-open/pull/530#discussion_r761163318 - we prefer to use the name of the object to make it more explicit.

Understand this is a contentious issue and will take a while until we're doing it consistently.

addons/udes_stock/models/stock_move_line.py
PeterC10: (Sorry for another copy/paste)

As per https://github.com/unipartdigital/udes-open/pull/530#discussion_r761163318 - we prefer to use the name of the object to make it more explicit.

Understand this is a contentious issue and will take a while until we're doing it consistently.

armandcela (approved):

addons/udes_stock/models/res_users.py
```diff
@@ -0,0 +1,40 @@
...
+        Warehouse = self.env["stock.warehouse"]
+
+        user_id = self.env.uid
+
+        if user_id != SUPERUSER_ID:
+            user = self.search([("id", "=", user_id)])
+            if not user:
+                raise ValidationError(_("Cannot find user"))
+
+        domain = [("company_id", "=", self.env.user.company_id.id)]
```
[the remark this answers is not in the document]
PeterAlabaster: the commit in 13.0 that added this:
```
    [FIX] udes_stock: Added check for system user in get_user_warehouse
     
    A check is carried out in get_user_warehouse to see if the user is active. However the system user is excluded from searches. An earlier check is now carried out to see if the user_id matches SUPERUSER_ID. If it doesn't then the check is carried out as before.
     
    Story/9859
     
```
i don't know if the self.env.uid / self.env.user is relevant, there's no comment to say otherwise so i'll change it

addons/udes_stock/models/res_users.py
PeterAlabaster: I don't believe there are any plans for multi-warehouse systems in the future

addons/udes_stock/models/stock_location.py
PeterAlabaster: Ignoring as per other comment

addons/udes_stock/models/stock_location.py
PeterAlabaster: Ignoring as per other comment

addons/udes_stock/models/stock_location.py
PeterAlabaster:
```python
os.path.commonpath(self.parent_path)
Traceback (most recent call last):
  File "/home/palabaster/projects/udes-vagrant/code/udes14/odoo/odoo/models.py", line 4986, in ensure_one
    _id, = self._ids
ValueError: too many values to unpack (expected 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/palabaster/projects/udes-vagrant/code/udes14/odoo/odoo/fields.py", line 930, in __get__
    record.ensure_one()
  File "/home/palabaster/projects/udes-vagrant/code/udes14/odoo/odoo/models.py", line 4989, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: stock.location(3702, 3703)
```
does not work as `self` is > 1

armandcela: I have been told that no need to use mapping in odoo13 and above :)

addons/udes_stock/models/res_users.py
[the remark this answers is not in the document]
PeterC10: I wouldn't trust that guy. 👀

I'm not sure what the suggested change is.

The `user_id != SUPERUSER_ID` was added to avoid unit tests falling over as the system user was being reported as not found on v13. :)

addons/udes_stock/models/stock_location.py
PeterAlabaster: note: discussed further in skype. Odoo 13 onwards allows not using .mapped() for relational fields (m2o, o2m) but not for normal fields like char / boolean:
```python
self
stock.location(4166, 4167)
self.location_id
stock.location(4164,)
self.location_id.child_ids
stock.location(4166, 4167)
self.name --> traceback
```

addons/udes_stock/models/res_users.py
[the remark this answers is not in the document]
PeterAlabaster: basically using `self.env.user` and `self.env.user.id` only, rather than switching between `self.env.user` and `self.env.uid`

PR #531, opened 2021-12-02
DO NOT MERGE Story/19 pete

peterpratt13:


peterpratt13:
I'm literally running this to see what happens in docker - sorry.
Basically in docker it reuses a pick- outside it doesn't....

PR #532, opened 2021-12-03
[ADD] udes_open: Add u_original_picking_id o stock.picking

peterpratt13:
Add u_original_picking_id to track thee first picking.
Add unit tests for u_user_process_partial

User-story: 291

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -80,6 +80,12 @@ class StockPicking(models.Model):
         compute="_compute_related_picking_ids",
         help="Created Back Orders",
     )
+    u_original_picking_id = fields.Many2one(
+        "stock.picking",
+        string="Original Picking",
+        help="The original picking of the refactored/backordered picking",
+        copy=True,
+    )
 
     # search helpers for source and destination package
     u_package_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -155,6 +161,23 @@ class StockPicking(models.Model):
...
 
+    def create(self, vals):
+        """
+        Add the original picking id on creation because of race conditions
+        of computed fields.
+        For example, when creating a backorder,
+        the backorder gets related but u_prev_picking_ids gets done after.
+        So instead, we just write to a new field which we copy across backordered pickings
+        """
+        # if not vals.get("backorder_id"):
```
PeterC10, line 172: Nit: Can we remove the commented out code if not needed?

PeterC10 (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -80,6 +80,12 @@ class StockPicking(models.Model):
         compute="_compute_related_picking_ids",
         help="Created Back Orders",
     )
+    u_original_picking_id = fields.Many2one(
+        "stock.picking",
+        string="Original Picking",
+        help="The original picking in the chain",
```
ws-mps, line 86: I wouldn't consider it as original picking of the chain, since a chain can be considered a multi steps route. I would probably consider this field the original picking of the backorders.

adamrpatrick (approved):

ws-mps (approved):

PR #541, opened 2021-12-15
[ADD] udes_common: Helper function selection_value_displayname()

PeterAlabaster:
Does what it says on the tin, gets the displayname for a given
selection fields value on a recordset

Also added a readme and filled in what i could for what already exists

User-story: 411

### DIFF
--- a/addons/udes_common/models/models.py
+++ b/addons/udes_common/models/models.py
@@ -5,7 +5,8 @@ import itertools
 from operator import itemgetter
 from .. import tools
 
-from odoo import models
+from odoo import models, _
+from odoo.exceptions import ValidationError
 
 _logger = logging.getLogger(__name__)
 
@@ -84,3 +85,23 @@ def statistics(self, cache=False):
 def trace(self, filter=None, max=None):
     """Trace database queries"""
     return tools.QueryTracer(self.env.cr, filter=filter, max=max)
+
+
+@add_if_not_exists(models.BaseModel)
+def selection_display_name(self, selection_field_name):
+    """Get the display name - for a given selection fields value - on a recordset
+    :param: self: <some.model> recordset (len 1) to get the value from
+    :param: selection_field_name: str() field name to get the value for
+
+    :returns: str() Display name of the currently set selection field value of the recordset
+    """
+    self.ensure_one()
+    try:
+        selection_field = self._fields[selection_field_name]
+    except KeyError:
+        raise ValidationError(
+            _(f"Field '{selection_field_name}' does not exist on model '{self._name}'")
+        )
+    # Odoo has a nice function that handles all the weird ways you can define selections!
+    selection_values_dict = dict(selection_field._description_selection(self.env))
+    return selection_values_dict.get(self[selection_field_name], False)
--- a/addons/udes_common/tests/test_selection_display_name.py
+++ b/addons/udes_common/tests/test_selection_display_name.py
@@ -1,0 +1,105 @@
+"""Tests for selection field display name"""
+
+from odoo.exceptions import ValidationError
+from odoo.tests import common, tagged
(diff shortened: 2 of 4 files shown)


addons/udes_common/tests/test_selection_display_name.py
```diff
@@ -0,0 +1,117 @@
...
+        """Ensure function returns False if the selection field on the recordset is not set"""
+        self.model_overview_report.report_type = False
+
+        normal_output = self.model_overview_report.report_type
+        self.assertEqual(normal_output, False)
+
+        alt_output = self.model_overview_report.selection_value_displayname("report_type")
+        self.assertEqual(alt_output, False)
+
+    # NB: It's not possible (as far as i'm aware) to mock a new field on a model without lots of pain
```
PeterAlabaster, line 105: Left this in incase anyone has ideas

PeterAlabaster: Removed this as there is really no need to test it as we know it works and it's too tricky to mock

armandcela (approved):

adamrpatrick (approved):

addons/udes_common/models/models.py
```diff
@@ -84,3 +85,23 @@ def statistics(self, cache=False):
 def trace(self, filter=None, max=None):
     """Trace database queries"""
     return tools.QueryTracer(self.env.cr, filter=filter, max=max)
+
+
+@add_if_not_exists(models.BaseModel)
+def selection_value_displayname(self, selection_field_name):
```
PeterC10, line 91: Super nit (well two): I think we can shorten this to `selection_display_name`

1. Don't think value is needed
2. Think `display_name` is tidier than `displayname`

addons/udes_common/models/models.py
```diff
@@ -84,3 +85,23 @@ def statistics(self, cache=False):
...
+    self.ensure_one()
+    try:
+        selection_field = self._fields[selection_field_name]
+    except KeyError:
+        raise ValidationError(
+            _(f"Field '{selection_field_name}' does not exist on model '{self._name}'")
+        )
+    # Odoo has a nice function that handles all the weird ways you can define selections!
+    selection_values_dict = dict(selection_field._description_selection(self.env))
+    return selection_values_dict.get(self[selection_field_name], False)
```
PeterC10, line 107: Not sure if I'm misunderstanding something but if we can't find a display name for the value I think it is better to return the code value rather than `False`.

addons/udes_common/tests/test_selection_display_name.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # A random recordset on a model with a normal selection field
+        cls.model_overview_report = cls.env.ref("base.report_ir_model_overview")
+        # A random recordset on a model with a generated selection field (function)
+        cls.main_partner = cls.env.ref("base.main_partner")
+        # There are no recordsets on res.partner.bank, but it has a lambda generated
+        # selection field, so we just create one
+        cls.res_partner_bank_recordset = ResPartnerBank.create(
+            dict(
+                acc_number="60-16-13 31926819",
```
PeterC10, line 35: Nit: Leaking my bank details online... :D

PeterAlabaster: Taken from odoo/odoo/addons/base/data/res_partner_bank_demo.xml

addons/udes_common/models/models.py
PeterAlabaster: This is more so if the selection fields value is set to False, the .get() will return None.
Just trying to keep consistency with how Odoo behaves when getting the normal selection field value

PeterC10 (approved):

PR #542, opened 2021-12-18
Story/540 archiving restrictions model & config

PeterAlabaster:
To get an idea what the views look like:
image
image

Superuser/user who can access:
image
image

Restricted user:
image
image

Check is both on frontend and backend to block against malicious users de-activating warehouses via odoorpc

### DIFF
--- a/addons/udes_security/models/models.py
+++ b/addons/udes_security/models/models.py
@@ -7,9 +7,27 @@ from odoo.osv.orm import setup_modifiers
 class BaseModel(models.AbstractModel):
     _inherit = "base"
 
+    @api.multi
+    def write(self, vals):
+        """Inherit write to add a check against res.archiving.restriction
+        when the active field is written to.
+        """
+        ResArchivingRestriction = self.env["res.archiving.restriction"]
+        if "active" in vals:
+            ResArchivingRestriction._check_can_archive(self._name)
+        return super().write(vals)
+
     @api.model
     def fields_view_get(self, view_id=None, view_type="form", toolbar=False, submenu=False):
-        """Override to remove buttons/editable functionality for view only desktop users"""
+        """Override to remove buttons/editable functionality for view only desktop users
+        and take into account archiving restrictions for the given user"""
+
+        def _remove_active(doc):
+            """Remove any field or smart button related to the active field from the doc"""
+            for node in doc.xpath("//button[@name='toggle_active']"):
+                node.getparent().remove(node)
+            for node in doc.xpath("//field[@name='active']"):
+                node.getparent().remove(node)
 
         def _remove_nodes(doc, xpath):
             """Remove any elements from the view that match the supplied xpath"""
@@ -29,6 +47,7 @@ class BaseModel(models.AbstractModel):
                     if field_name in res["fields"]:
                         setup_modifiers(node, res["fields"][field_name])
 
+        ResArchivingRestriction = self.env["res.archiving.restriction"]
         ModelException = self.env["udes.readonly_desktop_user_model_exception"].sudo()
 
         res = super().fields_view_get(
@@ -74,6 +93,12 @@ class BaseModel(models.AbstractModel):
 
             res["arch"] = etree.tostring(doc)
             # Remove archive option in tree view and make calendar readonly
-            res["readonly_desktop_user"] = True
+            res["readonly_desktop_user"] = True  # Used in basic_view.js
+        elif not ResArchivingRestriction._get_can_archive(model):
+            doc = etree.XML(res["arch"])
+            if doc.tag == "form":
+                _remove_active(doc)
+                res["arch"] = etree.tostring(doc)
+            res["cannot_archive_record"] = True  # Used in basic_view.js
 
         return res
--- a/addons/udes_security/security/ir.model.access.csv
+++ b/addons/udes_security/security/ir.model.access.csv
@@ -1,3 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 udes_allowed_file_type,udes_allowed_file_type,model_udes_allowed_file_type,base.group_user,1,0,0,0
 access_udes_readonly_desktop_user_model_exception,access_udes_readonly_desktop_user_model_exception,model_udes_readonly_desktop_user_model_exception,base.group_user,1,0,0,0
+access_res_archiving_restriction,access_res_archiving_restriction,model_res_archiving_restriction,base.group_no_one,1,1,1,1
+access_res_archiving_restriction_user,access_res_archiving_restriction_user,model_res_archiving_restriction,base.group_user,1,0,0,0
--- a/addons/udes_security/__manifest__.py
+++ b/addons/udes_security/__manifest__.py
@@ -17,6 +17,7 @@ Security enhancements used by UDES
         "views/base_menu.xml",
         "views/allowed_file_type_views.xml",
         "views/readonly_desktop_user_exception_views.xml",
+        "views/res_archiving_restriction_views.xml",
         "views/res_users_views.xml",
         "views/webclient_templates.xml",
         "data/auth_brute_force.xml",
--- a/addons/udes_security/models/__init__.py
+++ b/addons/udes_security/models/__init__.py
@@ -3,6 +3,7 @@ from . import ir_actions
 from . import ir_attachment
 from . import models
 from . import readonly_desktop_user_exception
+from . import res_archiving_restriction
 from . import res_authentication_attempt
 from . import res_groups
 from . import res_partner
--- a/addons/udes_security/models/res_archiving_restriction.py
+++ b/addons/udes_security/models/res_archiving_restriction.py
@@ -1,0 +1,87 @@
+import logging
+
+
+from odoo import fields, models, _, SUPERUSER_ID
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+
+class ResArchivingRestriction(models.Model):
+    """
+    Configurable access to _only_ record archiving & unarchiving
+
+    If an Archiving Restriction is created for a model, the permitted_group_ids set will determine
+    which groups have the facility to archive and unarchive records on the model.
+
+    This includes backend writes to the `active` field, as well as hiding
+    the Archive/Unarchive action from the model, and attempting to remove active
+    fields & smart buttons from form views.
+    """
+
+    _name = "res.archiving.restriction"
+    _description = "Record Archiving Restriction"
+    _sql_constraints = [
+        (
+            "model_id_uniq",
+            "unique(model_id)",
+            _("An Archiving Restriction already exists for this model."),
+        )
+    ]
+
+    model_id = fields.Many2one(
+        "ir.model", string="Model", help="Model to apply restriction to", required=True, index=True
+    )
+    name = fields.Char(string="Model Name", related="model_id.model", readonly=True)
+    permitted_group_ids = fields.Many2many(
+        "res.groups",
+        string="Permitted Groups",
+        index=True,
+        help="""
+        Groups Permitted to Archive & Unarchive on the given model.
+
+        If no groups are configured, ALL users will be prevented from archiving (except superuser)
+        If the user is in any of the configured groups, they will be permitted access
+        If no restriction on a model is intended, then do not create a restriction record for that model
+        """,
+    )
+
+    def _get_can_archive(self, model_name):
+        """
+        Check against the model to see if active user is
+        permitted to archive/unarchive the record.
+
+        Can archive if user has _any_ groups defined on the restriction,
+        or no restriction is defined
+
+        :param: model_name: str() model _name i.e "stock.picking.type"
+        :returns: bool()
+        """
+        ResArchivingRestriction = self.env["res.archiving.restriction"]
+
+        if self.env.user.id == SUPERUSER_ID:
+            return True
+        if not model_name:
+            return True
+        model_restriction = ResArchivingRestriction.search([("model_id.model", "=", model_name)])
+        if not model_restriction:
+            return True
+        user_group_ids = self.env.user.groups_id.ids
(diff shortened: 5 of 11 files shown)


addons/udes_security/models/models.py
```diff
@@ -7,9 +7,27 @@
 class BaseModel(models.AbstractModel):
     _inherit = "base"
 
+    @api.multi
+    def write(self, vals):
+        """Inherit write to add a check against res.archiving.restrictions
+        when the active field is written to.
+        """
+        ResArchivingRestrictions = self.env["res.archiving.restrictions"]
+        if "active" in vals:
```
PeterAlabaster, line 16: not sure if we also need to check that we are writing the opposite value to any of the records in `self` in case there are some wonky writes going on in certain models which update `"active" : True` when it is already set to True.
If there are and the user triggering those wonky writes doesn't have archiving permission, it will blow up.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+                "An Archiving Restriction already exists for this model. Please edit the existing one"
+            ),
+        )
+    ]
+
+    model_id = fields.Many2one(
+        "ir.model", string="Model to apply restriction to", required=True, index=True
+    )
+    name = fields.Char(string="Model Name", related="model_id.model", readonly=True)
+    group_ids = fields.Many2many(
```
PeterAlabaster, line 33: self nit: maybe could call this `group_permitted_ids`

addons/udes_security/models/models.py
PeterAlabaster: i was thinking something like this
```python
if "active" in vals:
    current_active = self.mapped("active")
    setting_active = vals.get("active")
    if len(current_active) > 1 or setting_active not in current_active:
        ResArchivingRestrictions._check_can_archive(self._name)
```

addons/udes_security/models/models.py
```diff
@@ -7,9 +7,27 @@
...
+
     @api.model
     def fields_view_get(self, view_id=None, view_type="form", toolbar=False, submenu=False):
-        """Override to remove buttons/editable functionality for view only desktop users"""
+        """Override to remove buttons/editable functionality for view only desktop users
+        and take into account archiving restrictions for the given user"""
+
+        def _remove_active(doc):
+            """Remove any field or smart button related to the active field from the doc"""
+            for node in doc.xpath("//button[@icon='fa-archive']"):
```
PeterC10, line 27: We shouldn't be removing buttons based on what icon they have imo, unlikely be we could be unintentionally removing a button that is needed.

Could we change this to `@name='toggle_active'`? This is the core functionality for archiving in Odoo.

addons/udes_security/models/models.py
```diff
@@ -74,6 +93,15 @@ def _set_fields_readonly(doc, res):
 
             res["arch"] = etree.tostring(doc)
             # Remove archive option in tree view and make calendar readonly
-            res["readonly_desktop_user"] = True
+            res["readonly_desktop_user"] = True  # Used in basic_view.js
+
+        can_archive_record = ResArchivingRestrictions._get_can_archive(model)
```
PeterC10, line 98: Can we `else` this with `if self.env.user.u_desktop_readonly and not ModelException.is_exempt(model):`?

If a user is readonly then they're not going to be able to archive anything anyway so the checks here aren't needed.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    _name = "res.archiving.restrictions"
+    _description = "Record Archiving Restrictions"
+    _sql_constraints = [
+        (
+            "model_id_uniq",
+            "unique(model_id)",
+            _(
+                "An Archiving Restriction already exists for this model. Please edit the existing one"
```
PeterC10, line 24: Super super super nit, feel free to resolve but I think the message should be set as one of the following:

`An Archiving Restriction already exists for this model, please edit the existing one`

or

`An Archiving Restriction already exists for this model. Please edit the existing one.`

or just:

`An Archiving Restriction already exists for this model`

addons/udes_security/models/res_archiving_restrictions.py
PeterC10: Yeah tricky one, as it's hard to work out straight away if `group_ids` means they are exempt or the only ones affected!

Personally I prefer `permitted_group_ids`. :)

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+    ]
+
+    model_id = fields.Many2one(
+        "ir.model", string="Model to apply restriction to", required=True, index=True
+    )
+    name = fields.Char(string="Model Name", related="model_id.model", readonly=True)
+    group_ids = fields.Many2many(
+        "res.groups",
+        string="Groups permitted to archive & unarchive",
```
PeterC10, line 35: Nit: `Groups Permitted to Archive & Unarchive`

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+        help="""
+        If no groups are configured, ALL users will be prevented from archiving (except superuser)
+        If the user is in any of the configured groups, they will be permitted access
+        If no restriction on a model is intended, then do not create a restriction record for that model
+        """,
+    )
+
+    def _get_can_archive(self, model_name):
+        """
+        Test against the model to see if our user is
```
PeterC10, line 46: Suggestion: `Check against the model to see if active user is permitted to archive/unarchive the record.`

"Test" and "our user" read a bit odd to me.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+    def _check_can_archive(self, model_name):
+        """Raises a warning if the logged in user does not meet
+        the group requirements to archive/unarchive a record
+
+        :param: model_name: str() model _name i.e "stock.picking.type"
+        """
+        if not self._get_can_archive(model_name):
+            raise ValidationError(
+                _(
+                    f"Cannot archive/unarchive record due user not meeting restrictions in place on {model_name}"
```
PeterC10, line 79: Suggestion: As this will be a user facing error if it ever comes up (which it shouldn't) I'd suggest rewording it slightly, e.g.:

`You do not have permission to archive/unarchive this record.`

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+        return bool(groups_on_user_and_restriction)
+
+    def _check_can_archive(self, model_name):
+        """Raises a warning if the logged in user does not meet
+        the group requirements to archive/unarchive a record
+
+        :param: model_name: str() model _name i.e "stock.picking.type"
+        """
+        if not self._get_can_archive(model_name):
+            raise ValidationError(
```
PeterC10, line 77: Suggestion: Log this, recording which user tried to archive the record.

addons/udes_security/security/ir.model.access.csv
```diff
@@ -1,3 +1,4 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 udes_allowed_file_type,udes_allowed_file_type,model_udes_allowed_file_type,base.group_user,1,0,0,0
 access_udes_readonly_desktop_user_model_exception,access_udes_readonly_desktop_user_model_exception,model_udes_readonly_desktop_user_model_exception,base.group_user,1,0,0,0
+access_res_archiving_restrictions,access_res_archiving_restrictions,model_res_archiving_restrictions,base.group_no_one,1,1,1,1
```
PeterC10, line 4: Suggestion: Add a readonly permission for all users for safety, like the readonly model exception.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+
+    If an Archiving Restriction is created for a model, the group_ids set will determine
+    which groups have the facility to archive and unarchive records on the model.
+
+    This includes backend writes to the `active` field, as well as hiding
+    the Archive/Unarchive action from the model, and attempting to remove active
+    fields & smart buttons from form views.
+    """
+
+    _name = "res.archiving.restrictions"
```
PeterC10, line 17: Usually the names of records are singular, e.g. `res.archiving.restriction`.

I know Odoo is inconsistent with this itself, but I think it should still be singular here, as we have one record per model that is restricted.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+            "model_id_uniq",
+            "unique(model_id)",
+            _(
+                "An Archiving Restriction already exists for this model. Please edit the existing one"
+            ),
+        )
+    ]
+
+    model_id = fields.Many2one(
+        "ir.model", string="Model to apply restriction to", required=True, index=True
```
PeterC10, line 30: We don't need to the string so long here, just `Model` will do. We can use this as the help text though. :)

addons/udes_security/models/models.py
```diff
@@ -74,6 +93,15 @@ def _set_fields_readonly(doc, res):
...
+        if not can_archive_record:
+            doc = etree.XML(res["arch"])
+            if doc.tag == "form":
```
PeterC10, line 101: Q: What about the archive option from the tree view when you select some records?

addons/udes_security/models/models.py
PeterC10: Will need to try out this scenario, I'm sure in some cases I've seen Odoo filtering out writes that are writing the existing value, but might be bad memory or wishful thinking on my part...

addons/udes_security/models/res_archiving_restrictions.py
PeterC10: Thinking about this more... maybe just `Permitted Groups` as the string, and then `Groups permitted to archive & unarchive` as the help text?

addons/udes_security/models/models.py
ws-mps: I don't think it does unfortunately. But it doesn't matter, let's keep it simple, if the user cannot change the value shouldn't be in vals, no need to only check restriction if the value is different.

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Can archive if user has _any_ groups defined on the restriction,
+        or no restriction is defined
+
+        :param: model_name: str() model _name i.e "stock.picking.type"
+        :returns: bool()
+        """
+        ResArchivingRestrictions = self.env["res.archiving.restrictions"]
+
+        if self.env.user.id == SUPERUSER_ID:
```
adamrpatrick, line 57: What about debug/ trusted users?

PeterC10: Trusted user is added here: https://github.com/unipartdigital/udes-open/blob/story/540/addons/udes_stock/data/res_archiving_restrictions.xml

But I think this is the wrong way around, should be debug user and not trusted user.

addons/udes_security/models/res_archiving_restrictions.py
PeterAlabaster: Just to confirm messaging:
Frontend (user message):
```
You do not have permission to archive/unarchive this record.
```
Backend  log message:
```
2021-12-21 09:40:17,449 412241 WARNING mvp odoo.addons.udes_security.models.res_archiving_restrictions: User 'bob', ID 6 blocked attempt at archiving on 'stock.picking.type'.
```

addons/udes_security/models/res_archiving_restrictions.py
PeterAlabaster: Yeah, i wondered about this because even I was being inconsistent with naming and `res.archiving.restriction` seemed slightly better, but didn't want to waste time doing it if it was going to be scrapped anyway. Will get it changed (Y)

addons/udes_security/models/res_archiving_restrictions.py
```diff
@@ -0,0 +1,82 @@
...
+            _(
+                "An Archiving Restriction already exists for this model. Please edit the existing one"
+            ),
+        )
+    ]
+
+    model_id = fields.Many2one(
+        "ir.model", string="Model to apply restriction to", required=True, index=True
+    )
+    name = fields.Char(string="Model Name", related="model_id.model", readonly=True)
```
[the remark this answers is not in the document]
PeterAlabaster: Solely used for nice displayname in the breadcrumb

addons/udes_security/models/models.py
PeterAlabaster: That is handled by `cannot_archive_record` that's shoved into res, and picked up in BasicView.js

addons/udes_security/models/res_archiving_restrictions.py
PeterAlabaster: Following principal with rest of Odoo, superuser can do anything, regardless of individual permissions set up.

@PeterC10 is right and trusted user is configured as the only group permitted access to picking type and warehouse, but re-reading the story, it says only debug user, not trusted user.

I'll clarify w/ product whether they want just debug user, or trusted user, or both

addons/udes_security/models/res_archiving_restrictions.py
PeterC10: Yep that looks good. :)

addons/udes_security/models/res_archiving_restrictions.py
PeterAlabaster: Confirmed it is only debug users, so will get it changed in the data file in udes_stock

adamrpatrick (approved):

ws-mps (approved):

adamrpatrick (approved):

PR #543, opened 2021-12-20
Story/247

peterpratt13:


addons/udes_stock/models/stock_picking.py
```diff
@@ -2766,3 +2766,10 @@ def _get_pick_details(self):
             "submitted_value": self.origin,
         }
         return pick_details
+
+    def unassign_users_from_pickings(self, skip_users=None):
```
kdwyer, line 2756: The usage of this method isn't very clear to me - is it the case that it will unassign _all_ users from the picking(s) (that is, `self`), unless they are in `skip_users`?

PeterC10:
Nit: For the delete commit we should prefix this with `[REM]` rather than `[DEL]`.

Reference: https://www.odoo.com/documentation/11.0/reference/guidelines.html#tag-and-module-name

addons/udes_stock/models/stock_picking.py
peterpratt13: Yeah, so basically, i just reused this method in two places. Once where i wanted to remove everyone from inside the pickings, and eveyone ecept those in skip_users. Changed to kwargs now anyway.

adamrpatrick (approved):

addons/udes_stock/models/stock_picking.py
ws-mps: We are already in a picking object, do we need from picking?

kdwyer: Can we explain how it should be called in the docstring?  Replacing `skip_users` with `kwargs` doesn't help the reader understand what arguments the method expects, nor how they will be used.

peterpratt13: Well it does nothing and shouldn't' be called unless udes_kafka_events is installed which i don't mention as that is not in open. Turns out no longer need it as the from tend calls unassign user from the user models. 
But this is the case for the methods with kwargs in res_users - make the frontend happy but to do nothing without udes_kafka-events

PeterAlabaster (approved):

ws-mps (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -2724,20 +2724,6 @@ def mark_empty_pickings(self):
...
             empty_picks.write({"u_mark": False})
 
-    def _unassign_picker(self, **kwargs):
-        """
-        Method to clear the user from pickings.
-        Designed to be overwritten by other more specific modules.
-        """
-        return
-
-    def _assign_picker(self, **kwargs):
```
armandcela, line 2734: I think there are some places in udes_workflows module that call _assign_picker and _unassign_picker.
IF removed it will raise error on udes_workflows , if module udes_kafka_events is not installed as the method is initially override in that module too.

armandcela (approved):

PR #544, opened 2021-12-20
[FIX] udes_stock: Check the package locations when selecting package

kenvac:
The user only finds out about different location in package when tries
to valid the picking the after all pickings been done. These can be time consuming
and more work for team lead to rectify the locations on a picking in case of
locations are different.

### DIFF
--- a/addons/udes_stock/controllers/stock_location.py
+++ b/addons/udes_stock/controllers/stock_location.py
@@ -123,7 +123,12 @@ class Location(UdesApi):
         "/api/stock-location/is_compatible_package/", type="json", methods=["GET"], auth="user"
     )
     def is_compatible_package_to_location(
-        self, package_name, location_id=None, location_name=None, location_barcode=None
+        self,
+        package_name,
+        location_id=None,
+        location_name=None,
+        location_barcode=None,
+        picking_id=None,
     ):
         """
             Checks that the package is in the specified location,
@@ -133,7 +138,8 @@ class Location(UdesApi):
             Refer to the API specs in the README for more details.
         """
         Package = request.env["stock.quant.package"]
-        StockMoveLine = request.env["stock.move.line"]
+        context = dict(request.env.context)
+        context.update(picking_id=picking_id)
 
         location = _get_location(request.env, location_id, location_name, location_barcode)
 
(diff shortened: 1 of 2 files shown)


addons/udes_stock/controllers/stock_location.py
```diff
@@ -138,6 +138,19 @@ def is_compatible_package_to_location(self, package_name,
...
+        package = Package.get_package(package_name, no_results=True)
+
+        if package:
+            pack_location = package.current_picking_move_line_ids.filtered(
+                lambda sml: sml.qty_done > 0
+            ).mapped("location_dest_id")
+            if pack_location and (len(pack_location) > 1 or location != pack_location):
+                raise UserError(
+                    _("You should not put the contents of a package in different locations.")
+                )
```
adamrpatrick, line 155: I'm not sure that this achieves the objective of the story. We want to ensure that damaged stock is not scanned onto a pallet of good stock and vice versa i.e. the issue doesn't arise because we're putting a package into more than one location.

adamrpatrick (changes requested):

addons/udes_stock/controllers/stock_location.py
adamrpatrick: Either way, I think this logic might exist better in `is_compatible_package`

adamrpatrick: Check that `current_picking_move_line_ids` returns the correct move lines i.e. that it filters on the `picking_id` as I don't believe that it's in the context. We may be better off using `get_move_lines` on packages with an extra domain to filter out done move lines.

addons/udes_stock/controllers/stock_location.py
```diff
@@ -123,21 +123,44 @@ def block(self, reason,
...
+        # Usecase for picking returns
+        if package:
+            # Note: Cannot use package.find_move_line or package.get_move_lines method
+            # as they search on package_id. Use of current_picking_move_line_ids
+            # won't work because the context will not have picking_id to compute values.
+            # That leave us with only option of searching stock.move.line on
+            # result_package_id and states
+            sml = StockMoveLine.search(
+                [("result_package_id", "in", package.ids), ("state", "not in", ["done", "cancel"])]
+            )
```
adamrpatrick, line 156: I think for this search to be more robust we need to pass in the picking ID to make extra sure that we're only grabbing the relevant move lines.

adamrpatrick (changes requested):

adamrpatrick (approved):

adamrpatrick:
Please also update the README at addons/udes_stock/README.md

LoKKeR96 (commented):
LGTM, personally I prefer the explanations in READMEs to be written in a use case example or more general fashion but otherwise good work. I would prefer if someone else also had a further look in case I missed anything.

PR #547, opened 2022-01-05
Story/460

kdwyer:


addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+class StockQuantPackage(models.Model):
+    """A package contains one or more quants or subpackages."""
+
+    _inherit = ["stock.quant.package", "mixin.stock.model"]
+    _name = "stock.quant.package"
+    _sql_constraints = [("name_unique", "unique (name)", "This name has already been used")]
+
+    @api.constrains("name")
+    def _check_name(self):
+        warehouse = self.env.ref("stock.warehouse0")
```
ws-mps, line 20: Better to use `User.get_user_warehouse()`  for multi WH systems.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+            # either.
+            pattern = f"{warehouse.u_package_barcode_regex}|{warehouse.u_pallet_barcode_regex}"
+            type_ = "package"
+        if not re.match(pattern, self.name):
+            _logger.debug("%s name %r does not match pattern %r", type_.title(), self.name, pattern)
+            raise ValidationError(_(f"Invalid {type_} name {self.name!r}."))
+
+    def _validate_pallet_name(self, vals):
+        """Raise an exception for names that don't match the pallet barcode regex."""
+        warehouse = self.env.ref("stock.warehouse0")
```
ws-mps, line 39: Same as above.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+            type_ = "pallet"
+        else:
+            # Absent any other information, we don't know if a package is a package
+            # of quants or a package of packages (pallet), so the name may match
+            # either.
+            pattern = f"{warehouse.u_package_barcode_regex}|{warehouse.u_pallet_barcode_regex}"
+            type_ = "package"
+        if not re.match(pattern, self.name):
+            _logger.debug("%s name %r does not match pattern %r", type_.title(), self.name, pattern)
+            raise ValidationError(_(f"Invalid {type_} name {self.name!r}."))
```
ws-mps, line 35: Don't think we want the name in the translation.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _validate_pallet_name(self, vals):
+        """Raise an exception for names that don't match the pallet barcode regex."""
+        warehouse = self.env.ref("stock.warehouse0")
+
+        if "parent_id" in vals:
+            pallet = self.browse(vals["parent_id"])
+            pattern = warehouse.u_pallet_barcode_regex
+            if not re.match(pattern, pallet.name):
+                _logger.debug("Pallet name %r does not match pattern %r", pallet.name, pattern)
```
ws-mps, line 45: Here you can use f string :P

Dockerfile
```diff
@@ -5,19 +5,25 @@ FROM unipartdigital/udes-tester:14.0
 ADD https://codeload.github.com/unipartdigital/odoo-package-hierarchy/zip/${ODOO_VERSION} \
     /opt/odoo-package-hierarchy.zip
 USER root
+RUN echo ${ODOO_VERSION}
```
PeterAlabaster, line 8: Do we need this? Or just makes debugging easier?

kdwyer: No, it was for debugging.  I'll probably drop all these Dckerfile changes and just cherry-pick @adamrpatrick's commit.

addons/udes_advanced_packaging/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "UDES Advanced Packaging",
+    "summary": "udes stock",
+    "description": "Advanced packaging features for UDES- Odoo 14",
```
PeterAlabaster, line 4:
```suggestion
    "description": "Advanced packaging features for UDES - Odoo 14",
```

PeterAlabaster: :laughing:

Dockerfile
kdwyer: Cherry-picked.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+        if not re.match(pattern, self.name):
+            _logger.debug("%s name %r does not match pattern %r", type_.title(), self.name, pattern)
+            raise ValidationError(_(f"Invalid {type_} name {self.name!r}."))
+
+    def _validate_pallet_name(self, vals):
+        """Raise an exception for names that don't match the pallet barcode regex."""
+        warehouse = self.env.ref("stock.warehouse0")
+
+        if "parent_id" in vals:
+            pallet = self.browse(vals["parent_id"])
```
PeterAlabaster, line 42: Had some comments before about needing to define the empty model instead of accessing it via `self` implicitly
```suggestion
    StockQuantPackage = self.env["stock.quant.package"]
            ...
            pallet = StockQuantPackage.browse(vals["parent_id"])
```

addons/udes_advanced_packaging/__manifest__.py
kdwyer: Fixed, albeit without committing the suggestion

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+            pattern = warehouse.u_pallet_barcode_regex
+            if not re.match(pattern, pallet.name):
+                _logger.debug("Pallet name %r does not match pattern %r", pallet.name, pattern)
+                raise ValidationError(_(f"Invalid pallet name {pallet.name!r}."))
+        return
+
+    @api.model
+    def create(self, vals):
+        """Create a new instance of the model."""
```
PeterAlabaster, line 51:
```suggestion
        """Inherit to include pallet name validation as part of record creation."""
```

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,58 @@
...
+        self._validate_pallet_name(vals)
+        return super().create(vals)
+
+    def write(self, vals):
+        """Update the records in this recordset."""
```
PeterAlabaster, line 56:
```suggestion
        """Inherit to include pallet name validation as part of record modification."""
```

PeterAlabaster (commented):
looks good!

addons/udes_advanced_packaging/__manifest__.py
PeterAlabaster: yeah i never like using the commit suggestion thing, it just makes it easier to see what changed :+1:

kdwyer: :100:

addons/udes_advanced_packaging/models/stock_quant_package.py
kdwyer: I'm inclined to use "extend" rather than "inherit" here - presumably it's already being inherited?

addons/udes_advanced_packaging/models/stock_quant_package.py
kdwyer: Fixed, please review.

addons/udes_advanced_packaging/models/stock_quant_package.py
kdwyer: Fixed, please review.

addons/udes_advanced_packaging/models/stock_quant_package.py
kdwyer: How about this, and something similar for `write`?
```python
"""
Create a new instance of the model.

Attempts to create an instance with invalid `name` or values, or parents
which are not pallets will raise an exception.
"""
```
Thinking about it, references to inheritance or extension seem redundant, as Odoo pretty much requires us to call `super()` if we define `create`, unless we want to reimplement `Model.create`.

PeterAlabaster: Sure, sometimes we do need to fully override core odoo functions as they aren't very extensible in a lot of places which was my thinking for inherit vs just saying what the core function does. You're right though and extend is a better word for it.

I see what you're saying though, if we see a super call we know it's extending. I like your suggestion above

addons/udes_advanced_packaging/models/stock_quant_package.py
kdwyer: Fixed, please review

ws-mps (approved):

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,100 @@
...
+    _sql_constraints = [("name_unique", "unique (name)", "This name has already been used")]
+
+    @api.constrains("name")
+    def _check_name(self):
+        User = self.env["res.users"]
+
+        warehouse = User.get_user_warehouse()
+        self.ensure_one()
+        # Prevent renaming pallet to package if pallet contains packages.
+        if self.search_count([("parent_id", "=", self.id)]) > 0:
```
PeterAlabaster, line 35: Sorry I didn't spot this one earlier, but the same thing applies (extracting out `Package = self.env["stock.quant.package"]` and doing `Package.search_count(...)`

kdwyer: Fixed in 7252a2b, please review.

PeterAlabaster (approved):

PR #549, opened 2022-01-05
[FIX] udes_stock: Disable name translation

caleb-shelton:
Disable name translation for picking types and routes,
this avoids complications when duplicating, where the
new duplicated name was not being saved and was being
treated as a translation.

User-story: 621

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -852,37 +852,3 @@ class StockLocation(models.Model):
         """Method to be override in other modules"""
         self.ensure_one()
         return None
-
-
-class Orderpoint(models.Model):
-
-    _inherit = "stock.warehouse.orderpoint"
-
-    @api.onchange("location_id")
-    @api.constrains("location_id")
-    def _is_limited(self):
-        """Prevents creating a second order point on a location.
-
-        If the location or an ancestor is configured to only allow a single
-        order point.
-
-        Raises a ValidationError if the constraint is breached.
-        """
-        self.ensure_one()
-        Orderpoint = self.env["stock.warehouse.orderpoint"]
-        orderpoints = Orderpoint.search(
-            [
-                ("location_id", "=", self.location_id.id),
-            ]
-        )
-        orderpoints -= self
-        if not orderpoints:
-            return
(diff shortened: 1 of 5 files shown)


PeterAlabaster:
I can't remember how Odoo works with overwriting fields, but does this preserve arguments to the fields from core odoo? like the display name `Route Name` from the original:`name = fields.Char('Route Name', required=True)`

caleb-shelton:
> I can't remember how Odoo works with overwriting fields, but does this preserve arguments to the fields from core odoo? like the display name `Route Name` from the original:`name = fields.Char('Route Name', required=True)`

Yes

PeterAlabaster (approved):
yjft

addons/udes_stock/models/stock_location.py
```diff
@@ -853,6 +853,13 @@ def update_scanning_date_start(self):
         self.ensure_one()
         return None
 
+class Route(models.Model):
```
ws-mps, line 856: Can we put this in its own file so we have one model per file, please?

Do you mind moving Orderpoint into its own file too please?

Thanks

addons/udes_stock/models/stock_location.py
```diff
@@ -853,6 +853,13 @@ def update_scanning_date_start(self):
...
+    _name = 'stock.location.route'
```
ws-mps, line 857: No need for this when inheriting.

addons/udes_stock/models/stock_location.py
```diff
@@ -853,6 +853,13 @@ def update_scanning_date_start(self):
...
+    _inherit = 'stock.location.route'
```
ws-mps, line 858: Double quotes

addons/udes_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -0,0 +1,36 @@
+from odoo import fields, models, _, api
+from odoo.exceptions import ValidationError
+
+
+class Orderpoint(models.Model):
+
```
ws-mps, line 6: Remove empty line

ws-mps (approved):

PR #550, opened 2022-01-11
[FIX] udes_transport_management: Fix get transport info

LoKKeR96:
This function should not exist here as it is preparing the data
in a format compatible for workflow. The function should exists
in a workflow controller.

User-story: 494

### DIFF
--- a/addons/udes_transport_management/models/transport.py
+++ b/addons/udes_transport_management/models/transport.py
@@ -1,29 +1,25 @@
 # -*- coding: utf-8 -*-
-
+from odoo.tools.translate import _
 from odoo import api, fields, models
 
 
 class Transport(models.Model):
-    _name = 'udes_transport_management.transport'
-    _description = 'UDES Transport Info'
+    _name = "udes_transport_management.transport"
+    _description = "UDES Transport Info"
 
-    vehicle_sequence = fields.Integer(string='Vehicle Sequence Number',
-                                      required=False, copy=False)
-    vehicle_description = fields.Char(string='Vehicle Description',
-                                      required=False, copy=False)
-    license_plate = fields.Char(string='Vehicle Registration',
-                                required=False, copy=False)
-    driver_name = fields.Char(string='Driver Name',
-                              required=False, copy=False)
+    vehicle_sequence = fields.Integer(string="Vehicle Sequence Number", required=False, copy=False)
+    vehicle_description = fields.Char(string="Vehicle Description", required=False, copy=False)
+    license_plate = fields.Char(string="Vehicle Registration", required=False, copy=False)
+    driver_name = fields.Char(string="Driver Name", required=False, copy=False)
 
-    picking_id = fields.Many2one('stock.picking',
-                                 'Transfer',
-                                 required=True,
-                                 index=True,
-                                 copy=False)
+    picking_id = fields.Many2one("stock.picking", "Transfer", required=True, index=True, copy=False)
 
     _sql_constraints = [
-        ('picking_id_uniq', 'unique(picking_id)', 'Only one transport information is allowed per picking.'),
+        (
+            "picking_id_uniq",
+            "unique(picking_id)",
+            "Only one transport information is allowed per picking.",
+        )
     ]
 
     def _update_picking_data(self):
@@ -43,8 +39,12 @@ class Transport(models.Model):
     @api.multi
     def _prepare_info(self, **kwargs):
         # Create a dict of all the fields
-        data = {'vehicle_sequence': self.vehicle_sequence, 'vehicle_description': self.vehicle_description,
-                'license_plate': self.license_plate, 'driver_name': self.driver_name}
+        data = {
+            "vehicle_sequence": self.vehicle_sequence,
+            "vehicle_description": self.vehicle_description,
+            "license_plate": self.license_plate,
+            "driver_name": self.driver_name,
+        }
 
         return data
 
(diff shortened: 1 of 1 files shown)


addons/udes_transport_management/models/transport.py
```diff
@@ -61,7 +61,8 @@ def get_info(self, **kwargs):
     def prepare_transport_info(self, **kwargs):
         """Prepare transport info method that can be override in order
         to be able to extend it in other modules"""
-        self.ensure_one()
+        # Not calling self.ensure_one() in case the record does not exist which
```
armandcela, line 64: For now it makes sense the comment, but in future i dont think that will be needed.

Suggestion: Remove the comment and:
`result = []`
`for transport in self:`
    `transport_info = {
                "label": "Trailer Number",
                "value": transport.vehicle_sequence or None,
                "lookup_key": "trailer_number",
            },
            {
                "label": "Trailer Unit ID",
                "value": transport.vehicle_description or None,
                "lookup_key": "trailer_unit_id",
            },
            {
                "label": "Vehicle Registration",
                "value": transport.license_plate or None,
                "lookup_key": "vehicle_registration",
            },
            {
                "label": "Driver Name",
                "value": transport.driver_name or None,
                "lookup_key": "driver_name",
            }`
    `result.append(transport_info)`
`return result`

LoKKeR96: I wrote it so that in the future we won't forget and add it back again for any reason. Would it be better removing it you think?

LoKKeR96: I addressed your suggestion in the code. I could not simply copy your code as it wouldn't return default values when no transport record is available.

armandcela (approved):

addons/udes_transport_management/models/transport.py
```diff
@@ -59,29 +59,38 @@ def get_info(self, **kwargs):
...
+            )
+        result = []
+        for transport in self:
+            transport_info = blank_transport.copy()
+            transport_info["trailer_number"].value = transport.vehicle_sequence
+            transport_info["trailer_unit_id"].value = transport.vehicle_description
+            transport_info["vehicle_registration"].value = transport.license_plate
+            transport_info["driver_name"].value = transport.driver_name
+            result.append(transport_info)
+        if not len(result):
```
peterpratt13, line 94: nit: if not result?

addons/udes_transport_management/models/transport.py
```diff
@@ -59,29 +59,38 @@ def get_info(self, **kwargs):
...
+                },
+                {
+                    "label": _("Driver Name"),
+                    "value": None,
+                    "lookup_key": "driver_name",
+                },
+            )
+        result = []
+        for transport in self:
+            transport_info = blank_transport.copy()
```
peterpratt13, line 88: Ignore this if i have missed something,

copying a tuple is this a method in odoo, or some other module?
I don't understand how this gets updated?
We have a tuple of dicts, then we pass the key to the tuple and it chooses the relevant dict and sets the value. 

We could do 
lambda transport: (
            {
                "label": _("Trailer Number"),
                "value": transport.vehicle_sequence or None,
                "lookup_key": "trailer_number",
            },
            {
                "label": _("Trailer Unit ID"),
                "value": transport.vehicle_description or None,
                "lookup_key": "trailer_unit_id",
            },
            {
                "label": _("Vehicle Registration"),
                "value": transport.license_plate or None,
                "lookup_key": "vehicle_registration",
            },
            {
                "label": _("Driver Name"),
                "value": transport.driver_name or None,
                "lookup_key": "driver_name",
            },
        )

Then do something like result =[t(transport) for transport in self]?

Their maybe something nicer. 

Could also consider updating the unit tests as they don't currently run

addons/udes_transport_management/models/transport.py
LoKKeR96: I just realised that's not really clear but it's basically saying If not 0 then. Since 0 would interpreted as false. Would you prefer me to make it more explicit?

peterpratt13: i think if not result would be ok, no need for len

ws-mps: Agree, no need for len, "not []" is commonly used.

addons/udes_transport_management/models/transport.py
LoKKeR96: About the tuple, I got confused and thought I could access it like in workflows. I am converting it into a list of dictionaries now. About the approach you suggested, it seems nice but I don't see how having a lambda can be used with no records in self if we need to pass a record to lambda in order to create a transport tuple or list with None values. Basically, I don't see a better way of doing this because lambda cannot be used for the empty case.

addons/udes_transport_management/models/transport.py
```diff
@@ -59,29 +59,38 @@ def get_info(self, **kwargs):
...
-                "lookup_key": "vehicle_registration",
-            },
-            {
-                "label": "Driver Name",
-                "value": self.driver_name or None,
-                "lookup_key": "driver_name",
-            }
-        ]
-        return transport_info
+        blank_transport = (
```
ws-mps, line 64: I think we are overcomplicating this method, let's do it this way:
```
transport_info = []
if self:
    self.ensure_one()
    transport_info = [XXX]
return result
```

addons/udes_transport_management/models/transport.py
```diff
@@ -59,29 +59,38 @@ def get_info(self, **kwargs):
         return res
 
     def prepare_transport_info(self, **kwargs):
```
ws-mps, line 61: This function will be removed, since it does not belong to this module.

LoKKeR96: @ws-mps All done

ws-mps (approved):

PR #551, opened 2022-01-11
[IMP] udes_open: Handle UoM in udes_stock

peterpratt13:
Use product_uom_qty throughout as product_qty is a computed
field. This is a backend change, and further changes need
to be done to handle different UoMs in both the frontend
and in Odoo.

User-story: 462

### DIFF
--- a/addons/udes_stock/README.md
+++ b/addons/udes_stock/README.md
@@ -85,6 +85,8 @@ Additional Details:
 ### Quants (model: stock.quant)
 
 Physical instances of products at a location are modelled as quants. Short of "quantities of stock", these are used to record stock levels are various parts of the warehouse. Using the analogy of object-orientated programming, products = classes, quants = objects.
+Note that quants are stored in their unit of measure, so if screws come in boxes of 1000 and we have 5000 in a location,
+then we would have `available_quantity = 5`.
 
 Additional Details:
 - Change _description to 'Quant'
@@ -128,8 +130,8 @@ This is essentially a collection of products that are to be moved from one locat
 | u_prev_picking_ids | one2many (stock.picking) | Shows the previous pickings in the chain |
 | u_next_picking_ids | one2many (stock.picking) | Shows the next pickings in the chain |
 | u_created_backorder_ids | one2many (stock.picking) | Shows backorders created from this picking |
-| u_quantity_done | float | Sums all done move quantities for the picking |
-| u_total_quantity | float | Sums all move quantities for the picking |
+| u_quantity_done | float | Sums all done move quantities for the picking (in the UoM of the moves) |
+| u_total_quantity | float | Sums all move quantities for the picking (in the UoM of the moves) |
 | u_has_discrepancies | boolean | Flags when u_quantity_done doesn't match u_total_quantity |
 | u_num_pallets | int | Sums all unique destination packages found in the picking's move lines |
 
@@ -158,6 +160,17 @@ The type of stock.picking can is defined by this type. It can represent a goods 
 ### Stock Move (model: stock.move)
 
 A move of an item of stock from one location to another.
+Note:
+* `product_qty`: the quantity of the product in the move in the UoM of the product (computed field)
+* `product_uom_qty`: the quantity of the product in the move in the UoM of the move
+Example, screws come in a box of 1000. The UoM of the product is 1000.
+If you wanted a move to always move 6 boxes at a time, the UoM of the move is 6000.
+If you wanted to move 12 boxes in one move, then:
+* product_qty = 12
+* product_uom_qty = 2
+so the full/real/absolute quantity (whatever we refer to the 12000 screws as) is never considered in the move.
+
+When using/creating/splitting moves, we use the UoM of the move. 
 
 | Helpers | Description |
 | ------- | ----------- |
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -43,13 +43,13 @@ class StockPicking(models.Model):
         "Quantity Done",
         compute="_compute_picking_quantities",
         digits=(0, 0),
-        help="Quantity done of the moves related to the picking",
+        help="Quantity done of the moves related to the picking in the UoM of the move(s)",
     )
     u_total_quantity = fields.Float(
         "Total Quantity",
         compute="_compute_picking_quantities",
         digits=(0, 0),
-        help="Total quantity todo of the moves related to the picking",
+        help="Total quantity todo of the moves related to the picking in the UoM of the move(s)",
     )
     u_has_discrepancies = fields.Boolean(
         "Has Discrepancies",
@@ -116,6 +116,13 @@ class StockPicking(models.Model):
             has_discrepancies = False
 
             for move in picking.move_lines.filtered(lambda ml: ml.state != "cancel"):
+                # The move quantity done is in the unit of measure of the move not the product
+                # So use the move UoM not the product for what is done and what is left to do.
+                # This avoids the issue where it says left to do is different to the number of
+                # boxes.
+                # NOTE: Different UoMs across a picking could make these computed fields unreliable
+                # in terms of how much work is done/left todo. Example 50% done - moved one crane
+                # but 1000000 screws in one box.
                 qty_done = move.quantity_done
                 qty_todo = move.product_uom_qty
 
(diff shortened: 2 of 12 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -122,8 +126,9 @@ def _compute_picking_quantities(self):
                 if not has_discrepancies and qty_done != qty_todo:
                     has_discrepancies = True
 
-                total_qty_done += qty_done
-                total_qty_todo += qty_todo
+                factor = move.product_id.uom_id.factor
```
LoKKeR96, line 129: If a package has 6 products, is the factor 1/6? Or 6? I'd imagine it has to be 1/6 if the u_total_quantity wants to be not in UOM, right?

peterpratt13: good spot

addons/udes_stock/models/stock_picking.py
```diff
@@ -122,8 +122,9 @@ def _compute_picking_quantities(self):
                 if not has_discrepancies and qty_done != qty_todo:
                     has_discrepancies = True
 
-                total_qty_done += qty_done
-                total_qty_todo += qty_todo
+                factor = move.product_uom.factor
```
adamrpatrick, line 125: Can you add a comment here to give an example of the relation between `factor` units of measure?

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1,31 +1,33 @@
...
+        products_info = [{"product": cls.apple, "uom_qty": 10}]
+        # cls.test_picking_in = cls.create_picking(
+        #     cls.picking_type_goods_in,
+        #     products_info=products_info,
+        #     confirm=True,
+        #     location_dest_id=cls.test_received_location_01.id,
+        # )
+        # cls.test_picking_pick = cls.create_picking(
+        #     cls.picking_type_pick, products_info=products_info, confirm=True
+        # )
```
adamrpatrick, line 22: Does this still need to be commented out?

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -66,535 +68,642 @@ def _assert_picking_related_pickings_match_expected_values(self, pickings, expec
...
 
-        # Empty locations not limited by default
-        self.assertGreater(len(self.test_picking_in.get_empty_locations(limit=None)), 1)
+# class TestStockPicking(TestStockPickingCommon):
+#     def test_get_empty_locations(self):
+#         """Get empty locations - for goods in"""
+#         self.assertEqual(self.test_picking_in.get_empty_locations(), self.test_received_location_01)
+#         # Add stock to a location - to check empty locations obtained
+#         self.create_quant(self.apple.id, self.test_received_location_01.id, 5)
+#         self.assertFalse(self.test_picking_in.get_empty_locations())
```
adamrpatrick, line 78: Same here, do these still need to be commented out?

adamrpatrick (changes requested):

adamrpatrick:
There are quite a few unit tests deleted in udes_stock/tests/test_stock_picking.py. I presume that these need adding back in?

addons/udes_stock/models/stock_picking.py
PeterAlabaster: I had a call w/ @peterpratt13 about this, because I wasn't sure I was understanding what was happening here.

So odoo calculates on the move under `product_qty` the quantity _in the uom of the product_
e.g
```
Product: cherry, UoM: Dozen (12)
Move: for 5x cherry, using UoM (on the move): Half a dozen (6)
product_qty will be 2.5 (computed to reflect the quantity in the uom set on the product)
```
this goes in-line with what @ws-mps was saying about the quants being stored in the uom of the product, so odoo flattens it to something common (products uom) and then it can push that to the quant.

So i think this factor calculation can be removed and instead we can just use `move.product_qty`, and `move.quantity_done` without converting things again

peterpratt13: yeah - so basically , need to change a bit because everything is Uom, missed the Real Quantity field as it is the Real Quantity in that UoM, different to the product uom. But everything is UoM not some

peterpratt13: So this (u_total_quanitiy_done in a picking might read 3, 1 crane, 2 boxxes of 100000000 screws.

PeterAlabaster (changes requested):
will approve once what was discussed on the call is sorted :+1: rest lgtm

addons/udes_stock/models/stock_picking.py
```diff
@@ -116,15 +116,21 @@ def _compute_picking_quantities(self):
             has_discrepancies = False
 
             for move in picking.move_lines.filtered(lambda ml: ml.state != "cancel"):
+                # The move quantity done is in the unit of measure of the move not the product
+                # So use the move UoM not the product for what is done and what is left to do.
+                # This avoids the issue where it says left to do is different to the number of
+                # boxes.
+                # NOTE: Different UoMs across a picking could make these computed fields unreliable
+                # in terms of how much work is done/left todo. Example 50% done - moved one crane
+                # but 1000000 screws in one box.
```
PeterAlabaster, line 125: this note will be useful in 2 years when clients come across this exact scenario and wonder why bob got lumped with moving 100000 screws and it only counted as 1 :laughing:

addons/udes_stock/tests/test_stock_quant.py
```diff
@@ -232,6 +234,9 @@ def test_fifo_without_nones(self):
         self.assertEqual(reserved_quant.package_id, self.pack2)
         self.assertEqual(reserved_quant, self.quant2)
 
+    @unittest.skip(
+        "Skp unittest as the _gather does not correctly return the sorted quants based on fifo when package ids have a mix of nulls"
+    )
```
PeterAlabaster, line 239: ideally i think we should create an issue for that as it could bite us in the ass later, and then ref the issue in this skip string

PeterAlabaster: oh wait the issue already exists https://taiga.unipart.digital/project/udes-x/issue/782
can we just ref this one and update the issue with what you found?

PeterAlabaster (approved):

addons/udes_stock/README.md
```diff
@@ -158,6 +160,17 @@ The type of stock.picking can is defined by this type. It can represent a goods
 ### Stock Move (model: stock.move)
 
 A move of an item of stock from one location to another.
+Note:
+* `product_qty`: the quantity of the product in the move in the UoM of the product (computed field)
+* `product_uom_qty`: the quantity of the product in the move in the UoM of the move
+Example, screws come in a bo of 1000. The UoM of the product is 1000.
```
adamrpatrick, line 166: Typo: "box" not "bo"

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -598,3 +590,161 @@ def test_assert_picking_quantities_computed_correctly(self):
...
+
+class TestStockPickingUoM(TestStockPickingCommon):
+    @classmethod
+    def setUpClass(cls):
+        super(TestStockPickingUoM, cls).setUpClass()
+        Uom = cls.env["uom.uom"]
+        unit = cls.env.ref("uom.product_uom_categ_unit")
+        cls.half_dozen = Uom.create(
+            {
+                "name": "HalfDozzen",
```
adamrpatrick, line 603: Typo: "Dozen" instead of "Dozzen"

addons/udes_stock/tests/test_stock_quant.py
```diff
@@ -226,40 +228,46 @@ def test_fifo_without_nones(self):
...
 
-        # Should choose between quant2 and quant3 based on `in_date`
-        # Choose quant2 as it has a package_id
+        # Should choose between quant2 and quant3 based on `in_date`.
+        # Choose quant2 as it has a smaller package id.
         self.assertEqual(reserved_quant.in_date, oldest_time)
         self.assertEqual(reserved_quant.package_id, self.pack2)
         self.assertEqual(reserved_quant, self.quant2)
 
+    @unittest.skip("Skp unittest due to Issue 782")
```
ws-mps, line 237: Typo: Skp

adamrpatrick (approved):

PR #554, opened 2022-01-18
[ADD] udes_advanced_packaging: Packagisation port/improve...

caleb-shelton:
Improve full pack reservation by removing extra whole package quantities
when unreserving.

User-story: 459

### DIFF
--- a/addons/udes_advanced_packaging/models/stock_move.py
+++ b/addons/udes_advanced_packaging/models/stock_move.py
@@ -1,0 +1,12 @@
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_assign(self):
+        """Extend _action_assign to trigger full package reservation"""
+        res = super(StockMove, self)._action_assign()
+
+        self.picking_id._reserve_full_packages()
+        return res
--- a/addons/udes_advanced_packaging/models/stock_move_line.py
+++ b/addons/udes_advanced_packaging/models/stock_move_line.py
@@ -1,0 +1,14 @@
+from odoo import models
+from collections import defaultdict
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _get_all_products_quantities(self):
+        """This function computes the different product quantities for the given move_lines"""
+        # TODO: Issue 962, make this function work with different UoMs
+        res = defaultdict(int)
+        for move_line in self:
+            res[move_line.product_id] += move_line.product_uom_qty
+        return res
--- a/addons/udes_advanced_packaging/models/stock_picking.py
+++ b/addons/udes_advanced_packaging/models/stock_picking.py
@@ -1,0 +1,70 @@
+from collections import defaultdict
+
+from odoo import models, _
+from odoo.exceptions import ValidationError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _should_reserve_full_packages(self):
+        """Method to determine if picking should reserve entire packages"""
+        self.picking_type_id.ensure_one()
+        return self.picking_type_id.u_reserve_as_packages
+
+    def _reserve_full_packages(self):
+        """
+        If the picking type of the pickings in self has full package
+        reservation enabled, whole packages will be reserved.
+        """
+
+        # do not reserve full packages when bypass_reserve_full packages
+        # is set in the context as True
+        if not self.env.context.get("bypass_reserve_full_packages"):
+            pickings = self.filtered(lambda p: p._should_reserve_full_packages())
+            pickings.move_line_ids.package_id
+
+            for picking in pickings:
+                remaining_qtys = defaultdict(int)
+
+                # get all packages
+                packages = picking.move_line_ids.package_id
+                for package in packages:
+                    move_lines = picking.move_line_ids.filtered(lambda ml: ml.package_id == package)
+                    pack_products = frozenset(package._get_all_products_quantities().items())
+                    mls_products = frozenset(move_lines._get_all_products_quantities().items())
+                    if pack_products != mls_products:
+                        # move_lines do not match the quants
+                        pack_mls = package._get_current_move_lines()
+                        other_pickings = pack_mls.picking_id - picking
+                        if other_pickings:
+                            raise ValidationError(
+                                _("The package is reserved in other pickings: %s")
+                                % ",".join(other_pickings.mapped("name"))
+                            )
+
+                        quants = package._get_contained_quants()
+                        for product, qty in quants.get_quantities_by_key(
+                            only_available=True
+                        ).items():
+                            remaining_qtys[product] += qty
+                if remaining_qtys:
+
+                    products_info = []
+                    for product, qty in remaining_qtys.items():
+                        products_info.append({"product": product, "uom_qty": qty})
+
+                    # TODO: Issue 962, may need to adjust for it to work with different UoMs
+
+                    move_values = self._prepare_move(
+                        picking, [products_info], u_uom_initial_demand=0
+                    )
+
+                    picking._create_move(move_values)
+
+                    # add bypass_zero_qty_log_message context to avoid temporary moves being logged
+                    picking.with_context(bypass_zero_qty_log_message=True).action_confirm()
+
+                    # add bypass_reserve_full_packages at the context
+                    # to avoid to be called again inside _create_move()
+                    picking.with_context(bypass_reserve_full_packages=True).action_assign()
--- a/addons/udes_advanced_packaging/models/stock_picking_type.py
+++ b/addons/udes_advanced_packaging/models/stock_picking_type.py
@@ -1,0 +1,11 @@
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    u_reserve_as_packages = fields.Boolean(
+        string="Reserve Entire Packages",
+        default=False,
+        help="Flag to indicate reservations should be rounded up to entire packages.",
+    )
--- a/addons/udes_advanced_packaging/models/stock_quant_package.py
+++ b/addons/udes_advanced_packaging/models/stock_quant_package.py
@@ -101,3 +101,34 @@ class StockQuantPackage(models.Model):
         """
         self.validate_parent_name(vals)
         return super().write(vals)
+
+    def _get_all_products_quantities(self):
+        """This function computes the different product quantities for the given package"""
+        # TODO: Issue 962, make this work with different UoMs
+        res = {}
+        for quant in self._get_contained_quants():
+            if quant.product_id not in res:
+                res[quant.product_id] = 0
+            res[quant.product_id] += quant.quantity
+        return res
+
+    def assert_reserved_full_package(self, move_lines):
+        """Check that a package is fully reserved at move_lines."""
+        self.ensure_one()
+
+        pack_products = frozenset(self._get_all_products_quantities().items())
+        mls_products = frozenset(move_lines._get_all_products_quantities().items())
+        if pack_products != mls_products:
+            # move_lines do not match the quants
+            picking = move_lines.picking_id
+            picking.ensure_one()
+            pack_mls = self._get_current_move_lines()
+            other_pickings = pack_mls.picking_id - picking
+            if other_pickings:
+                raise ValidationError(
+                    _("The package is reserved in other pickings: %s")
+                    % ",".join(other_pickings.mapped("name"))
+                )
+            raise ValidationError(
+                _("Cannot mark partially reserved package %s as done.") % self.name
+            )
--- a/addons/udes_advanced_packaging/tests/test_package_reservation.py
+++ b/addons/udes_advanced_packaging/tests/test_package_reservation.py
@@ -1,0 +1,221 @@
+from odoo.exceptions import ValidationError
+from odoo.addons.udes_stock.tests.common import BaseUDES
+
+
+class TestPackageReservation(BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestPackageReservation, cls).setUpClass()
+        # enable full package reservation
+        cls.picking_type_pick.u_reserve_as_packages = True
+
+        cls.product_info = [{"product": cls.apple, "uom_qty": 1}]
+
+    def test_reserve_full_package_one_product(self):
+        """
+        Test that the a package is fully reserved,
+        even though a picking is requested a lower quantity
+        """
+        # create a package
+        package = self.create_package()
+
+        # create a quant of 4 apples in a stock sublocation for the package
+        quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=package.id
+        )
+        # the quant should be unreserved
+        self.assertEqual(quant.reserved_quantity, 0)
+
+        # create a Pick for 1 apple
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=self.product_info, confirm=True, assign=False
+        )
+
+        picking.action_assign()
+
+        move_count_pre_assign = len(picking.move_lines)
+
+        # the quant should be fully reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.move_line_ids
+        self.assertEqual(apple_mls.package_id, package)
+
+        # the apple move should still be present after unreserving
+        picking.do_unreserve()
+        move_count_post_unreserve = len(picking.move_lines)
+        self.assertEqual(move_count_pre_assign, move_count_post_unreserve)
+
+    def test_reserve_full_package_two_products(self):
+        """
+        Test to reserve a full package for one product
+        when initial demand < package quantity
+        """
+
+        # create a package
+        package = self.create_package()
+
+        # create a quant of 3 apples in a stock sublocation for the package
+        apple_quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 3, package_id=package.id
+        )
+        strawberry_quant_1 = self.create_quant(
+            self.strawberry.id,
+            self.test_stock_location_01.id,
+            1,
+            package_id=package.id,
+            lot_name="sn01",
+        )
+        strawberry_quant_2 = self.create_quant(
+            self.strawberry.id,
+            self.test_stock_location_01.id,
+            1,
+            package_id=package.id,
+            lot_name="sn02",
+        )
+        all_quants = apple_quant + strawberry_quant_1 + strawberry_quant_2
+
+        # all quants should be unreserved
+        self.assertTrue(
+            all([q.reserved_quantity == 0 for q in all_quants]), "One or more quants are reserved"
+        )
+
+        for q in all_quants:
+            with self.subTest(quant_id=q.id):
+                self.assertEqual(q.reserved_quantity, 0)
+
+        # create a Pick for 1 apple
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=self.product_info, confirm=True, assign=False
+        )
+
+        move_count_pre_assign = len(picking.move_lines)
+        picking.action_assign()
+
+        # apple quant should be fully reserved
+        self.assertEqual(apple_quant.reserved_quantity, 3)
+        # strawberry quants should be fully reserved
+        self.assertEqual(strawberry_quant_1.reserved_quantity, 1)
+        self.assertEqual(strawberry_quant_2.reserved_quantity, 1)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.move_line_ids.filtered(lambda ml: ml.product_id == self.apple)
+        self.assertEqual(apple_mls.package_id, package)
+
+        # strawberry stock.move.lines of the picking should have the package
+        strawberry_mls = picking.move_line_ids.filtered(lambda ml: ml.product_id == self.strawberry)
+        self.assertEqual(strawberry_mls.package_id, package)
+
+        # picking should have only the original removes remaining after unreserving,
+        # as extra products from full package reservation should have been removed
+        picking.do_unreserve()
+        move_count_post_unreserve = len(picking.move_lines)
+        self.assertEqual(move_count_pre_assign, move_count_post_unreserve)
+
+    def test_partial_reservation_package_one_still_functional(self):
+        """
+        Test to reserve a full package for one product
+        when initial demand < package quantity
+        """
+
+        # disable full package reservation
+        self.picking_type_pick.u_reserve_as_packages = False
+
+        # create a package
+        package = self.create_package()
+
+        # create a quant of 4 apples in a stock sublocation for the package
+        quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=package.id
+        )
+        # the quant should be unreserved
+        self.assertEqual(quant.reserved_quantity, 0, "The quant is reserved")
+
+        # create a Pick for 1 apple
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=self.product_info, confirm=True, assign=True
+        )
+
+        # the quant should be partially reserved
+        self.assertEqual(quant.reserved_quantity, 1)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.move_line_ids.filtered(lambda ml: ml.product_id == self.apple)
+        self.assertEqual(apple_mls.package_id, package)
+
+        # update_picking with package_name = package.name should raise an error
+        with self.assertRaises(ValidationError) as e:
+            package.assert_reserved_full_package(apple_mls)
+        self.assertEqual(
+            e.exception.args[0], f"Cannot mark partially reserved package {package.name} as done."
+        )
+
+    def test_multiple_pickings_do_not_block_each_other_doing_full_package_reservation(self):
+        """Test the case when there is more than one picking in the set being reserved."""
+
+        # create stock of 2 products, in package, to reserve.
+        apple_pack = self.create_package()
+        apple_quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=apple_pack.id
+        )
+
+        banana_pack = self.create_package()
+        banana_quant = self.create_quant(
+            self.banana.id, self.test_stock_location_01.id, 4, package_id=banana_pack.id
+        )
+        self.assertEqual(apple_quant.reserved_quantity, 0)
+        self.assertEqual(banana_quant.reserved_quantity, 0)
+
+        # create two picks for 1 of each product
+        self.create_picking(
+            self.picking_type_pick,
+            products_info=[{"product": self.apple, "uom_qty": 1}],
+            confirm=True,
+            assign=True,
+        )
+        self.create_picking(
+            self.picking_type_pick,
+            products_info=[{"product": self.banana, "uom_qty": 1}],
+            confirm=True,
+            assign=True,
+        )
+
+        # both quants should be fully reserved
+        self.assertEqual(apple_quant.reserved_quantity, 4)
+        self.assertEqual(banana_quant.reserved_quantity, 4)
+
+    def test_initial_demand_and_product_quantity(self):
+        """
+        Test the initial demand is set to the first product quantity value.
+        When unreserving test that the extra quantity added by full package reservation
+        is removed from the product quantity.
+        """
+
+        # create stock of 1 product, in package, to reserve.
+        apple_pack = self.create_package()
+        self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=apple_pack.id
+        )
+
+        # create a pick for 1 apple
+        pick = self.create_picking(
+            self.picking_type_pick,
+            products_info=[{"product": self.apple, "uom_qty": 1}],
+            confirm=True,
+            assign=False,
+        )
+
+        move = pick.move_lines
+        # initial demand should match product quantity.
+        self.assertEqual(move.u_uom_initial_demand, move.product_uom_qty)
+
+        pick.action_assign()
+
+        # initial demand should not match product quantity due to full package reservation.
+        self.assertNotEqual(move.u_uom_initial_demand, move.product_uom_qty)
+
+        pick.do_unreserve()
+
+        # product quantity should revert back to initial demand.
+        self.assertEqual(move.u_uom_initial_demand, move.product_uom_qty)
--- a/addons/udes_advanced_packaging/views/stock_picking_type.xml
+++ b/addons/udes_advanced_packaging/views/stock_picking_type.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_picking_type_form_inherit" model="ir.ui.view">
+        <field name="name">stock.picking.type.form.udes_advanced_packaging</field>
+        <field name="model">stock.picking.type</field>
+        <field name="inherit_id" ref="stock.view_picking_type_form"/>
+        <field name="arch" type="xml">
+
+            <!-- Add udes fields -->
+            <xpath expr="//form/sheet/group[last()]" position="after">
+                <group string="Advanced Packaging" groups='base.group_no_one'>
+                    <field name="u_reserve_as_packages" />
+                </group>
+            </xpath>
+        </field>
+    </record>
+
+</odoo>
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -1,9 +1,37 @@
 # -*- coding: utf-8 -*-
-from odoo import models, api, _
+from odoo import models, fields, api, _
+from odoo.addons import decimal_precision as dp
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
+
+    u_uom_initial_demand = fields.Float(
+        "Initial Demand",
+        digits="Product Unit of Measure",
+        help="The original quantity when the move was created",
+    )
+
+    # Override product_uom_qty to change string from "Demand" to "Quantity", because of
+    # how we are now tracking initial demand in u_uom_initial_demand.
+    product_uom_qty = fields.Float(string="Quantity")
+
+    @api.model
+    def create(self, vals):
+        """Extend create to set the initial demand to product_uom_qty if no value set"""
+        if "u_uom_initial_demand" not in vals:
+            vals["u_uom_initial_demand"] = vals.get("product_uom_qty")
+        return super().create(vals)
+
+    def _do_unreserve(self):
+        """
+        Extend _do_unreserve to remove extra quantities added to initial_demand when
+        unreserving a package.
+        """
+        res = super()._do_unreserve()
+        for move in self:
+            move.write({"product_uom_qty": move.u_uom_initial_demand})
+        return res
 
     def _prepare_move_line(self, move, uom_qty, uom_id=None, **kwargs):
         """
@@ -96,14 +124,10 @@ class StockMove(models.Model):
         """
         self.ensure_one()
         if not all(ml.move_id == self for ml in move_lines):
-            raise ValueError(
-                _("Cannot split move lines from a move they are not part of.")
-            )
+            raise ValueError(_("Cannot split move lines from a move they are not part of."))
         if (
             move_lines == self.move_line_ids
-            and not self.move_orig_ids.filtered(
-                lambda m: m.state not in ("done", "cancel")
-            )
+            and not self.move_orig_ids.filtered(lambda m: m.state not in ("done", "cancel"))
             and not self.state == "partially_available"
         ):
             new_move = self
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -17,6 +17,16 @@ class StockMoveLine(models.Model):
         readonly=True,
     )
 
+    def _log_message(self, record, move, template, vals):
+        """
+        Extend _log_message to stop messages being created
+        when bypass_zero_qty_log_message context is set.
+        """
+        if self.env.context.get("bypass_zero_qty_log_message"):
+            if "product_uom_qty" in vals and not vals["product_uom_qty"]:
+                return
+        return super()._log_message(record, move, template, vals)
+
     def get_lines_incomplete(self):
         """Return the move lines in self that are not completed,
         i.e., quantity done < quantity to do
--- a/addons/udes_stock/views/stock_picking.xml
+++ b/addons/udes_stock/views/stock_picking.xml
@@ -80,6 +80,17 @@
                 </group>
             </xpath>
 
+            <!-- Add in u_uom_initial_demand field -->
+            <xpath expr="//field[@name='move_ids_without_package']/tree/field[@name='product_uom_qty']" position="after">
+                <field name="u_uom_initial_demand" readonly="1"/>
+            </xpath>
+
+            <!-- Remove the explicit string attribute so it uses the field's Python string value which is:
+                 "Demand" -> "Quantity" -->
+            <xpath expr="//field[@name='move_ids_without_package']/tree/field[@name='product_uom_qty']" position="attributes">
+                <attribute name="string"/>
+            </xpath>
+
         </field>
     </record>
 
--- a/addons/udes_advanced_packaging/__manifest__.py
+++ b/addons/udes_advanced_packaging/__manifest__.py
@@ -6,6 +6,8 @@
     "website": "http://github/unipartdigital/udes-open",
     "category": "UDES",
     "version": "0.1",
-    "depends": ["package_hierarchy", "udes_stock"],
-    "data": [],
+    "depends": ["uom", "stock", "package_hierarchy", "udes_stock"],
+    "data": [
+        "views/stock_picking_type.xml",
+    ],
 }
--- a/addons/udes_advanced_packaging/models/__init__.py
+++ b/addons/udes_advanced_packaging/models/__init__.py
@@ -2,3 +2,7 @@
 
 from . import stock_quant_package  # noqa: F401
 from . import stock_warehouse  # noqa: F401
+from . import stock_move
+from . import stock_move_line
+from . import stock_picking
+from . import stock_picking_type
--- a/addons/udes_advanced_packaging/tests/__init__.py
+++ b/addons/udes_advanced_packaging/tests/__init__.py
@@ -1,3 +1,3 @@
 """Tests for the UDES Advanced Packaging module."""
-
 from . import test_stock_quant_package  # noqa: F401
+from . import test_package_reservation
--- a/addons/udes_advanced_packaging/tests/test_stock_quant_package.py
+++ b/addons/udes_advanced_packaging/tests/test_stock_quant_package.py
@@ -1,6 +1,7 @@
 """Tests for the Package model."""
 
 from odoo.exceptions import ValidationError
+from odoo.tools import mute_logger
 from psycopg2.errors import UniqueViolation
 
 from odoo.addons.udes_stock.tests.common import BaseUDES
@@ -69,7 +70,7 @@ class PackageCreationTestCase(BaseUDES):
         # Odoo converts the UniqueViolation to a ValidationError, but this
         # hasn't happened "yet" see how PGERROR_TO_OE is handled in
         # BaseModel.load.
-        with self.assertRaises(UniqueViolation):
+        with self.assertRaises(UniqueViolation), mute_logger("odoo.sql_db"):
             self.package.create({"name": name})
 
     def test_prevents_using_package_as_pallet(self):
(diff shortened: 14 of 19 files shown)


addons/udes_advanced_packaging/models/stock_move_line.py
```diff
@@ -0,0 +1,14 @@
+from odoo import models
+from collections import defaultdict
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _get_all_products_quantities(self):
```
peterpratt13, line 8: If this is called on two different pickings, one with UoM of things in 10's the other in things in 1, and they have prouct_uom_qty = 10 each.
This may not be as useful, because if would say {apple: 20} but in terms of the actual quantity of apples (if they are in units) it is 110.
I think this would only be done on ML in the same picking (as a guess)
So we could:
a. Use product_qty so then it is in the UoM of the product (would give 110 here)
b. raise error if ml with the same products have differnent UoMs (possibly safer)
c. Add to the doc string

Optionally, add a kwarg to return it in the UoM of the move, or the product.

Not sure how it is used, haven't checked but just something to consider.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+
+from odoo import models, _
+from odoo.exceptions import ValidationError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _should_reserve_full_packages(self):
+        """Method to determine if picking should reserve entire packages"""
```
peterpratt13, line 12: Will this ever be used on a group of pickings with the same type.
Could maybe do the ensure_one on the picking type?

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+                    picking.with_context(
+                        bypass_reserve_full_packages=True, quant_ids=all_quants.ids
+                    )._create_moves(
+                        remaining_qtys,
+                        values={"priority": picking.priority},
+                        confirm=True,
+                        assign=True,
+                    )
+
+    def _create_moves(
```
peterpratt13, line 75: We should use the create_moves behaviour in udes_stock here i believe

peterpratt13: prepare moves, then pass to  create move

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # do not reserve full packages when bypass_reserve_full packages
+        # is set in the context as True
+        if not self.env.context.get("bypass_reserve_full_packages"):
+            for picking in self.filtered(lambda p: p._should_reserve_full_packages()):
+                all_quants = Quant.browse()
+                remaining_qtys = defaultdict(int)
+
+                # get all packages
+                packages = picking.mapped("move_line_ids.package_id")
```
peterpratt13, line 32: Do we want to pre-cache this at the top for all relevant pickings

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+                for package in packages:
+                    move_lines = picking.mapped("move_line_ids").filtered(
+                        lambda ml: ml.package_id == package
+                    )
+                    pack_products = frozenset(package._get_all_products_quantities().items())
```
peterpratt13, line 37: I think this may cause a mismatch, this line gets the quantity info from the product uom, the below is in terms of the UoM of the move line (potentially different)

addons/udes_advanced_packaging/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    def _action_assign(self):
+        """Extend _action_assign to unreserve partial lines and trigger full package reservation"""
+        res = super(StockMove, self)._action_assign()
+
+        self.exists().unreserve_partial_lines()
+
+        self.mapped("picking_id")._reserve_full_packages()
+        return res
+
+    def unreserve_partial_lines(self):
```
PeterC10, line 16: We can get rid of this.

addons/udes_advanced_packaging/views/stock_picking_type.xml
```diff
@@ -0,0 +1,19 @@
...
+
+    <record id="view_picking_type_form_inherit" model="ir.ui.view">
+        <field name="name">stock.picking.type.form.inherit</field>
+        <field name="model">stock.picking.type</field>
+        <field name="inherit_id" ref="stock.view_picking_type_form"/>
+        <field name="arch" type="xml">
+
+            <!-- Add udes fields -->
+            <xpath expr="//form/sheet/group[last()]" position="after">
+                <group class="udes_wide_list" string="UDES Config" groups='base.group_no_one'>
```
PeterC10, line 12: CSS class `udes_wide_list` doesn't exist, we don't need it here though so can get rid of this class attribute.

Also we should set the string to "Packagisation" or "Advanced Packaging". In v11 we've got one group with a million config options, but now we can split out the config options into their own groups for each module. :)

addons/udes_advanced_packaging/views/stock_picking_type.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_picking_type_form_inherit" model="ir.ui.view">
+        <field name="name">stock.picking.type.form.inherit</field>
```
PeterC10, line 5: Use `udes_advanced_packaging` at the end of the name instead of `inherit`. That way if we ever need to debug views it is a bit more clear which module the override came from.

Have asked Jack to do similar in another story for the `udes_stock_cron` module.

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+        self.assertTrue(all([ml.package_id == package for ml in apple_mls]))
+
+    def test_partial_reservation_package_one_still_functional(self):
+        """Test to reserve a full package for one product
+        when initial demand < package quantity
+        """
+
+        # enable full package reservation
+        self.picking_type_pick.u_reserve_as_packages = False
```
PeterC10, line 104: The comment says enable but it looks like we're disabling?

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class TestPackageReservation(BaseUDES):
+    def test_reserve_full_package_one_product(self):
+        """Test to reserve a full package for one product
+        when initial demand < package quantity
+        """
+
+        # enable full package reservation
+        self.picking_type_pick.u_reserve_as_packages = True
```
PeterC10, line 13: We're repeating this same line quite a bit, we should create a `setUpClass` method so we only need to do this once. :)

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # create a package
+        package = self.create_package()
+
+        # create a quant of 3 apples in a stock sublocation for the package
+        apple_quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 3, package_id=package.id
+        )
+        sw_quant_1 = self.create_quant(
```
PeterC10, line 54: Suggestion: Can we just call this `strawberry_quant_1`?

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: What is the advantage of using ensure_one?

PeterC10: Currently this is only used in a lambda expression, so the pickings are passed one at a time effectively.

But yeah, no reason to not allow it to be called on a group, so long as they are all the same type.

@caleb-shelton imagine you pass in the following pickings to this method via `self`:

1. A pick picking
2. A goods out picking

Pick picking has full package reservation enabled, but the Goods Out doesn't. What do you return? True for Pick or false for Goods Out? In this scenario we can't give a proper answer, so `ensure_one()` would throw an error as we are using this method wrong.

The following should do the trick I think:

`self.picking_type_id.ensure_one()` - you don't always need to use mapped in Odoo 14, but if it errors you can do `self.mapped("picking_type_id").ensure_one()'.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+                    mls_products = frozenset(move_lines._get_all_products_quantities().items())
+                    if pack_products != mls_products:
+                        # move_lines do not match the quants
+                        pack_mls = MoveLine.search(
+                            [
+                                ("package_id", "child_of", package.id),
+                                ("state", "not in", ["done", "cancel"]),
+                            ]
+                        )
+                        other_pickings = pack_mls.mapped("picking_id") - picking
```
peterpratt13, line 47: Like boss pete said, the use of mapped everywhere is no longer needed in all places.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+                if remaining_qtys:
+                    # Context variables:
+                    # - filter the quants used in _create_moves() to be
+                    # the ones of the packages to be completed
+                    # - add bypass_reserve_full_packages at the context
+                    # to avoid to be called again inside _create_moves()
+                    picking.with_context(
+                        bypass_reserve_full_packages=True, quant_ids=all_quants.ids
+                    )._create_moves(
+                        remaining_qtys,
```
peterpratt13, line 69: Again, these are in the UoM of the product not the move. May want to preserve

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,156 @@
...
+            # with context variable bypass_reservation_update in order to avoid
+            # to execute code specific for Odoo UI at stock.move.line.write()
+            self.with_context(bypass_reservation_update=True).action_assign()
+            if result_package:
+                # update result_package_id of the new move_line_ids
+                package = Package.get_package(result_package)
+                new_move_line_ids = self.move_line_ids - old_move_line_ids
+                new_move_line_ids.write({"result_package_id": package.id})
+
+    def assert_valid_state(self):
```
peterpratt13, line 150: NIt: valid state is ambiguos - is there a better name? assert_is_closed maybe?

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,43 @@ def write(self, vals):
...
+            res[quant.product_id] += quant.quantity
+        return res
+
+    def assert_reserved_full_package(self, move_lines):
+        """Check that a package is fully reserved at move_lines."""
+        MoveLine = self.env["stock.move.line"]
+
+        self.ensure_one()
+
+        pack_products = frozenset(self._get_all_products_quantities().items())
```
peterpratt13, line 128: Again UoM stuff as above

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,43 @@ def write(self, vals):
...
+        mls_products = frozenset(move_lines._get_all_products_quantities().items())
+        if pack_products != mls_products:
+            # move_lines do not match the quants
+            picking = move_lines.mapped("picking_id")
```
peterpratt13, line 132: mapped no longer needed (and in other places will stop commenting :))

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,43 @@ def write(self, vals):
...
+            pack_mls = MoveLine.search(
+                [("package_id", "child_of", self.id), ("state", "not in", ["done", "cancel"])]
+            )
+            other_pickings = pack_mls.mapped("picking_id") - picking
+            if other_pickings:
+                raise ValidationError(
+                    _("The package is reserved in other pickings: %s")
+                    % ",".join(other_pickings.mapped("name"))
+                )
+            raise ValidationError(_("Cannot mark as done a partially reserved package."))
```
peterpratt13, line 143: nit: Would it be useful to include the package name here in the error?

PeterC10: I'm not convinced we should be adding this method as part of this story anyway.

On v11 it's used by the monster `update_picking` method, which we won't have in v14. But here we are only using it in a unit test.

addons/udes_advanced_packaging/models/stock_move_line.py
caleb-shelton: I ported this behaviour over, are you suggesting this as an improvement to how current UDES works?

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: Do you mean outside of the for loop? And if so, I am not sure if this would work the same way.

addons/udes_advanced_packaging/tests/test_package_reservation.py
caleb-shelton: fixed

addons/udes_advanced_packaging/views/stock_picking_type.xml
caleb-shelton: Fixed in latest commit

PeterC10: Small nit on that, could we have "Advanced Packaging" instead of Packing for the group string?

addons/udes_advanced_packaging/models/stock_picking.py
peterpratt13: picking.mapped("move_line_ids.package_id")
just this outside then its cached for all pickings then when looping through the picking it gets the relevant ones

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        # the quant should be unreserved
+        self.assertEqual(quant.reserved_quantity, 0, "The quant is reserved")
+
+        # create a Pick for 1 apple
+        create_info = [{"product": self.apple, "uom_qty": 1}]
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=create_info, confirm=True, assign=False
+        )
+        picking.action_assign()
```
peterpratt13, line 33: Why do we call action_assign after in create_picking. 
set assign to true? maybe

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # the quant should be fully reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids").filtered(lambda ml: ml.product_id == self.apple)
```
peterpratt13, line 39: no need for filter only one ml

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+            self.strawberry.id,
+            self.test_stock_location_01.id,
+            1,
+            package_id=package.id,
+            lot_name="sn02",
+        )
+        all_quants = apple_quant + strawberry_quant_1 + strawberry_quant_2
+
+        # all quants should be unreserved
+        self.assertTrue(
```
peterpratt13, line 71: Channelling keving - We could use some self.subTest here so we know which quants fail more easily

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: Fixed on latest commit

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: Fixed on latest commit

addons/udes_advanced_packaging/views/stock_picking_type.xml
caleb-shelton: Fixed

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: Do you mean to copy this method into udes_stock?

peterpratt13: No it already exists i believe

caleb-shelton: I probably need to rebase then lol

addons/udes_advanced_packaging/models/stock_picking.py
peterpratt13: typo 
pickings = self.filtered(...)
pickings.mapped("move_line_ids.package_id") before
then 
for picking in pickings:
     ...
     picking.mapped("move_line_ids.package_id")
     ....

addons/udes_advanced_packaging/models/stock_picking.py
peterpratt13: No i think _create_moves does not exist with the same name, but create_move or creete_moves does, you should re-use that functionality

addons/udes_advanced_packaging/models/stock_picking_type.py
```diff
@@ -0,0 +1,11 @@
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    u_reserve_as_packages = fields.Boolean(
+        string="Reserve entire packages",
```
PeterC10, line 8: Nit: `Reserve Entire Packages`

addons/udes_advanced_packaging/models/stock_quant.py
```diff
@@ -0,0 +1,24 @@
+from odoo import models
+
+from collections import defaultdict
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    def group_quantity_by_product(self, only_available=False):
```
PeterC10, line 9: Let's move this to `udes_stock`, this will be useful in other stock modules in the long term.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,43 @@ def write(self, vals):
         """
         self.validate_parent_name(vals)
         return super().write(vals)
+
+    def has_same_content(self, other):
```
PeterC10, line 105: We can get rid of this, this is used for package swap in v11.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,80 @@
...
+                    # - filter the quants used in _create_moves() to be
+                    # the ones of the packages to be completed
+                    # - add bypass_reserve_full_packages at the context
+                    # to avoid to be called again inside _create_moves()
+                    move_values = picking._prepare_move()
+                    picking.with_context(
+                        bypass_reserve_full_packages=True, quant_ids=all_quants.ids
+                    )._create_moves(pickings, move_values)
+
+    def assert_is_closed(self):
```
PeterC10, line 74: We should be able to use this, once we're using the create moves method in UDES 14.

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: This was ported from UDES11, is this a suggestion of improvement from existing behaviour, as it seems to be working fine for UDES11

addons/udes_advanced_packaging/tests/test_package_reservation.py
caleb-shelton: fixed in latest commit

addons/udes_advanced_packaging/models/stock_picking_type.py
caleb-shelton: fixed in latest commit

addons/udes_advanced_packaging/models/stock_quant_package.py
caleb-shelton: fixed in latest commit

addons/udes_advanced_packaging/models/stock_picking.py
peterpratt13: It works in 11 as uom doesn't exist. Everything is in eaches (units of 1's).
With different units of measure between the moves and products you may be comparing the quantities of the moves with the quants. 
Example:
if apples come in packs of 2 - and you have 10. You have a quantity of 5 (5*2=10, we never see the 10 in the fields as they all respect the UoM)
If you have a move for a pack of 5 apples- in a UoM of, the product_uom_qty = 1, the product_qty=5. Full pack reservation would be 3 packs of apples - to cover the 5. 

So this isn't in 11 - but this code assumes everywhere the same thing UoM matches. 
You can create an issue to investigate if things need to be changed in this code, change it now, or if you are happy it should work add a unit test to show it :).
If ti is the former, can you add a comment to which bits this impacts and link to the issue. (Even better write a unit test - if it passes awesome no issue, if not add a skip and link to the issue as well :).

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # create a package
+        package = self.create_package()
+
+        # create a quant of 4 apples in a stock sublocation for the package
+        quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=package.id
+        )
+        # the quant should be unreserved
+        self.assertEqual(quant.reserved_quantity, 0, "The quant is reserved")
```
peterpratt13, line 26: Nit can we drop the doc string - the error message from the asser Equal is probably more useful

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+            self.picking_type_pick, products_info=create_info, confirm=True, assign=False
+        )
+        picking.action_assign()
+
+        # the quant should be fully reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids")
+        self.assertTrue(all([ml.package_id == package for ml in apple_mls]))
```
peterpratt13, line 40: Are we expecting more than one move line here?
can just use assertEqual on the package_id

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # create a quant of 4 apples in a stock sublocation for the package
+        quant = self.create_quant(
+            self.apple.id, self.test_stock_location_01.id, 4, package_id=package.id
+        )
+        # the quant should be unreserved
+        self.assertEqual(quant.reserved_quantity, 0, "The quant is reserved")
+
+        # create a Pick for 1 apple
+        create_info = [{"product": self.apple, "uom_qty": 1}]
```
peterpratt13, line 117: nit: create_info is used multiple places, maybe move this into the setUpClass and re-use it. 
nit - rename to product_info like elsewhere

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+            self.picking_type_pick, products_info=create_info, confirm=True, assign=False
+        )
+        picking.action_assign()
+
+        # the quant should be partially reserved
+        self.assertEqual(quant.reserved_quantity, 1)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids").filtered(lambda ml: ml.product_id == self.apple)
+        self.assertTrue(all([ml.package_id == package for ml in apple_mls]))
```
peterpratt13, line 128: nit: Expect one move line so don't need the list comprehension assertTrue behaviour, assertEqual would be clearer

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        pick_2 = self.create_picking(
+            self.picking_type_pick,
+            products_info=[{"product": self.banana, "uom_qty": 1}],
+            confirm=True,
+            assign=False,
+        )
+
+        pickings = pick_1 | pick_2
+        pickings.action_assign()
```
peterpratt13, line 168: again action_assign

addons/udes_stock/models/stock_move.py
```diff
@@ -1,10 +1,31 @@
...
+from odoo import models, fields, api, _
+from odoo.addons import decimal_precision as dp
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    u_initial_demand = fields.Float("Initial Demand", digits="Product Unit of Measure")
+
+    product_uom_qty = fields.Float(string="Quantity")
```
peterpratt13, line 11: Why do we override?

addons/udes_advanced_packaging/views/stock_template.xml
```diff
@@ -0,0 +1,9 @@
+<odoo>
+    <data>
+        <template id="message_head" inherit_id="stock.message_head">
```
PeterC10, line 3: Add a comment above here to explain why we've done this.

addons/udes_stock/models/stock_move.py
PeterC10: We're changing the string from "Demand" to "Quantity".

This is because we've added `u_initial_demand` which will just be set to the initial quantity. Then we always have a track of what this was when `product_uom_quantity` is updated, e.g. because of full package reservation.

PeterC10: @caleb-shelton might be worth adding a comment above to make this clear

addons/udes_stock/models/stock_move.py
```diff
@@ -1,10 +1,31 @@
...
+
+    @api.model
+    def create(self, vals):
```
PeterC10, line 14: Nit: Add a doc-string.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -17,6 +17,12 @@ class StockMoveLine(models.Model):
         readonly=True,
     )
 
+    def _log_message(self, record, move, template, vals):
```
PeterC10, line 20: Nit: Add a doc-string

addons/udes_stock/models/stock_move.py
```diff
@@ -1,10 +1,31 @@
 # -*- coding: utf-8 -*-
-from odoo import models, api, _
+from odoo import models, fields, api, _
+from odoo.addons import decimal_precision as dp
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    u_initial_demand = fields.Float("Initial Demand", digits="Product Unit of Measure")
```
peterpratt13, line 9: Is this the move UoM? Can we just make it clearer - looks to me like the move UoM because we use product_uom_qty

addons/udes_stock/views/stock_picking.xml
```diff
@@ -80,6 +80,14 @@
                 </group>
             </xpath>
 
+            <xpath expr="//field[@name='move_ids_without_package']/tree/field[@name='product_uom_qty']" position="after">
```
PeterC10, line 83: Nit: Add comments above each XPATH to explain what we're doing.

addons/udes_stock/models/stock_quant.py
```diff
@@ -50,6 +52,24 @@ def get_quantities_by_key(self, get_key=lambda q: q.product_id, only_available=F
             products[get_key(quant)] += value
         return products
 
+    def group_quantity_by_product(self, only_available=False):
+        """
+        Returns a dictionary with the total quantity per product,
+        mapped by product_id.
```
peterpratt13, line 58: nit: add that we remove products that do not have any quantity

addons/udes_advanced_packaging/models/stock_picking.py
PeterC10: As discussed elsehwere, let's add TODOs referencing issue 962.

addons/udes_advanced_packaging/models/stock_move_line.py
PeterC10: As discussed elsewhere, let's add TODOs referencing issue 962.

addons/udes_stock/models/stock_move.py
PeterC10: Yeah let's rename it to u_uom_initial_demand

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # both quants should be fully reserved
+        self.assertEqual(apple_quant.reserved_quantity, 4)
+        self.assertEqual(banana_quant.reserved_quantity, 4)
```
peterpratt13, line 172: Test for package reeservation - unreservation to check u_initial_demand?

addons/udes_stock/models/stock_quant.py
PeterC10: We should also say `mapped by product` instead of `mapped by product_id`, as we use the product record, not ID.

addons/udes_advanced_packaging/models/stock_move.py
```diff
@@ -0,0 +1,12 @@
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_assign(self):
+        """Extend _action_assign to unreserve partial lines and trigger full package reservation"""
```
ws-mps, line 8: What do we mean by `to unreserve partial lines` ?

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
```
ws-mps, line 1: No longer needed. Check in the other files too.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+    _inherit = "stock.picking"
+
+    def _should_reserve_full_packages(self):
+        """Method to determine if picking should reserve entire packages"""
+        self.picking_type_id.ensure_one()
+        return self.picking_type_id.u_reserve_as_packages
+
+    def _reserve_full_packages(self):
+        """
+        If the picking type of the picking in self has full package
```
ws-mps, line 18: Typo: pickings

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+        reservation enabled, whole packages will be reserved.
+        """
+        MoveLine = self.env["stock.move.line"]
+
+        # do not reserve full packages when bypass_reserve_full packages
+        # is set in the context as True
+        if not self.env.context.get("bypass_reserve_full_packages"):
+            pickings = self.filtered(lambda p: p._should_reserve_full_packages())
+            pickings.move_line_ids.package_id
```
ws-mps, line 27: Seems that we are missing `.mapped()` here : `pickings.mapped("move_line_ids.package_id")`

addons/udes_advanced_packaging/models/stock_picking.py
ws-mps: Resolving here, both _get_all_products_quantities() should return the same measure, so we add TODOs there.

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,36 @@ def write(self, vals):
         """
         self.validate_parent_name(vals)
         return super().write(vals)
+
+    def _get_all_products_quantities(self):
+        """This function computes the different product quantities for the given package"""
```
ws-mps, line 106: Pete C comment applies here:
As discussed elsewhere, let's add TODOs referencing issue 962.

addons/udes_advanced_packaging/models/stock_picking.py
ws-mps: Forgot they changed this.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+                    pack_products = frozenset(package._get_all_products_quantities().items())
+                    mls_products = frozenset(move_lines._get_all_products_quantities().items())
+                    if pack_products != mls_products:
+                        # move_lines do not match the quants
+                        pack_mls = MoveLine.search(
+                            [
+                                ("package_id", "child_of", package.id),
+                                ("state", "not in", ["done", "cancel"]),
+                            ]
+                        )
```
ws-mps, line 47: If we don't care about the order, would be better to add order="id".

addons/udes_advanced_packaging/models/stock_picking.py
ws-mps: Is this still needed?

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,36 @@ def write(self, vals):
...
+
+        self.ensure_one()
+
+        pack_products = frozenset(self._get_all_products_quantities().items())
+        mls_products = frozenset(move_lines._get_all_products_quantities().items())
+        if pack_products != mls_products:
+            # move_lines do not match the quants
+            picking = move_lines.mapped("picking_id")
+            picking.ensure_one()
+            pack_mls = MoveLine.search(
```
ws-mps, line 127: It is the second time we have the same search, we should put this into a function: `_get_current_move_lines()`, probably we want it in `udes_stock`, since seems a useful helper function.

addons/udes_advanced_packaging/models/stock_quant_package.py
ws-mps: I think we can keep the function here, we might use it later on in other code. We won't have the monster function but we will have other functions that might need it.

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class TestPackageReservation(BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestPackageReservation, cls).setUpClass()
+        # enable full package reservation
+        cls.picking_type_pick.u_reserve_as_packages = True
+
+    def test_reserve_full_package_one_product(self):
```
ws-mps, line 13: Can we add "do_unreserve" at the end of the test and check that the state of the picking is the same as before calling `action_assign` please?

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        picking.action_assign()
+
+        # the quant should be fully reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids")
+        self.assertTrue(all([ml.package_id == package for ml in apple_mls]))
+
+    def test_reserve_full_package_two_products(self):
```
ws-mps, line 42: Can we add "do_unreserve" at the end of the test and check that the state of the picking is the same as before calling action_assign please? In this case, we are creating a move for strawberries that after unreserving won't go, which means we need a tidy up function to clear extra created moves that have now quantity 0.

addons/udes_advanced_packaging/tests/test_package_reservation.py
ws-mps: We can have more than one move line if we have more than one move because they are not merged on action_confirm, once that is resolved we should have one line only.

addons/udes_stock/models/stock_move.py
```diff
@@ -1,10 +1,31 @@
...
+        return super().create(vals)
+
+    def _do_unreserve(self):
+        """
+        Override _do_unreserve to remove extra quantities added to initial_demand when
+        unreserving a package with full package reservation.
+        """
+        res = super()._do_unreserve()
+        for move in self:
+            move.write({"product_uom_qty": move.u_initial_demand})
```
ws-mps, line 26: This is where we should delete moves where its initial demand is 0.

If this is about full package reservation, why it is not in the other module?

addons/udes_stock/models/stock_quant.py
ws-mps: Isn't this the same as: get_quantities_by_key ?

addons/udes_advanced_packaging/models/stock_picking.py
PeterC10: This should be sorted now, as the existing helper method in `udes_stock` is now being used.

addons/udes_advanced_packaging/models/stock_move.py
```diff
@@ -0,0 +1,12 @@
...
+        res = super(StockMove, self)._action_assign()
+
+        self.mapped("picking_id")._reserve_full_packages()
```
PeterC10, line 11: Don't need `.mapped` here.

addons/udes_advanced_packaging/models/stock_move.py
PeterC10: Looks like a copy/paste error, @caleb-shelton change this to just mention full package reservation in the doc-string. :)

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,36 @@ def write(self, vals):
...
+                [("package_id", "child_of", self.id), ("state", "not in", ["done", "cancel"])]
+            )
+            other_pickings = pack_mls.mapped("picking_id") - picking
```
PeterC10, line 130: Don't need `.mapped` here.

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=create_info, confirm=True, assign=False
+        )
+        picking.action_assign()
+
+        # the quant should be fully reserved
+        self.assertEqual(quant.reserved_quantity, 4)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids")
```
PeterC10, line 39: Don't need `.mapped` here.

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        picking.action_assign()
+
+        # apple quant should be fully reserved
+        self.assertEqual(apple_quant.reserved_quantity, 3)
+        # strawberry quants should be fully reserved
+        self.assertEqual(strawberry_quant_1.reserved_quantity, 1)
+        self.assertEqual(strawberry_quant_2.reserved_quantity, 1)
+
+        # apple stock.move.lines of the picking should have the package
+        apple_mls = picking.mapped("move_line_ids").filtered(lambda ml: ml.product_id == self.apple)
```
PeterC10, line 89: Don't need `.mapped` here.

addons/udes_advanced_packaging/tests/test_package_reservation.py
```diff
@@ -0,0 +1,172 @@
...
+        self.assertTrue(all([ml.package_id == package for ml in apple_mls]))
+
+        # strawberry stock.move.lines of the picking should have the package
+        strawberry_mls = picking.mapped("move_line_ids").filtered(
```
PeterC10, line 93: Don't need `.mapped` here.

addons/udes_advanced_packaging/tests/test_package_reservation.py
PeterC10: Merging is fixed by this line in `_prepare_move` in `udes_stock`: https://github.com/unipartdigital/udes-open/pull/554/files#diff-bd6ad1e2e7579def4095cbf373ff632d73cf4bcb1c1c3cc55f8299eaaf72accfR368

addons/udes_stock/models/stock_move.py
PeterC10: The idea is that in `udes_stock` we just add a field that will hold whatever the quantity was first set to. Atm only full package reservation will change the quantity from the initial value, but we might have other functionality that does so in future.

We don't need to delete any moves where the initial demand is 0, Odoo's merging of moves on confirm isn't affected by this. This is just for info only.

PeterC10: So imagine you have a picking for 5 apples, but only pack available has 20. After reserving stock the quantity is set from 5 to 20. But when unreserving we now set the quantity back to the initial demand.

PeterC10: I just spotted the doc-string though, @caleb-shelton please edit this so it doesn't mention full package reservation. :)

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+                            )
+
+                        quants = package._get_contained_quants()
+                        for product, qty in quants.group_quantity_by_product(
+                            only_available=True
+                        ).items():
+                            remaining_qtys[product] += qty
+                if remaining_qtys:
+                    # Context variables:
+                    # - filter the quants used in _create_moves() to be
```
PeterC10, line 62: Remove the comment talking about filtering quants used in `_create_moves`, it's not relevant.

addons/udes_advanced_packaging/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+                    # the ones of the packages to be completed
+                    # - add bypass_reserve_full_packages at the context
```
PeterC10, line 64: Move this comment to be above `picking.with_context(bypass_reserve_full_packages=True).action_assign()`. Also change `_create_moves` to `_create_move`.

addons/udes_advanced_packaging/models/stock_picking.py
caleb-shelton: Fixed in latest commit

addons/udes_advanced_packaging/tests/test_package_reservation.py
caleb-shelton: "Test to see if the quant is reserved?"

peterpratt13: I would just remove it, self.assertEqual(quant.reserved_quantity, 0)
it will spit out 5!=0 which is clear, 
otherwise the doc string needs to do something similar
"Expected 0 reserved quantity, got {quant.reserved_quantity}"

PeterAlabaster: @caleb-shelton think this one still needs doing

addons/udes_stock/models/stock_move.py
PeterAlabaster: Nit: Change `Override` to `Extend`

addons/udes_stock/models/stock_move_line.py
PeterAlabaster: Nit: Change `Override` to `Extend`

addons/udes_advanced_packaging/tests/test_package_reservation.py
PeterAlabaster: This assertion can be deleted now as you've added the for loop below

addons/udes_advanced_packaging/models/stock_quant_package.py
```diff
@@ -101,3 +101,34 @@ def write(self, vals):
...
+        if pack_products != mls_products:
+            # move_lines do not match the quants
+            picking = move_lines.picking_id
+            picking.ensure_one()
+            pack_mls = StockQuantPackage._get_current_move_lines()
+            other_pickings = pack_mls.picking_id - picking
+            if other_pickings:
+                raise ValidationError(
+                    _("The package is reserved in other pickings: %s")
+                    % ",".join(other_pickings.mapped("name"))
```
PeterAlabaster, line 132: is this .mapped() left because it's required or can it be removed?

PeterAlabaster:
```➜  udes_advanced_packaging git:(story/459_v2) ag '.mapped'       
models/stock_picking.py
44:                                % ",".join(other_pickings.mapped("name"))

models/stock_quant_package.py
132:                    % ",".join(other_pickings.mapped("name"))
```

caleb-shelton: I believe this is required, because it is not a relational field

addons/udes_stock/models/stock_quant.py
PeterC10: Yep exactly the same, have removed and replaced.

PeterC10 (approved):

PeterC10:
Before merging into master please reword the advanced packing commit to give a bit more insight into the changes.

PeterAlabaster (approved):

PR #555, opened 2022-01-19
[ADD] udes_open: Added goods in flag

nickcosta85:
Story/722

Signed-off-by: Nick Costa <nicholas.costa@unipart.io>

### DIFF
--- a/addons/udes_stock/data/picking_types.xml
+++ b/addons/udes_stock/data/picking_types.xml
@@ -92,6 +92,7 @@
       <field name="show_operations" eval="1"/>
       <field name="u_selection_method">source_doc</field>
       <field name="active">True</field>
+      <field name="goods_in">True</field>
     </record>
 
     <!-- Update "Goods Out" picking type -->
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -663,3 +663,16 @@ class StockPickingType(models.Model):
         action["domain"] = [("location_id", "child_of", location_ids)]
 
         return action
+
+    def _get_goods_in_picking_type_for_warehouse(self, company_id, warehouse_id):
+
+        Warehouse = self.env["stock.warehouse"]
+        warehouse = Warehouse.search(
+            [("company_id", "=", company_id), ("id", "=", warehouse_id)]
+        )
+        if not warehouse:
+            raise ValidationError(
+                ("Cannot find warehouse for warehouse id %s , company id %s")
+                % (warehouse_id, company_id)
+            )
+        return warehouse.u_goods_in_picking_type_id
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -6,9 +6,7 @@ CORE_LIFECYCLE_ACTIONS = [
     ("group_by_move_line_key", "Group by Move Line Key"),
     ("group_by_move_key", "Group by Move Key"),
 ]
-POST_ASSIGN_ACTIONS = CORE_LIFECYCLE_ACTIONS + [
-    ("by_maximum_quantity", "Maximum Quantity")
-]
+POST_ASSIGN_ACTIONS = CORE_LIFECYCLE_ACTIONS + [("by_maximum_quantity", "Maximum Quantity")]
 TARGET_STORAGE_FORMAT_OPTIONS = [
     ("pallet_products", "Pallet of products"),
     ("pallet_packages", "Pallet of packages"),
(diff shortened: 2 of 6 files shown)


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -24,6 +22,8 @@ class StockPickingType(models.Model):
 
     show_operations = fields.Boolean(default=True)
 
+    goods_in = fields.Boolean()
```
PeterAlabaster, line 25: would be good to have help string as to what this flag means/does

PeterAlabaster: and ideally we would split black format commit with the [ADD] commit :)

PeterC10: A few questions/points. :)

1. Why do we need this field? Goods In will be its own picking type.
2. We should put a string so that it is displayed like a title, the Odoo default would be set to `Goods in`, but `Goods In` looks more proper/professional.
3. As `stock.picking.type` is a core Odoo module, we should be prefixing new fields with `u_`.

nickcosta85: > and ideally we would split black format commit with the [ADD] commit :)

I will split the commits after I have fixed everything thanks.

nickcosta85: > A few questions/points. :)
> 
> 1. Why do we need this field? Goods In will be its own picking type.
> 2. We should put a string so that it is displayed like a title, the Odoo default would be set to `Goods in`, but `Goods In` looks more proper/professional.
> 3. As `stock.picking.type` is a core Odoo module, we should be prefixing new fields with `u_`.

1) Yes as per the discussion on my other MR I will most likey remove this, thanks.
2) If for some reason it stays in I will add one thanks.
3) OK I will try and do so from now on thanks.

adamrpatrick:
As discussed, we have found a different solution. Closing the MR.

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -84,6 +84,11 @@ def _domain_warehouse_picking_types(self):
         string="Stock Investigation Picking Type",
         help="Picking type used to create stock investigation pickings.",
     )
+    u_goods_in_picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        string="Goods In Picking Type",
+        help="Goods in picking type used for finding the correct picking type from a the warehouse.",
```
PeterC10, line 90: Nit: `from a the`.

TBH I think just `Goods In picking type for this warehouse` is sufficient.

PeterC10:
We can drop the first commit, the change there isn't relevant anymore.

The other 2 commits can be squashed I think, but please use the module name, not the repo name in the commit message.

addons/udes_stock/models/stock_warehouse.py
nickcosta85: This should be good now thanks.

addons/udes_stock/data/picking_types.xml
```diff
@@ -92,6 +92,7 @@
       <field name="show_operations" eval="1"/>
       <field name="u_selection_method">source_doc</field>
       <field name="active">True</field>
+      <field name="goods_in">True</field>
```
adamrpatrick, line 95: We don't need this bit anymore

addons/udes_stock/models/stock_picking_type.py
adamrpatrick, line 25: We don't need this field anymore

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -660,10 +659,20 @@ def get_action_package_tree(self):
...
-            "active_ids": location_ids,
-            "hide_create_date": False,
-        })
+        action["context"].update({"active_ids": location_ids, "hide_create_date": False})
         action["domain"] = [("location_id", "child_of", location_ids)]
 
         return action
+
+    def _get_goods_in_picking_type_for_warehouse(self, company_id, warehouse_id):
+
```
adamrpatrick, line 668: Add a docstring to explain what the purpose of this method is.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -660,10 +659,20 @@ def get_action_package_tree(self):
...
+        Warehouse = self.env["stock.warehouse"]
+        warehouse = Warehouse.search(
+            [("company_id", "=", company_id), ("id", "=", warehouse_id)]
+        )
+        if not warehouse:
+            raise ValidationError(
+                ("Cannot find warehouse for warehouse id %s , company id %s")
+                % (warehouse_id, company_id)
+            )
+        return warehouse.u_goods_in_picking_type_id
```
adamrpatrick, line 678: Just been playing around with this and there's actually already a field that does this for us `in_type_id`, so we don't need `u_goods_in_picking_type_id`

nickcosta85:
Closing this as I will just use the in_type_id as Adam pointed out.

PR #556, opened 2022-01-20
[ADD] udes_stock_cron: Migrated over the reserve_stock method in stoc…

jrbirch99:
Migrated over the reserve_stock method in stock_picking.py for the "reserve stock" cron job. This included Migrating over other methods which are also available in the udes_stock_cron module. Wrote some unit tests to check that partial reserves are allowed. Made a small change to the reserve_stock module - changed the way the string is formatted when a UserError exception is raised - this was to fix the error message that would be shown when no Internal Reference is input for a product.

Story/392

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -45,7 +45,7 @@ class StockMove(models.Model):
 
         """
         move_line_values = []
-
+        
         for move, qty in moves_info.items():
             vals = self._prepare_move_line(move, qty, **kwargs)
             move_line_values.append(vals)
--- a/addons/udes_stock_cron/__manifest__.py
+++ b/addons/udes_stock_cron/__manifest__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    "name": "UDES Stock Cron",
+    "summary": "udes stock cron",
+    "description": "Core models and configuration for UDES Stock cron - Odoo 14",
+    "author": "Unipart digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common", "udes_cron", "udes_stock"],
+    "data": [
+        "data/ir_cron.xml",
+        "views/stock_picking_type.xml",
+    ],
+}
--- a/addons/udes_stock_cron/data/ir_cron.xml
+++ b/addons/udes_stock_cron/data/ir_cron.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+
+        <record id="reserve_stock_action" model="ir.cron">
+        <field name="name">Reserve stock</field>
+        <field name="active" eval="True" />
+        <field name="user_id" ref="base.user_root" />
+        <field name="interval_number">5</field>
+        <field name="interval_type">minutes</field>
+        <field name="numbercall">-1</field>
+        <field name="doall">0</field>
+        <field name="model_id" ref="stock.model_stock_picking" />
+        <field name="state">code</field>
+        <field name="code">model.reserve_stock()</field>
+        </record>
+        
+    </data>
+</odoo>
--- a/addons/udes_stock_cron/models/stock_move.py
+++ b/addons/udes_stock_cron/models/stock_move.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+from odoo import models, api, _
+
+import logging
+_logger = logging.getLogger(__name__)
+
+# Map move state to the refactor stage.
+STAGES = {
+    "confirmed": "confirm",
+    "waiting": "confirm",
+    "assigned": "assign",
+    "partially_available": "assign",
+    "done": "validate",
+}
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_assign(self):
+        """Extend _action_assign to trigger refactor action and preprocess
+        location suggestions.
+        n.b. _action_assign does not return anything in core Odoo, so we
+        don't return any extra moves that may have been created
+        by refactoring.
+        """
+        res = super(StockMove, self)._action_assign()
+
+        self.exists().unreserve_partial_lines()
+        assign_moves = self.exists()._action_refactor(stage="assign")
+
+        for picking_type, moves in assign_moves.groupby("picking_type_id"):
+            # location suggestions
+            if picking_type.u_drop_location_preprocess:
+                moves.mapped("picking_id").apply_drop_location_policy()
+
+        assign_moves.mapped("picking_id")._reserve_full_packages()
+        return res
+
+    def unreserve_partial_lines(self):
+        """Unreserve any partially reserved lines if not allowed by the picking type."""
+        # Override to prevent assigning partial lines via the check
+        # availability button when u_handle_partial_lines is False,
+        pass
+
+    def _unreserve_partial_lines(self):
+        """Unreserve any partially reserved lines if not allowed by the picking type."""
+        Move = self.env["stock.move"]
+
+        moves_to_unreserve = Move.browse()
+        partial_moves = self.filtered(lambda m: m.state == "partially_available")
+        for picking_type, grouped_moves in partial_moves.groupby("picking_type_id"):
+            if picking_type.u_handle_partials and not picking_type.u_handle_partial_lines:
+                moves_to_unreserve += grouped_moves
+        moves_to_unreserve._do_unreserve()
+        return None
+
+
+    def _action_refactor(self, stage=None):
+        """Refactor moves in self.
+        :param stage: One of confirm|assign|done, if set, filters the moves
+            which will be refactored to only the state(s) that match:
+                - 'confirm': confirmed, waiting
+                - 'assign': assigned, partially_available
+                - 'done': done
+           Methods doing a refactor are expected to take a single recordset of
+           moves on which they will act, and to return the recordset of
+           equivalent moves after they have been transformed.
+           The output moves may be identical to the input, may contain none
+           of the input moves, or anywhere in between.
+           The output should contain a functionally similar set of moves.
+        """
+        if stage is not None and stage not in STAGES.values():
+            raise UserError(_("Unknown stage for move refactor: %s") % stage)
+        moves = self
+
+        if self._context.get("disable_move_refactor"):
+            return moves
+
+        rf_moves = moves.filtered(
+            lambda m: m.picking_type_id and m.state not in ["draft", "cancel"]
+        )
+        if stage is not None:
+            rf_moves = rf_moves.filtered(lambda m: STAGES[m.state] == stage)
+
+        for picking_type, pt_moves in rf_moves.groupby("picking_type_id"):
+            for stage, st_moves in pt_moves.groupby(lambda m: STAGES[m.state]):
+                if stage == "confirm":
+                    action = picking_type.u_post_confirm_action
+                elif stage == "assign":
+                    action = picking_type.u_post_assign_action
+                elif stage == "validate":
+                    action = picking_type.u_post_validate_action
+                else:
+                    continue  # Don't refactor cancel or draft moves.
+                
+                if action:
+                    _logger.info(
+                        "Refactoring %s at %s using %s: %s",
+                        picking_type.name,
+                        stage,
+                        action,
+                        st_moves.ids,
+                    )
+                    func = getattr(st_moves.with_context(refactor_stage=stage), "refactor_action_" + action)
+                    new_moves = func()
+                    if new_moves is not None:
+                        moves -= st_moves
+                        moves |= new_moves
+
+        return moves
+
--- a/addons/udes_stock_cron/models/stock_move_line.py
+++ b/addons/udes_stock_cron/models/stock_move_line.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from odoo import models, fields, _
+from collections import defaultdict
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def any_destination_locations_default(self):
+        """Checks if all location_dest_id's are the picks default
+           location_dest_id of the picking.
+        """
+        default_dest = self.mapped("picking_id.location_dest_id")
+        default_dest.ensure_one()
+        return any(ml.location_dest_id == default_dest for ml in self)
+    
+    def _get_all_products_quantities(self):
+        """This function computes the different product quantities for the given move_lines
+        """
+        res = defaultdict(int)
+        for move_line in self:
+            res[move_line.product_id] += move_line.product_uom_qty
+        return res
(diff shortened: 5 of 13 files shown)


addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "author": "Unipart digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common", "udes_cron", "udes_stock"],
+    "data": [
+        "data/ir_cron.xml",
+        "views/stock_picking_type.xml",
+    ],
+}
```
armandcela, line 15:
```suggestion
}

```
Nit: new line at the end of file

addons/udes_stock_cron/data/ir_cron.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="interval_type">minutes</field>
+        <field name="numbercall">-1</field>
+        <field name="doall">0</field>
+        <field name="model_id" ref="stock.model_stock_picking" />
+        <field name="state">code</field>
+        <field name="code">model.reserve_stock()</field>
+        </record>
+        
+    </data>
+</odoo>
```
armandcela, line 19:
```suggestion
</odoo>

```

addons/udes_stock/models/stock_move.py
```diff
@@ -45,7 +45,7 @@ def _prepare_move_lines(self, moves_info, **kwargs):
 
         """
         move_line_values = []
-
+
```
armandcela, line 48: Nit: indentation doesnt look consistent

addons/udes_stock_cron/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+        default_dest.ensure_one()
+        return any(ml.location_dest_id == default_dest for ml in self)
+    
+    def _get_all_products_quantities(self):
+        """This function computes the different product quantities for the given move_lines
+        """
+        res = defaultdict(int)
+        for move_line in self:
+            res[move_line.product_id] += move_line.product_uom_qty
+        return res
```
armandcela, line 22:
```suggestion
        return res
        
```

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+            # We want to reserve batches atomically, that is we will
+            # reserve pickings until all pickings in a batch have been
+            # assigned, even if we exceed the number of reservable pickings.
+            # However, the value of the handle partial flag is false we
+            # should not reserve stock if the batch cannot be completely
+            # reserved.
+            to_reserve = picking_type.u_num_reservable_pickings
+            reserve_all = to_reserve == -1
+            base_domain = [("picking_type_id", "=", picking_type.id), ("state", "=", "confirmed")]
+            limit = 1
```
armandcela, line 86: As far this can be a variable i would suggest to define in def reserve_stock with optional variable limit=1 and remove from here maybe.
`def reserve_stock(self, limit=1):`

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+import logging
+_logger = logging.getLogger(__name__)
+
+
+PG_CONCURRENCY_ERRORS_TO_RETRY = (
+    errorcodes.LOCK_NOT_AVAILABLE,
+    errorcodes.SERIALIZATION_FAILURE,
+    errorcodes.DEADLOCK_DETECTED,
+)
+MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
```
armandcela, line 16: Can this value be a configurable value maybe from parameters

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+                _("This policy(%s) is not meant to be used in " "preprocessing") % policy
+            )
+    
+    def get_suggested_locations(self, move_line_ids, limit=None, sort=True):
+        """Dispatch the configured suggestion location policy to
+        retrieve the suggested locations
+        """
+        result = self.env["stock.location"]
+
+        # WS-MPS: use self.ensure_one() and don't loop (self is used in all but
```
armandcela, line 233: Would suggest to remove specific names, and leave only the comment.
I understand that has been like that in the module that you are porting, seen that in other parts too.

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+                        if "sort" in expected_args:
+                            sort = False
+        if sort:
+            return result.sorted(lambda l: l.name)
+        else:
+            return result
+    
+    def _get_suggested_location_exactly_match_move_line(self, move_line_ids):
+        self._check_picking_move_lines_suggest_location(move_line_ids)
+        location = move_line_ids.mapped("location_dest_id")
```
armandcela, line 276: Not sure: think that in version 14 is not needed .mapped for id-s. Needs to be tried first though.
```suggestion
        location = move_line_ids.location_dest_id
```

addons/udes_stock_cron/models/stock_move.py
```diff
@@ -0,0 +1,111 @@
+# -*- coding: utf-8 -*-
+from odoo import models, api, _
+
+import logging
+_logger = logging.getLogger(__name__)
+
+# Map move state to the refactor stage.
```
PeterC10, line 7: We shouldn't be migrating any refactoring logic as part of this story. We'll have future stories to handle this, as we used a different approach in v13 which we'll want to use in v14.

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+
+        location.ensure_one()
+
+        if location.u_blocked or location.usage == "view":
+            return self.env["stock.location"]
+
+        return location
+
+    def _check_picking_move_lines_suggest_location(self, move_line_ids):
+        pick_move_lines = self.mapped("move_line_ids").filtered(lambda ml: ml in move_line_ids)
```
armandcela, line 286:
```suggestion
        pick_move_lines = self.move_line_ids.filtered(lambda ml: ml in move_line_ids)
```
Same here, not sure though has to be tested before when self is a multi record set

addons/udes_stock_cron/models/stock_picking_batch.py
```diff
@@ -0,0 +1,50 @@
...
+                if ready_picks and unready_picks:
+                    if batch.user_id:
+                        batch.state = "in_progress"
+                    else:
+                        batch.state = "waiting"
+
+                if done_picks and not ready_picks and not unready_picks:
+                    batch.state = "done"
+            else:
+                batch.state = "done"
```
armandcela, line 50: Nit: 
```suggestion
                batch.state = "done"
                
```

addons/udes_stock_cron/models/stock_picking_type.py
```diff
@@ -0,0 +1,97 @@
...
+        selection=POST_ASSIGN_ACTIONS,
+        string="Post Assign Action",
+        help="Choose the action to be taken after reserving a picking.",
+    )
+
+    u_post_validate_action = fields.Selection(
+        selection=CORE_LIFECYCLE_ACTIONS,
+        string="Post Validate Action",
+        help="Choose the action to be taken after validating a picking.",
+    )
```
armandcela, line 97: Nit:
```suggestion
    )
    
```

addons/udes_stock_cron/views/stock_picking_type.xml
```diff
@@ -0,0 +1,22 @@
...
+                    <field name="u_handle_partial_lines"/>
+                    <field name="u_drop_location_policy"/>
+                    <field name="u_drop_location_preprocess"/>
+                    <field name="u_reserve_as_packages"/>
+                    <field name="u_post_assign_action"/>
+                </group>
+            </xpath>
+        </field> 
+    </record>
+</odoo>
```
armandcela, line 22: Nit:
```suggestion
</odoo>

```

armandcela (changes requested):

PeterC10:
There's a fair bit of functionality that shouldn't be ported as part of this story, or shouldn't be living in the `udes_stock_cron` module.

Will give a proper explanation tomorrow, but this is a side effect of having such a big `udes_stock` module in v11, it does get really awkward to migrate at times. :)

addons/udes_stock_cron/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from odoo import models, fields, _
+from collections import defaultdict
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def any_destination_locations_default(self):
+        """Checks if all location_dest_id's are the picks default
```
peterpratt13, line 9: nit: Checks if any not all.
(Do we want this here or in udes_stock -seems pretty generic?)

addons/udes_stock_cron/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+           location_dest_id of the picking.
+        """
+        default_dest = self.mapped("picking_id.location_dest_id")
+        default_dest.ensure_one()
+        return any(ml.location_dest_id == default_dest for ml in self)
+    
+    def _get_all_products_quantities(self):
```
peterpratt13, line 16: This could be in udes_stock imo

addons/udes_stock_cron/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+        """This function computes the different product quantities for the given move_lines
+        """
+        res = defaultdict(int)
+        for move_line in self:
+            res[move_line.product_id] += move_line.product_uom_qty
```
peterpratt13, line 21: If this is called on two different pickings, one with UoM of things in 10's the other in things in 1, and they have prouct_uom_qty = 10 each.
This may not be as useful, because if would say {apple: 20} but in terms of the actual quantity of apples  (if they are in units) it is  110. 
Sowe could:
a. Use product_qty so then it is in the UoM of the product (would give 110 here)
b. raise error if ml with the same products have differnent UoMs
c. Add to the doc string

Not sure how it is used, haven't checked but just something to consider.

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+    errorcodes.LOCK_NOT_AVAILABLE,
+    errorcodes.SERIALIZATION_FAILURE,
+    errorcodes.DEADLOCK_DETECTED,
+)
+MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    batch_id = fields.Many2one("stock.picking.batch", copy=True)
```
peterpratt13, line 21: Perhaps this should be in udes_stock - i guess this was to get something to work in tests?

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+                                lambda x: x.state not in ["assigned", "cancel", "done"]
+                            )
+                            mls = pickings.mapped("move_line_ids")
+                            if unsatisfied:
+                                # Unreserve if the picking type cannot handle partials or it
+                                # can but there is nothing allocated (no stock.move.lines)
+                                if not picking_type.u_handle_partials or not mls:
+                                    # construct error message, report only products
+                                    # that are unreservable.
+                                    not_done = lambda x: x.state not in (
```
peterpratt13, line 140: This is the same lambdaas in unstaisfied

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+                                        "done",
+                                        "assigned",
+                                        "cancel",
+                                    )
+                                    moves = unsatisfied.mapped("move_lines").filtered(not_done)
+                                    products = moves.mapped("product_id.default_code")
+                                    picks = moves.mapped("picking_id.name")
+                                    msg = (
```
peterpratt13, line 148: Translatable?

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,387 @@
...
+        return refactored_moves.mapped("picking_id")
+
+    def reserve_stock(self):
+        """
+        Reserve stock according to the number of reservable pickings.
+        If this method is called on an empty recordset it will attempt to
+        reserve stock for all eligible picking types.  If the recordset is not
+        empty, it will reserve stock for the picks in the recordset. If the
+        recordset is not empty, it is the callers responsibility to make sure
+        that the pickings belong to at most one batch, otherwise this method
```
peterpratt13, line 47: What is the issue with if self is populated, raising an error when the pickings in self are in more than one batch?

addons/udes_stock_cron/models/stock_picking.py
peterpratt13: This can be done in a few places :)

addons/udes_stock_cron/data/ir_cron.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+
+        <record id="reserve_stock_action" model="ir.cron">
+        <field name="name">Reserve stock</field>
+        <field name="active" eval="True" />
```
peterpratt13, line 7: Does the DO want this disabled? It is not clear to me

peterpratt13:
you might be able to remove a bit of this code - because i think this story is more about the reserve stock code, so that function and those within it, along with the cron config should be sufficient.

addons/udes_stock_cron/data/ir_cron.xml
PeterC10: DO needs clarifying for sure. I would expect the cron job to be on by default, but disabled for all picking types by default. We should clarify this with product.

addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    "name": "UDES Stock Cron",
+    "summary": "udes stock cron",
```
PeterC10, line 4: Nit: Summary shouldn't be all lowercase.

addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "description": "Core models and configuration for UDES Stock cron - Odoo 14",
+    "author": "Unipart digital",
```
PeterC10, line 6: Nit: `Unipart Digital`, although we've got this same problem in other modules.

addons/udes_stock_cron/__manifest__.py
PeterC10, line 15: Nit: Missing newline at end of the file. Run black formatting to be safe.

PeterC10:
New PR: https://github.com/unipartdigital/udes-open/pull/557

PR #557, opened 2022-01-21
[ADD] udes_stock_cron

jrbirch99:
Created a new module called udes_stock_cron. Migrated over the reserve_stock method in stock_picking.py for the "reserve stock" cron job. This included Migrating over other methods which are also available in the udes_stock_cron module. Wrote some units tests to check my changes.  Made a small change to the reserve_stock module - changed the way the string is formatted when a UserError exception is raised - this was to fix the error message that would be shown when no Internal Reference is input for a product.

Story/392

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -96,10 +96,14 @@ class StockMove(models.Model):
         """
         self.ensure_one()
         if not all(ml.move_id == self for ml in move_lines):
-            raise ValueError(_("Cannot split move lines from a move they are not part of."))
+            raise ValueError(
+                _("Cannot split move lines from a move they are not part of.")
+            )
         if (
             move_lines == self.move_line_ids
-            and not self.move_orig_ids.filtered(lambda m: m.state not in ("done", "cancel"))
+            and not self.move_orig_ids.filtered(
+                lambda m: m.state not in ("done", "cancel")
+            )
             and not self.state == "partially_available"
         ):
             new_move = self
--- a/addons/udes_stock_cron/__manifest__.py
+++ b/addons/udes_stock_cron/__manifest__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    "name": "UDES Stock Cron",
+    "summary": "Udes stock cron",
+    "description": "Core models and configuration for UDES Stock cron - Odoo 14",
+    "author": "Unipart Digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["stock", "stock_picking_batch", "udes_cron", "udes_stock"],
+    "data": [
+        "data/ir_cron.xml",
+        "views/stock_picking_type.xml",
+    ],
+}
--- a/addons/udes_stock_cron/data/ir_cron.xml
+++ b/addons/udes_stock_cron/data/ir_cron.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+
+        <record id="reserve_stock_action" model="ir.cron">
+        <field name="name">Reserve stock</field>
+        <field name="active" eval="True" />
+        <field name="user_id" ref="base.user_root" />
+        <field name="interval_number">5</field>
+        <field name="interval_type">minutes</field>
+        <field name="numbercall">-1</field>
+        <field name="doall">0</field>
+        <field name="model_id" ref="stock.model_stock_picking" />
+        <field name="state">code</field>
+        <field name="code">model.reserve_stock()</field>
+        </record>
+        
+    </data>
+</odoo>
--- a/addons/udes_stock_cron/models/stock_picking.py
+++ b/addons/udes_stock_cron/models/stock_picking.py
@@ -1,0 +1,197 @@
+# -*- coding: utf-8 -*-
+import time
+from psycopg2 import OperationalError, errorcodes
+from collections import defaultdict
+
+from odoo import fields, models, _, api
+from odoo.exceptions import UserError, ValidationError
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+PG_CONCURRENCY_ERRORS_TO_RETRY = (
+    errorcodes.LOCK_NOT_AVAILABLE,
+    errorcodes.SERIALIZATION_FAILURE,
+    errorcodes.DEADLOCK_DETECTED,
+)
+MAX_TRIES_ON_CONCURRENCY_FAILURE = 5
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _reserve_stock_assign(self):
+        """Perform assign of the pickings at the move level because refactoring
+        may change the pickings.
+        This is a function so it can be extended and/or overriden.
+        """
+        # Assign at the move level disabling refactoring
+        moves = self.move_lines
+        moves._action_assign()
+        return moves.picking_id
+
+    def reserve_stock(self):
+        """
+        Reserve stock according to the number of reservable pickings.
+        If this method is called on an empty recordset it will attempt to
+        reserve stock for all eligible picking types.  If the recordset is not
+        empty, it will reserve stock for the picks in the recordset. If the
+        recordset is not empty, it is the callers responsibility to make sure
+        that the pickings belong to at most one batch, otherwise this method
+        cannot respect the priority order of pickings, in this case the
+        behaviour of this method is undefined.
+        In either scenario the picking type flags for reserving complete
+        batches and handling partial batches are respected.
+        The number of reservable pickings is defined on the picking type.
+        0 reservable pickings means this function should not reserve stock
+        -1 reservable picking means all reservable stock should be reserved.
+        """
+
+        Picking = self.env["stock.picking"]
+        PickingType = self.env["stock.picking.type"]
+
+        if self:
+            picking_types = self.picking_type_id
+        else:
+            picking_types = PickingType.search(
+                [("active", "=", True), ("u_num_reservable_pickings", "!=", 0)]
+            )
+
+        # We will either reserve up to the reservation limit or until all
+        # available picks have been reserved, depending on the value of
+        # u_num_reservable_pickings.
+        # However we must also take into account the atomic batch reservation
+        # flag (u_reserve_batches) and the handle partial flag
+        # (u_handle_partials).
+
+        for picking_type in picking_types:
+            _logger.info("Reserving stock for picking type %r.", picking_type)
+
+            # We want to reserve batches atomically, that is we will
+            # reserve pickings until all pickings in a batch have been
+            # assigned, even if we exceed the number of reservable pickings.
+            # However, the value of the handle partial flag is false we
+            # should not reserve stock if the batch cannot be completely
+            # reserved.
+            to_reserve = picking_type.u_num_reservable_pickings
+            reserve_all = to_reserve == -1
+            base_domain = [
+                ("picking_type_id", "=", picking_type.id),
+                ("state", "=", "confirmed"),
+            ]
+            limit = 1
+            processed = Picking.browse()
+            unsatisfied_state = lambda p: p.state not in ("assigned", "cancel", "done")
+
+            while reserve_all or to_reserve > 0:
+
+                if self:
+                    pickings = self.filtered(
+                        lambda p: p.picking_type_id == picking_type
+                    )
+                    # Removed processed pickings from self
+                    pickings -= processed
+                else:
+                    domain = base_domain[:]
+                    if processed:
+                        domain.append(("id", "not in", processed.ids))
+                    pickings = Picking.search(domain, limit=limit)
+
+                _logger.info(f"Reserving stock for pickings {pickings}.")
+                if not pickings:
+                    # No pickings left to process.
+                    # If u_num_reservable_pickings is -1, or there are
+                    # fewer available pickings that the limit, the loop must
+                    # terminate here.
+                    break
+
+                batch = pickings.batch_id
+                if batch and batch.state == "draft":
+                    # Add to seen pickings so that we don't try to process
+                    # this batch again.
+                    processed |= batch.picking_ids
+                    continue
+
+                if batch and picking_type.u_reserve_batches:
+                    pickings = batch.picking_ids
+
+                # MPS: mimic Odoo's retry behaviour
+                tries = 0
+                while True:
+
+                    try:
+                        with self.env.cr.savepoint():
+                            pickings = pickings._reserve_stock_assign()
+                            batch._compute_state()
+
+                            processed |= pickings
+
+                            unsatisfied = pickings.filtered(unsatisfied_state)
+
+                            mls = pickings.move_line_ids
+                            if unsatisfied:
+                                # Unreserve if the picking type cannot handle partials or it
+                                # can but there is nothing allocated (no stock.move.lines)
+                                if not mls:
+                                    # construct error message, report only products
+                                    # that are unreservable.
+                                    moves = unsatisfied.move_lines.filtered(
+                                        unsatisfied_state
+                                    )
+                                    products = moves.product_id.name_get()
+                                    picks = moves.picking_id.name
+                                    msg = (
+                                        f"Unable to reserve stock for products {products} "
+                                        f"for pickings {picks}."
+                                    )
+                                    raise UserError(msg)
+                            break
+                    except UserError as e:
+                        self.invalidate_cache()
+                        # Only propagate the error if the function has been
+                        # manually triggered
+                        if self:
+                            raise e
+                        tries = -1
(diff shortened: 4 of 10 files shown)


addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "author": "Unipart digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common", "udes_cron", "udes_stock"],
+    "data": [
+        "data/ir_cron.xml",
+        "views/stock_picking_type.xml",
+    ],
+}
```
armandcela, line 15: Nit: Newline
```suggestion
}

```

addons/udes_stock_cron/data/ir_cron.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="interval_type">minutes</field>
+        <field name="numbercall">-1</field>
+        <field name="doall">0</field>
+        <field name="model_id" ref="stock.model_stock_picking" />
+        <field name="state">code</field>
+        <field name="code">model.reserve_stock()</field>
+        </record>
+        
+    </data>
+</odoo>
```
armandcela, line 19: Nit:
```suggestion
</odoo>

```

addons/udes_stock_cron/views/stock_picking_type.xml
```diff
@@ -0,0 +1,19 @@
...
+            <!-- Add udes fields -->
+            <xpath expr="//form/sheet/group[last()]" position="after">
+                <group string="Scheduled Actions" groups='base.group_no_one'>
+                    <field name="u_num_reservable_pickings"/>
+                    <field name="u_reserve_batches"/>
+                </group>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
armandcela, line 19: Nit:
```suggestion
</odoo>

```

addons/udes_stock_cron/models/stock_picking_type.py
```diff
@@ -0,0 +1,24 @@
...
+
+    u_num_reservable_pickings = fields.Integer(
+        string="Number of Pickings to Reserve",
+        default=0,
+        help="The number of pickings in the queue to reserve stock for. "
+        "If batch reservation is enabled, entire picking batches are "
+        "reserved in the order of their earliest picking in the queue "
+        "until at least this number of pickings are reserved. "
+        "0 indicates no pickings should be reserved. "
+        "-1 indicates all pickings should be reserved.",
```
PeterC10, line 16: Might be worth considering if there is a nicer way to do this. Looking at the config screen it isn't clear that `-1` means reserve all confirming pickings.

Personally I think it would look nicer to have a boolean field for saying whether a picking type gets reserved via a cron job, then another field for saying how many, but adding an extra field isn't exactly nice...

Can ignore this comment for now. :)

addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    "name": "UDES Stock Cron",
+    "summary": "udes stock cron",
+    "description": "Core models and configuration for UDES Stock cron - Odoo 14",
+    "author": "Unipart digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common", "udes_cron", "udes_stock"],
```
armandcela, line 10: Nit:
I think that list of depends module shouldnt be longer than needed:
I think that we add on depends modules if needed or if they are specifically somewhere mostly on xml or any external identifier called.
Example stock is not needed to add as we are adding udes_stock which depends on stock and by hierarchy makes udes_stock_cron too depending on stock, but in our example we need it as you are inheriting a view from stock and using stock.xml_id in inherited_id.
But maybe we dont need base as in fact all modules depend on base, maybe we dont need udes_common if udes_cron depends on it and we are not using the name of the module in udes_stock_cron.
Hope that i was clear

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,201 @@
...
+    def reserve_stock(self):
+        """
+        Reserve stock according to the number of reservable pickings.
+        If this method is called on an empty recordset it will attempt to
+        reserve stock for all eligible picking types.  If the recordset is not
+        empty, it will reserve stock for the picks in the recordset. If the
+        recordset is not empty, it is the callers responsibility to make sure
+        that the pickings belong to at most one batch, otherwise this method
+        cannot respect the priority order of pickings, in this case the
+        behaviour of this method is undefined.
```
peterpratt13, line 44: Q. Can we just throw an error in the case  where multiple batches are present in the passed picking record set

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,201 @@
...
+                            mls = pickings.mapped("move_line_ids")
+                            if unsatisfied:
+                                # Unreserve if the picking type cannot handle partials or it
+                                # can but there is nothing allocated (no stock.move.lines)
+                                if not mls:
+                                    # construct error message, report only products
+                                    # that are unreservable.
+                                    moves = unsatisfied.mapped("move_lines").filtered(
+                                        lambda m: m.state
+                                        not in (
```
peterpratt13, line 141: repeated lambda expression to above - var?

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,201 @@
...
+                    pickings = Picking.search(domain, limit=limit)
+
+                if not pickings:
+                    # No pickings left to process.
+                    # If u_num_reservable_pickings is -1, or there are
+                    # fewer available pickings that the limit, the loop must
+                    # terminate here.
+                    break
+
+                batch = pickings.mapped("batch_id")
```
peterpratt13, line 108: think armand pointed out no longer need mapped everywhere :)

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,201 @@
...
+                    continue
+
+                if batch and picking_type.u_reserve_batches:
+                    pickings = batch.picking_ids
+
+                # MPS: mimic Odoo's retry behaviour
+                tries = 0
+                while True:
+
+                    try:
```
peterpratt13, line 122: This function is fairly long -can we split up anywhere in a sensible place?

jrbirch99: I think this definitely needs to happen, but maybe this could be part of another story? I think migrating over the function and ensuring it works was the purpose of this story. Then any design changes could be made in another story.

addons/udes_stock_cron/models/stock_picking_type.py
adamrpatrick: Yep we can tackle this later

addons/udes_stock/models/stock_move.py
```diff
@@ -45,7 +45,7 @@ def _prepare_move_lines(self, moves_info, **kwargs):
 
         """
         move_line_values = []
-
+
```
adamrpatrick, line 48: Seems to be some extra spaces added here

addons/udes_stock_cron/models/stock_picking.py
jrbirch99: I think this could be done, or I was thinking you could filter out all the pickings that belong to more than one batch? But as I said in one of the other comments I don't think I should mess with the design too much, and it could be implemented in another story.

addons/udes_stock/models/stock_move.py
PeterC10, line 48: Nit: Empty space added. Worth running black again to be safe.

addons/udes_stock_cron/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    "name": "UDES Stock Cron",
+    "summary": "udes stock cron",
+    "description": "Core models and configuration for UDES Stock cron - Odoo 14",
+    "author": "Unipart digital",
```
PeterC10, line 6: Nit: `Unipart Digital`

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+    # Want to inherit the methods from BaseUDES but do not want the database to
+    # rollback after each test, as the method in test: reserve_stock, commits to
+    # the database via the cursor - so will not be able to rollback
+    def setUp(cls):
+        pass
+
+    def tearDown(cls):
+        pass
```
adamrpatrick, line 38: Does this mean that the records created in the test cases persist in the DB after running the tests?

jrbirch99: The commit in the reserve stock method means that the database can't roll back to before the tests were run. I overwrote tearDown because it tries to roll back after every test case. There is also a tearDownClass which is meant to roll back the database after all the tests are done, however it can't rollback properly due to the commit in the reserve stock method. As it stands there will still be records in the database after running the tests. I thought that maybe I could extend tearDownClass for these tests and make it drop the database?

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+
+        Move = self.test_picking_pick.move_lines.filtered(
+            lambda m: m.name == "Test product Banana"
+        )
+        self.assertEqual(Move.name, "Test product Banana")
+        self.assertEqual(Move.reserved_availability, 0.0)
+        self.assertEqual(Move.state, "confirmed")
+
+        with self.assertRaises(UserError):
```
adamrpatrick, line 91: Let's check that we are getting the correct `UserError` exception being raised here seeing as we also raise another `UserError` in the method being tested.

E.G. 
```
with self.assertRaises(UserError) as e:
     blah
self.assertEqual(e.exception.name, "Unable to reserve stock....BLAH")
```

addons/udes_stock_cron/tests/test_stock_picking.py
adamrpatrick: Until we think of a better way, let's extend `tearDownClass` to `unlink` the records created as part of these tests

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,194 @@
...
+                                # Unreserve if the picking type cannot handle partials or it
+                                # can but there is nothing allocated (no stock.move.lines)
+                                if not mls:
+                                    # construct error message, report only products
+                                    # that are unreservable.
+                                    moves = unsatisfied.move_lines.filtered(unsatisfied_state)
+                                    products = moves.product_id.default_code
+                                    picks = moves.picking_id.name
+                                    msg = (
+                                        f"Unable to reserve stock for products {products} "
```
PeterC10, line 143: Q: What does this error message look like if a product doesn't have a barcode?

Might be better to use something like `name_get()`.

adamrpatrick: Ah that's a good point, I think the current logic fails if a product doesn't have a barcode. We should make sure that doesn't trip it up.

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+            self.picking_type_pick,
+            products_info=products_info,
+            confirm=True,
+            location_dest_id=self.test_received_location_01.id,
+        )
+
+        Move = self.test_picking_pick.move_lines.filtered(
+            lambda m: m.name == "Test product Banana"
+        )
+        self.assertEqual(Move.name, "Test product Banana")
```
peterpratt13, line 87: nit: can probably just assert one move exists rather than having to filter as tehre should only be one move

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+        # Check that a stock picking with no stock available cannot reserve stock
+        products_info = [{"product": self.banana, "uom_qty": 10.0}]
+        self.test_picking_pick = self.create_picking(
+            self.picking_type_pick,
+            products_info=products_info,
+            confirm=True,
+            location_dest_id=self.test_received_location_01.id,
+        )
+
+        Move = self.test_picking_pick.move_lines.filtered(
```
peterpratt13, line 84: no need to filter - see below

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+    # rollback after each test, as the method in test: reserve_stock, commits to
+    # the database via the cursor - so will not be able to rollback
+    def setUp(cls):
+        pass
+
+    def tearDown(cls):
+        pass
+
+    @classmethod
+    def create_quant(self, product_id, location_id, quantity, **kwargs):
```
peterpratt13, line 41: doesn't this eists in udes_stock already?

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+        Quant = self.env["stock.quant"]
+        vals = {
+            "product_id": product_id,
+            "location_id": location_id,
+            "quantity": quantity,
+        }
+        return Quant.create(vals, **kwargs)
+
+    def test_reserve_available_stock(self):
+        # Check there is stock available
```
peterpratt13, line 56: nit: Move these comments into a doc string?

addons/udes_stock_cron/tests/test_stock_picking.py
```diff
@@ -0,0 +1,96 @@
...
+        # Check there is no reserved quantity and that the stock_picking is in state 'confirmed'
+        # Run reserve_stock, which should then reserve quantity and change state of stock_picking
+        self.assertEqual(self.test_quant_apple.quantity, 10.0)
+
+        Move = self.test_picking_pick.move_lines.filtered(
+            lambda m: m.name == "Test product Apple"
+        )
+        self.assertEqual(Move.name, "Test product Apple")
```
peterpratt13, line 64: Same here - no need to filter if one move line, just assert the number. Could assert the prdoduct if you like

addons/udes_stock_cron/tests/test_stock_picking.py
adamrpatrick: As discussed, we've found a better way through the use of `enter_test_mode` and `leave_test_mode`

addons/udes_stock_cron/models/stock_picking.py
jrbirch99: If there is not default_code (which is the products internal_reference) for the product then the msg reads "Unable to reserve stock for products False for pickings *pickings*". It doesn't error for the user. There is also no user error when the barcode for the product is left out.

PeterC10: It would show in the logs at least, so it would be better to have an identifier for a product even without a barcode. :)

Try it with `name_get()`, that should include the product name and reference.

addons/udes_stock_cron/tests/test_stock_picking.py
peterpratt13: nit: can we change Move to move ( normally reserve capitalised names for the model)

peterpratt13:
LGTM - just one nit about the Move -> move in the unit tests from me :)

armandcela (approved):

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,198 @@
...
+                                    # construct error message, report only products
+                                    # that are unreservable.
+                                    moves = unsatisfied.move_lines.filtered(
+                                        unsatisfied_state
+                                    )
+                                    moves_name = moves.product_id.name_get()
+                                    products = moves.product_id.default_code
+                                    picks = moves.picking_id.name
+                                    msg = (
+                                        f"Unable to reserve stock for products {products} with record {moves_name} "
```
PeterC10, line 147: We don't need `moves_name` here.

Just do: `products = moves.product_id.name_get()` and have the error message as it was before. :)

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -0,0 +1,201 @@
...
+                    )
+                    # Removed processed pickings from self
+                    pickings -= processed
+                else:
+                    domain = base_domain[:]
+                    if processed:
+                        domain.append(("id", "not in", processed.ids))
+                    pickings = Picking.search(domain, limit=limit)
+
+                if not pickings:
```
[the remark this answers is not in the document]
jrbirch99: This should be done now

adamrpatrick (approved):

PeterC10 (approved):

peterpratt13 (approved):

PR #559, opened 2022-02-01
Story/913

jrbirch99:
Migrated over the push_from_drop function from UDES 11. Migrated over the associated unit tests.

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from odoo import models, fields, api, _
 from odoo.addons import decimal_precision as dp
 
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -1,7 +1,5 @@
-# -*- coding: utf-8 -*-
 from odoo import api, models, fields, _
 from odoo.exceptions import ValidationError
-
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
--- a/addons/udes_stock_routing/models/stock_rule.py
+++ b/addons/udes_stock_routing/models/stock_rule.py
@@ -1,0 +1,49 @@
+from odoo import api, fields, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    u_push_on_drop = fields.Boolean(
+        "Push on Drop-off",
+        default=False,
+        help="When stock is dropped in the "
+        "from_location of this path, "
+        "automatically push it onwards.",
+    )
+
+    @api.model
+    def get_path_from_location(self, location):
+        """
+        Find a single stock.rule for which the given location is
+        a valid starting location. If one not found return an empty recordset. 
+        Search for the stock.rules attached to the parent locations of location. If there is more than one stock.rule in the location hierarchy,
+        sort them by choosing the parent_path with the largest integer at the end of its path. This is the closest parent location to the input
+        location and hence the most relevant stock.rule. 
+        Note: If there is more than one stock.rule defined on the closest location, the newest stock.rule on that location will be chosen. 
+        """
+        Rule = self.env["stock.rule"]
+        push_steps = Rule.search(
+            [
+                (
+                    "location_src_id",
+                    "parent_of",
+                    [
+                        location.id,
+                    ],
+                ),
+                ("u_push_on_drop", "=", True),
+            ]
+        )
+        if push_steps:
+            return push_steps.sorted(key=lambda p: p.location_src_id.parent_path.strip("/").split("/")[-1], reverse=True)[0]
+        return self.browse()
+
+    def _run_push(self, move):
+        """
+        Odoo has default behaviour to create another picking after a move is confirmed. Don't want this functionality
+        if u_push_on_drop is turned on.    
+        """
+        if self.u_push_on_drop:
+            return 
+        return super()._run_push(move)
--- a/addons/udes_stock_routing/__init__.py
+++ b/addons/udes_stock_routing/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import tests
--- a/addons/udes_stock_routing/__manifest__.py
+++ b/addons/udes_stock_routing/__manifest__.py
@@ -1,0 +1,13 @@
+{
+    "name": "UDES Stock Routing",
+    "summary": "UDES stock routing",
+    "description": "Core models and configuration for UDES Stock Routing - Odoo 14",
+    "author": "Unipart Digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "UDES",
+    "version": "0.1",
+    "depends": ["udes_stock"],
+    "data": [
+        "views/stock_rule_views.xml",
+    ],
+}
--- a/addons/udes_stock_routing/models/__init__.py
+++ b/addons/udes_stock_routing/models/__init__.py
@@ -1,0 +1,2 @@
+from . import stock_move
+from . import stock_rule
--- a/addons/udes_stock_routing/models/stock_move.py
+++ b/addons/udes_stock_routing/models/stock_move.py
@@ -1,0 +1,95 @@
+from odoo import models, api
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def push_from_drop(self):
+        """
+        Creates new moves for the moves that have just dropped stock in a location.  
+        """
+        Move = self.env["stock.move"]
+        MoveLine = self.env["stock.move.line"]
+        Push = self.env["stock.rule"]
+
+        done_moves = self.filtered(lambda m: m.state == "done")
+
+        # load all the move lines, grouped by location
+        move_lines_by_location = done_moves.move_line_ids.groupby("location_dest_id")
+
+        # Build mapping of push rule -> move lines to push
+        move_lines_by_push = {}
+        for location, loc_mls in move_lines_by_location:
+            # Get the push rule that moves from the location.
+            push_step = Push.get_path_from_location(location)
+            if not push_step:
+                continue
+            if push_step not in move_lines_by_push:
+                move_lines_by_push[push_step] = MoveLine.browse()
+            move_lines_by_push[push_step] |= loc_mls
+
+        created_moves = Move.browse()
+        for push, move_lines in move_lines_by_push.items():
+            created_moves |= self._create_moves_for_push(push, move_lines)
+
+        confirmed_moves = created_moves._action_confirm()
+        confirmed_moves._action_assign()
+
+    def _get_push_move_vals(self, move_lines):
+        """
+        Create and return a dict of values from single move in self,
+        used to create new move for push rule.
+
+        Set quantity from total of move's move lines and set link to original move.
+        """
+        self.ensure_one()
+
+        move_vals = {
+            "product_uom_qty": sum(move_lines.mapped("qty_done")),
+            "move_orig_ids": [(6, 0, self.ids)],
+        }
+        return move_vals   
+
+    @api.model
+    def _create_moves_for_push(self, push, move_lines):
+        """Create moves for a push rule to cover the quantity in move_lines"""
+        Move = self.env["stock.move"]
+
+        # Group mls by move so we can preserve move information.
+        mls_by_move = move_lines.groupby("move_id")
+        created_moves = Move.browse()
+        base_vals = {
+            "picking_type_id": push.picking_type_id.id,
+            "location_id": push.location_src_id.id,
+            "location_dest_id": push.location_id.id,
+            "picking_id": None,
+        }
+        for move, mls in mls_by_move:
+            move_vals = base_vals.copy()
+            move_vals.update(move._get_push_move_vals(mls))
+            created_moves |= move.copy(move_vals)
+        return created_moves
+    
+    def _action_done(self, cancel_backorder=False):
+        """
+        Extend _action_done to push from drop. 
+        """
+        done_moves = super()._action_done(cancel_backorder=cancel_backorder)
+        done_moves.push_from_drop()
+        return done_moves
+
+    def _get_new_picking_values(self):
+        """
+        Extend _get_new_picking_values() to propagate the origin and partner_id fields for next "push" picking.
+        """
+        values = super()._get_new_picking_values()
+        previous_pickings = self.move_orig_ids.picking_id
+        if not values.get("origin"):
+            previous_origin = list(set(previous_pickings.mapped("origin")))
+            if len(previous_origin) == 1:
+                values["origin"] = previous_origin[0]
+        if not values.get("partner_id"):
+            previous_partner = previous_pickings.partner_id
+            if len(previous_partner) == 1:
+                values["partner_id"] = previous_partner.id
+
+        return values
--- a/addons/udes_stock_routing/tests/__init__.py
+++ b/addons/udes_stock_routing/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_push_from_drop
--- a/addons/udes_stock_routing/tests/test_push_from_drop.py
+++ b/addons/udes_stock_routing/tests/test_push_from_drop.py
@@ -1,0 +1,238 @@
+from odoo.addons.udes_stock.tests.common import BaseUDES
+
+
+class PushFromDropBase(BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        """Setup branching Inbound flow
+        Goods In: Customers to Input
+        Putaway: Input to Stock
+
+        Putaway Push: Input/Received to Stock
+        Damaged Push: Input/Damaged to QC
+
+        Putaway Push and Damaged Push both push from drop.
+        """
+        super(PushFromDropBase, cls).setUpClass()
+
+        Location = cls.env['stock.location']
+        Push = cls.env['stock.rule']
+
+        input_zone = cls.env.ref('stock.stock_location_company')
+        input_zone.usage = 'view'
+        cls.picking_type_goods_in.default_location_dest_id = input_zone.id
+        cls.picking_type_putaway.default_location_src_id = input_zone.id
+
+        cls.received_location.location_id = input_zone
+        cls.received_damaged_location = Location.create({
+            'name': "TEST_RECEIVEDDAMAGED",
+            'barcode': "LTESTRECEIVEDDAMAGED",
+            'location_id': input_zone.id,
+        })
+
+        cls.location_qc_zone = Location.create({
+            'name': "QC Stock",
+            'usage': 'view',
+        })
+        cls.location_qc_01 = Location.create({
+            'name': "QC Stock 01",
+            'barcode': "LTESTQC01",
+            'location_id': cls.location_qc_zone.id,
+        })
+
+        # Update path_in_putaway to use correct location:
+        cls.push_putaway.write({
+            "location_src_id": cls.received_location.id,
+            "u_push_on_drop": True,
+        })
+
+        cls.push_damaged_putaway = Push.create({
+            "name": "TestDamagedPutaway",
+            "route_id": cls.route_in.id,
+            "sequence": 30,
+            "location_src_id": cls.received_damaged_location.id,
+            "location_id": cls.location_qc_zone.id,
+            "picking_type_id": cls.picking_type_putaway.id,
+            "u_push_on_drop": True,
+            "action": "push",
+        })
+
+    def setUp(self):
+        """
+        Ensure all tests start with push_from_drop set to True
+        """
+        super(PushFromDropBase, self).setUp()
+        self.push_putaway.u_push_on_drop = True
+        self.push_damaged_putaway.u_push_on_drop = True
+
+
+class TestGetPathFromLocation(PushFromDropBase):
+    """Tests for the stock.rule.get_path_from_location method in
+    isolation"""
+    def test_no_route(self):
+        """Test that get_path_from_location returns an empty recordset when no
+        matching path is found"""
+        Push = self.env['stock.rule']
+        self.push_damaged_putaway.unlink()
+        move = Push.get_path_from_location(self.received_damaged_location)
+        self.assertFalse(move)
+
+    def test_one_route(self):
+        """Test that get_path_from_location returns a single matching path when
+        one exists"""
+        Push = self.env['stock.rule']
+        move = Push.get_path_from_location(self.received_damaged_location)
+        self.assertEqual(move, self.push_damaged_putaway)
+
+    def test_many_routes(self):
+        """Test that get_path_from_location returns a single matching path when
+        more than one possible match exists, and the returned value is the one
+        for the closest parent of the search location.
+        """
+        Push = self.env['stock.rule']
+        input_zone = self.env.ref('stock.stock_location_company')
+
+        # create new push from the Input zone.
+        Push.create({
+            "name": "TestPushMultiRoute",
+            "route_id": self.route_in.id,
+            "sequence": 40,
+            "location_src_id": input_zone.id,
+            "location_id": self.location_qc_zone.id,
+            "picking_type_id": self.picking_type_putaway.id,
+            "u_push_on_drop": True,
+            "action": "push",
+        })
+        # Damaged route found correctly
+        move = Push.get_path_from_location(self.received_damaged_location)
+        self.assertEqual(move, self.push_damaged_putaway)
+
+        # Non-damaged route found correctly
+        move = Push.get_path_from_location(self.received_location)
+        self.assertEqual(move, self.push_putaway)
+
+
+class TestCreateMovesForPush(PushFromDropBase):
+    """Tests for the stock.move._create_moves_for_push method in
+    isolation"""
+    def setUp(self):
+        """Create a picking to work with for tests"""
+        super(TestCreateMovesForPush, self).setUp()
+        products_info = [{'product': self.apple, 'uom_qty': 10},
+                         {'product': self.banana, 'uom_qty': 99}]
+        self.goods_in = self.create_picking(self.picking_type_goods_in,
+                                            products_info = products_info, confirm=True, assign=True)
+        self.move_lines = self.goods_in.move_line_ids
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty})
+
+    def test_correct_move_qty(self):
+        """Test that created moves have the correct quantity"""
+        Move = self.env['stock.move']
+        moves = Move._create_moves_for_push(self.push_damaged_putaway,
+                                          self.move_lines)
+        self.assertEqual(len(moves), 2)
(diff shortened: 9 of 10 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -77,6 +80,16 @@ class StockPicking(models.Model):
         help="Destination package (used to search on pickings)",
     )
 
+    # Highlights pickings that can be deleted for unlink()
+    u_mark = fields.Boolean(
+        "Marked",
+        default=True,
+        oldname="active",
```
PeterC10, line 87: We don't need `oldname` as we're creating a new field in UDES 14.

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+                         {'product': self.banana, 'uom_qty': 99}]
+        self.goods_in = self.create_picking(self.picking_type_goods_in,
+                                            products_info=products_info)
+        self.goods_in.action_confirm()
+        self.goods_in.action_assign()
+        self.move_lines = self.goods_in.move_line_ids
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty})
+
+    def test01_correct_move_qty(self):
```
PeterC10, line 133: Suggestion: Remove the numbers from the test names, makes it awkward if we ever want to put a test between `test01` and `test02` for example.

addons/udes_stock/models/stock_rule.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+from odoo import api, fields, models, _
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    u_push_on_drop = fields.Boolean(
+        "Push on drop-off",
```
PeterC10, line 10: `Push on Drop-off` - use https://capitalizemytitle.com for strings if unsure :)

addons/udes_stock/models/stock_picking.py
```diff
@@ -77,6 +80,16 @@ class StockPicking(models.Model):
         help="Destination package (used to search on pickings)",
     )
 
+    # Highlights pickings that can be deleted for unlink()
+    u_mark = fields.Boolean(
```
PeterC10, line 84: We need to review this field, it isn't very clear what it does. (We have story 953 to investigate this, but may need to decide now...)

But I'm not sure it needed for this story, the desired outcomes only mentions creating the follow on pickings, not deleting empty ones.

Will review later and get back to you.

PeterC10:
I think these changes should be in a new module, e.g. `udes_stock_routing` - as the aim for UDES 14 is to the keep the core `udes_stock` module with only the absolute necessities.

Again will review and come back to it later tonight. :)

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+    """Tests for the stock.move._create_moves_for_push method in
+    isolation"""
+    def setUp(self):
+        """Create a picking to work with for tests"""
+        super(TestCreateMovesForPush, self).setUp()
+        products_info = [{'product': self.apple, 'uom_qty': 10},
+                         {'product': self.banana, 'uom_qty': 99}]
+        self.goods_in = self.create_picking(self.picking_type_goods_in,
+                                            products_info=products_info)
+        self.goods_in.action_confirm()
```
peterpratt13, line 127: Don't need these -can pass them as kwargs into create_picking

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+        self.goods_in.action_assign()
+        self.move_lines = self.goods_in.move_line_ids
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty})
+
+    def test01_correct_move_qty(self):
+        """Test that created moves have the correct quantity"""
+        Move = self.env['stock.move']
+        res = Move._create_moves_for_push(self.push_damaged_putaway,
```
peterpratt13, line 136: can we rename res to moves?

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+        self.assertEqual(len(res), 2)
+        for move in res:
+            self.assertEqual(move.location_id, self.received_damaged_location)
+            self.assertEqual(move.location_dest_id, self.location_qc_zone)
+
+    def test03_correct_move_orig_info(self):
+        """Test that created moves set move_orig_ids correctly"""
+        Move = self.env['stock.move']
+
+        moves = self.goods_in.mapped('move_lines')
```
peterpratt13, line 160: do we need the mapped here?

addons/udes_stock/models/stock_move.py
```diff
@@ -175,3 +175,98 @@ def split_out_move_lines(self, move_lines, **kwargs):
...
+        Creates new moves for the moves that have just dropped stock in a location.  
+        """
+        Move = self.env["stock.move"]
+        MoveLine = self.env["stock.move.line"]
+        Push = self.env["stock.rule"]
+
+        done_moves = self.filtered(lambda m: m.state == "done")
+
+        # load all the move lines, grouped by location
+        move_lines_by_location = done_moves.mapped("move_line_ids").groupby("location_dest_id")
```
peterpratt13, line 190: Do we still need the mapped?

addons/udes_stock/models/stock_move.py
```diff
@@ -175,3 +175,98 @@ def split_out_move_lines(self, move_lines, **kwargs):
...
+            if push_step not in move_lines_by_push:
+                move_lines_by_push[push_step] = MoveLine.browse()
+            move_lines_by_push[push_step] |= loc_mls
+
+        created_moves = Move.browse()
+        for push, move_lines in move_lines_by_push.items():
+            created_moves |= self._create_moves_for_push(push, move_lines)
+
+        confirmed_moves = created_moves._action_confirm()
+        pickings = confirmed_moves.mapped("picking_id")
```
peterpratt13, line 208: still need mapped?

addons/udes_stock/models/stock_move.py
```diff
@@ -175,3 +175,98 @@ def split_out_move_lines(self, move_lines, **kwargs):
...
+            move_vals = base_vals.copy()
+            move_vals.update(move._get_push_move_vals(mls))
+            created_moves |= move.copy(move_vals)
+        return created_moves
+    
+    def _action_done(self, cancel_backorder=False):
+        """
+        Extend _action_done to push from drop. 
+        """
+        done_moves = super(StockMove, self)._action_done(cancel_backorder=False)
```
peterpratt13, line 252: should it be cancel_backorder=cancel_backorder?

addons/udes_stock/models/stock_picking.py
peterpratt13: I guess this is the unlink of empty at the end of  push_from_drop. 
Could just comment it out and link to taiga story if not in scope.

addons/udes_stock/models/stock_rule.py
```diff
@@ -0,0 +1,47 @@
...
+        help="When stock is dropped in the "
+        "from_location of this path, "
+        "automatically push it onwards.",
+    )
+
+    @api.model
+    def get_path_from_location(self, location):
+        """
+        Find a single stock.rule for which the given location is
+        a valid starting location.
```
peterpratt13, line 21: Can we add - if one not found return an empty record set

addons/udes_stock/models/stock_rule.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.browse()
+
+    def _run_push(self, move):
+        """
+        Odoo has default behaviour to create another picking after a move is confirmed. Don't want this functionality
+        if u_push_on_drop is turned on.    
+        """
+        if self.u_push_on_drop:
+            return 
+        super()._run_push(move)
```
peterpratt13, line 47: nit: return super()._run_push(move)

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
+import unittest
```
peterpratt13, line 1: Is this used?

addons/udes_stock/models/stock_picking.py
PeterC10: You won't end up with empty pickings from this without refactoring though, so we shouldn't need it all for this story I think.

Will confirm with Miquel tomorrow though.

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+        self.assertEqual(new_banana_move.product_uom_qty, 99)
+
+    def test02_correct_move_locations(self):
+        """Test that created moves have the correct src and dest locations"""
+        Move = self.env['stock.move']
+        res = Move._create_moves_for_push(self.push_damaged_putaway,
+                                          self.move_lines)
+
+        self.assertEqual(len(res), 2)
+        for move in res:
```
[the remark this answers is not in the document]
PeterC10: Where's the Kevin emoji when you need it... :)

addons/udes_stock/tests/test_push_from_drop.py
PeterC10: @jrbirch99 in Odoo 14 you don't need to do mapped for relation fields (i.e. `one2many`, `many2one` and `many2many`).

`mapped` is still needed for other field types though such as `char`.

addons/udes_stock/models/stock_picking.py
peterpratt13: yep, as they are all new.

addons/udes_stock/models/stock_picking.py
```diff
@@ -393,3 +406,48 @@ def _create_move(self, move_values):
         """
         Move = self.env["stock.move"]
         return Move.create(move_values)
+
+    def find_empty_pickings_to_unlink(self):
```
adamrpatrick, line 410: Do we need this method for the push from drop functionality?

addons/udes_stock/models/stock_picking.py
```diff
@@ -393,3 +406,48 @@ def _create_move(self, move_values):
...
+            pts = PickingType.search([("u_auto_unlink_empty", "=", False)])
+            if pts:
+                domain.append(("picking_type_id", "not in", pts.ids))
+        while True:
+            records = self.search(domain, limit=1000)
+            if not records:
+                break
+            yield records
+
+    def unlink_empty(self):
```
adamrpatrick, line 433: Same here, do we need this for push from drop?

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -1,8 +1,15 @@
-from odoo import models
+from odoo import models, fields
 
 
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
+    u_auto_unlink_empty = fields.Boolean(
```
adamrpatrick, line 7: I think we also need to add the new `u_push_on_drop` field here and also add it to the picking type form

addons/udes_stock/models/stock_rule.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+from odoo import api, fields, models, _
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    u_push_on_drop = fields.Boolean(
```
adamrpatrick, line 9: This needs adding to the relevant view

addons/udes_stock/data/routes.xml
```diff
@@ -28,6 +28,7 @@
       <field name="procure_method">make_to_order</field>
       <field name="action">push</field>
       <field name="picking_type_id" ref="picking_type_putaway"/>
+      <field name="u_push_on_drop">True</field>
```
adamrpatrick, line 31: Let's have these all off for now, we'll enable specifically in another module

addons/udes_stock/models/__init__.py
```diff
@@ -13,3 +13,4 @@
 from . import stock_move
 from . import res_users
 from . import stock_picking_type
+from . import stock_rule
```
adamrpatrick, line 16: We've discussed this and come to the conclusion that the push from drop functionality should be put into its own module e.g. "udes_routing".

jrbirch99: Should this been in udes_open or udes_closed_14/addons ?

adamrpatrick: udes-open please :-)

addons/udes_stock/models/stock_move.py
caleb-shelton: @jrbirch99 To expand, instead of .mapped("x") you can write .x in most cases.

addons/udes_stock/tests/test_push_from_drop.py
jrbirch99: I just want to check I understand; I should pass `products_info` to create picking and not `products_info = products_info`?

peterpratt13: ah, so action_confirm() and action_assign() are triggered inside create_picking function, so you can do 
```self.create_picking(self.picking_type_goods_in,
                                            products_info=products_info, confirm=True, assign=True)
```
instead of calling them separately. Still pass products_info as you have though

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+            ml.write({'qty_done': ml.product_uom_qty})
+
+    def test01_picking_has_correct_src_and_dest(self):
+        """Test that pickings created by push from drop have the src and dest
+        location of the rule that created them."""
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty,
+                      'location_dest_id': self.received_damaged_location.id})
+        self.goods_in._action_done()
+        putaway = self.goods_in.u_next_picking_ids
```
caleb-shelton, line 198: This line is repeated in each test, you could abstract this/move it into setUp

caleb-shelton, line 198: You may be able to move this line from each test into the setUp function

addons/udes_stock/models/stock_move.py
```diff
@@ -175,3 +175,98 @@ def split_out_move_lines(self, move_lines, **kwargs):
...
+        done_moves.push_from_drop()
+        return done_moves
+
+    def _get_new_picking_values(self):
+        """
+        Extend _get_new_picking_values() to propagate the origin and parnter_id fields for next "push" picking.
+        """
+        values = super()._get_new_picking_values()
```
peterpratt13, line 260: nit: Sometimes we use super() and sometimes super(StockMove, self) doing the same thing, we probably only need the latter if we are mangling self or pointing it to another record set beforehand but we don't appear to be in the cases I've seen :)

addons/udes_stock/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,253 @@
...
+        self.goods_in.action_assign()
+        self.move_lines = self.goods_in.move_line_ids
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty})
+
+    def test01_picking_has_correct_src_and_dest(self):
+        """Test that pickings created by push from drop have the src and dest
+        location of the rule that created them."""
+        for ml in self.move_lines:
+            ml.write({'qty_done': ml.product_uom_qty,
```
peterpratt13, line 195: Why are we re-writing the qty_done here - we have done it already in the setUp method.
We could write the dest location here as well, then do caleb's suggestion of next_picking_ids in the setup :)

peterpratt13: We may not be able to do it in the setup method because of the unlinking in test02_no_push_rule, 
could add a method instead 
_complete_picking or something similar, and can call action_done there and return the next pickings (possibly)

addons/udes_stock/models/stock_rule.py
jrbirch99: > And also, if more are found, it is sorted by parent path? I'm not clear what this means :)

When I made the change I thought parent_path (Odoo 14) was similar to parent_left (which was used in Odoo 11), but they are quite different. In Odoo 11, the parent left value would be sorted in descending order. I guess as you said it doesn't make any sense to sort by parent path, maybe I should remove the sorted() line and just use the oldest stock.rule record?

addons/udes_stock/tests/test_push_from_drop.py
jrbirch99: No, I was able to remove it.

addons/udes_stock/models/stock_picking.py
jrbirch99: Think this is just needed when refactoring. I will remove it along with, find_empty_pickings_to_unlink(), and the u_mark field.

addons/udes_stock/models/stock_picking.py
jrbirch99: See below comment.

addons/udes_routing/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
PeterC10, line 1: We don't need to set this for Odoo 14, can we remove this from all Python files in the MR?

addons/udes_routing/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+
+from odoo import api, fields, models, _
```
PeterC10, line 3: `_` import isn't being used.

addons/udes_routing/models/stock_move.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Extend _action_done to push from drop. 
+        """
+        done_moves = super(StockMove, self)._action_done(cancel_backorder=cancel_backorder)
+        done_moves.push_from_drop()
+        return done_moves
+
+    def _get_new_picking_values(self):
+        """
+        Extend _get_new_picking_values() to propagate the origin and parnter_id fields for next "push" picking.
```
PeterC10, line 84: Typo: `parnter_id`

addons/udes_routing/views/stock_rule_views.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_stock_rule_form" model="ir.ui.view">
+        <field name="name">Update stock.rule views</field>
```
PeterC10, line 4: Suggestion: Give the view a code name which mentions the module that is overriding it. E.g. `stock.rule.form.udes_routing` - makes it easier if we ever need to debug views.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
+import logging
+
 from odoo import api, models, fields, _
 from odoo.exceptions import ValidationError
 
+_logger = logging.getLogger(__name__)
```
PeterC10, line 7: Q: Why are we adding this?

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -1,4 +1,4 @@
-from odoo import models
+from odoo import models, fields
```
PeterC10, line 1: `fields` isn't being used and should be removed.

addons/udes_routing/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,240 @@
...
+        new_apple_move = move.filtered(lambda m: m.product_id.id == self.apple.id)
+        new_banana_move = move.filtered(lambda m: m.product_id.id == self.banana.id)
+
+        self.assertEqual(new_apple_move.product_uom_qty, 10)
+        self.assertEqual(new_banana_move.product_uom_qty, 99)
+
+    def test_correct_move_locations(self):
+        """Test that created moves have the correct src and dest locations"""
+        Move = self.env['stock.move']
+        move = Move._create_moves_for_push(self.push_damaged_putaway,
```
peterpratt13, line 144: super nit moves

addons/udes_stock/models/stock_move.py
peterpratt13: This is related to
done_moves = super(StockMove, self)._action_done(cancel_backorder=cancel_backorder)
and elsewhere possibly, could be super()._action_done(cancel_backorder=cancel_backorder)?

addons/udes_routing/tests/test_push_from_drop.py
```diff
@@ -0,0 +1,240 @@
...
+        self.assertEqual(putaway.origin, new_origin)
+        self.assertEqual(putaway.partner_id, new_partner)
+
+    def _complete_picking_and_return_next_picking_id(self):
+        self._set_picking_destination_location()
+        self.goods_in._action_done()
+        return self.goods_in.u_next_picking_ids
+    
+    def _set_picking_destination_location(self):
+        for ml in self.move_lines:
```
peterpratt13, line 239: nit: Can place these writes in _complete_picking_and_return_next_picking_id as looks like only used in there

addons/udes_routing/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+{
+    "name": "UDES Routing",
+    "summary": "udes routing",
```
peterpratt13, line 4: nit: UDES should probably be shouting - but i am no authority on spelling or grammar :)

addons/udes_routing/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
...
+            [
+                (
+                    "location_src_id",
+                    "parent_of",
+                    [
+                        location.id,
+                    ],
+                ),
+                ("u_push_on_drop", "=", True),
+            ]
```
peterpratt13, line 33: we can probably add limit 1, if we are expecting only to get one. I presume we can't get more than one?

addons/udes_routing/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
...
+        "automatically push it onwards.",
+    )
+
+    @api.model
+    def get_path_from_location(self, location):
+        """
+        Find a single stock.rule for which the given location is
+        a valid starting location. If one not found return an empty recordset.
+        """
+        push_steps = self.search(
```
peterpratt13, line 23: nit  - i think convention is to do the search on the named model, Rule or StockRule in this case. 
Others prefer self.search which is fine, but i believe the  former was shouted for louder :)

addons/udes_routing/models/stock_rule.py
```diff
@@ -0,0 +1,46 @@
...
+                    [
+                        location.id,
+                    ],
+                ),
+                ("u_push_on_drop", "=", True),
+            ]
+        )
+        if push_steps:
+            return push_steps[0]
+        return self.browse()
```
peterpratt13, line 37: If only expect one can just return the record set, if there are more then not sure which we choose?

jrbirch99:
I am also changing the name of the module from udes_routing to udes_stock_routing, as discussed with Adam.

addons/udes_stock_routing/models/stock_rule.py
```diff
@@ -0,0 +1,44 @@
...
+        "automatically push it onwards.",
+    )
+
+    @api.model
+    def get_path_from_location(self, location):
+        """
+        Find a single stock.rule for which the given location is
+        a valid starting location. If one not found return an empty recordset.
+        """
+        push_steps = self.env["stock.rule"].search(
```
peterpratt13, line 21: sorry, another nit: Rule = self.env["stock.rule"] at the top then do Rule.search... below

addons/udes_stock_routing/models/stock_rule.py
```diff
@@ -0,0 +1,44 @@
...
+                    "parent_of",
+                    [
+                        location.id,
+                    ],
+                ),
+                ("u_push_on_drop", "=", True),
+            ]
+        )
+        if push_steps:
+            return push_steps.sorted(key=lambda p: p.location_src_id.parent_path.strip("/").split("/")[-1], reverse=True)[0]
```
peterpratt13, line 34: Can we add some doc strings to this please. As i can't tell what is going on, parent_path is a char field, and we format it to the last part of the path, assuming the path follows that format ( don't know if there is a constraint on it or not). Then we choose the last record sorted by the path name? 
This is just a question, i haven't used this method before so i don't understand why we choose that particular record if there are more.

addons/udes_routing/models/stock_rule.py
peterpratt13: Resolving as we need to sort a different way by the looks of things.

addons/udes_stock_routing/models/stock_rule.py
jrbirch99: Sorry I am still trying to work it all out. This is how I thought it worked before commiting this code: If there are two paths defined at different levels of the location hierarchy, for example: if we have a location hierarchy of Check, Check/CheckzoneA and Check/CheckzoneA/droplocation. When stock is dropped in Check/CheckzoneA/droplocation this method searches for the stock.rules (paths) attached to Check and CheckzoneA, if there are two paths defined on this location for example:
Check has parent_path "1/24/"
CheckzoneA has parent_path "1/24/25/"
(From what I can tell the string is always formatted in this way)
Then we sort by taking only the last number in the parent path. So after stripping and splitting:
Check would be represented by 24
CheckzoneA would be represented by 25 
We want the largest number as it is the more specific path. That why the numbers are sorted by descending order as opposed to ascending order. 

However after checking the test cases again (the test case test_many_routes), I think that in Odoo 14 it will only return the closest path e.g. it will only return CheckzoneA with parent_path "1/24/25/" and not Check with parent_path "1/24/". This is different to what happened in Odoo 11 where both parent locations where returned (in fact Odoo 11 used a completely different variable to parent path called parent_left, see: https://www.odoo.com/forum/help-1/what-is-the-meaning-of-parent-left-and-parent-right-domain-operator-25221). The sorting might not be needed when dealing with different stock.rules (paths) at different locations on the hierarchy. But, if there are multiple stock.rules defined on the same location, e.g. if we have two stock.rules defined on CheckzoneA, then I am still thinking about which stock location to choose? As it stands the newest stock.rule will be chosen. 

Perhaps there should be a flag to indicate which stock.rule should be used when two stock.rules are specified on the same location?   

I hope that all makes sense.

peterpratt13: Can we just update the doc string - i understand now :)

adamrpatrick (approved):

PeterC10 (approved):
LGTM, just a few minor things that Pete P has mentioned. Once sorted I'll be happy to tag. :)

addons/udes_stock/models/stock_move.py
jrbirch99: This should be resolved now :)

peterpratt13 (approved):

PR #562, opened 2022-02-03
[ADD] udes_ldap: UDES LDAP addon.

kdwyer:
story/955

Code and configuration for the UDES LDAP module.

UDES LDAP overrides the Odoo auth_ldap and the OCA auth_ldaps addons
to enable authentication over LDAP using the user's LDAP credentials to
bind to the LDAP server rather than a service account.

### DIFF
--- a/addons/udes_ldap/models/res_company_ldap.py
+++ b/addons/udes_ldap/models/res_company_ldap.py
@@ -1,0 +1,55 @@
+"""
+UDES LDAP configuration model.
+
+In this model we override the standard authentication approach by using the
+credentials of the logging-in user to search for themselves on the LDAP
+server.  This removes the need for a service account to perform the search.
+"""
+
+from ldap.dn import escape_dn_chars
+from odoo import models, fields
+
+
+class CompanyLDAP(models.Model):
+    """LDAP configuration model."""
+
+    _inherit = "res.company.ldap"
+
+    u_ldap_binddn_fmt = fields.Char(
+        string="LDAP Binddn Format",
+        help="A format string for the Distinguished Name used to bind to the "
+        "LDAP server.  The user's uid will be formatted into this string.",
+        default="",
+        required=True,
+    )
+
+    def get_ldap_dicts(self):
+        """Add our custom field to the conf dict and removed unused fields."""
+        res = super().get_ldap_dicts()
+        fields_to_remove = ["ldap_binddn", "ldap_password"]
+        for rec in res:
+            ldap = self.sudo().browse(rec["id"])
+            rec["u_ldap_binddn_fmt"] = ldap.u_ldap_binddn_fmt
+            for field in fields_to_remove:
+                rec.pop(field, None)
+        return res
+
+    def authenticate(self, conf, login, password):
+        """
+        Set the login credentials before the superclass authenticates.
+
+        We use the login credentials provided by the user to bind to the LDAP
+        server.
+        """
+        # The login name is untrusted and must be escaped to prevent
+        # LDAP-injection attacks.
+        escaped_login = escape_dn_chars(login)
+        # Use %-formatting to be consistent with ldap_filter, as we don't
+        # want to change python-ldap.
+        binddn = conf["u_ldap_binddn_fmt"] % escaped_login
+        conf["ldap_binddn"] = binddn
+        conf["ldap_password"] = password
+        # We don't pass the escaped login because it gets escaped when the
+        # filter is created.
+        # TODO Does user creation work if login contains exotic characters?
+        return super().authenticate(conf, login, password)
--- a/addons/udes_ldap/views/res_company_ldap_views.xml
+++ b/addons/udes_ldap/views/res_company_ldap_views.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <menuitem id="udes_ldap.ldap_configuration_action"
+      parent="base.menu_users"
+      action="auth_ldap.action_ldap_installer"
+      sequence="40"
+      groups=""
+      name="LDAP Servers"
+    />
+
+    <record id="view_ldap_installer_form" model="ir.ui.view">
+        <field name="name">udes_ldap.res.company.ldap.form</field>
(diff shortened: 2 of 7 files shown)


addons/udes_ldap/models/res_company_ldap.py
```diff
@@ -0,0 +1,52 @@
+"""
+UDES LDAP configuration model.
+
+In this model we override the standard authentication approach by using the
+credentials of the logging-in user to search for them selves on the LDAP
```
ws-mps, line 5: typo: them selves

addons/udes_ldap/models/res_company_ldap.py
```diff
@@ -0,0 +1,52 @@
...
+from ldap.dn import escape_dn_chars
+from odoo import models, fields
+
+
+class CompanyLDAP(models.Model):
+    _inherit = "res.company.ldap"
+
+    u_ldap_binddn_fmt = fields.Char(
+        string="LDAP binddn format",
+        help="A format string for the Distinguished Name used to bind to the"
```
ws-mps, line 18: Q: Do we need space at the end of line?

addons/udes_ldap/models/res_company_ldap.py
```diff
@@ -0,0 +1,52 @@
...
+        required=True,
+    )
+
+    def get_ldap_dicts(self):
+        """Add our custom field to the conf dict and removed unused fields."""
+        res = super().get_ldap_dicts()
+        for rec in res:
+            ldap = self.sudo().browse(rec["id"])
+            rec["u_ldap_binddn_fmt"] = ldap.u_ldap_binddn_fmt
+            del rec["ldap_binddn"]
```
ws-mps, line 30: Is this "list" ever going to change? Do we want to use a list + loop instead of `del` twice?

addons/udes_ldap/views/res_company_ldap_views.xml
```diff
@@ -0,0 +1,36 @@
...
+      groups=""
+      name="LDAP Servers"
+    />
+
+    <record id="view_ldap_installer_form" model="ir.ui.view">
+        <field name="name">udes_ldap.res.company.ldap.form</field>
+        <field name="model">res.company.ldap</field>
+        <field name="inherit_id" ref="auth_ldaps.view_ldap_installer_form"/>
+        <field name="arch" type="xml">
+            <!-- Hide inherited fields for login informtion. -->
```
ws-mps, line 22: typo: informtion

addons/udes_ldap/views/res_company_ldap_views.xml
```diff
@@ -0,0 +1,36 @@
...
+          <xpath expr="//field[@name='ldap_binddn']" position="attributes">
+              <attribute name="invisible">1</attribute>
```
ws-mps, line 24: Do we want it/them invisible or just remove it completely from the view?

addons/udes_ldap/models/res_company_ldap.py
kdwyer: There could be more than one LDAP configured, in theory.  Or are you suggesting an inner loop to drop the keys that we don't want?

Another option might be to mark those fields as [deprecated](https://github.com/unipartdigital/odoo/blob/ee08a6f63711c950bc6c3eae61fd7f959824c0c4/odoo/fields.py#L310), which would [prevent them being retrieved](https://github.com/unipartdigital/odoo/blob/ee08a6f63711c950bc6c3eae61fd7f959824c0c4/odoo/fields.py#L446), but that doesn't seem to be very common :grimacing:

addons/udes_ldap/models/res_company_ldap.py
kdwyer: Fixed

addons/udes_ldap/models/res_company_ldap.py
kdwyer: Yes we do!  Fixed.

addons/udes_ldap/views/res_company_ldap_views.xml
kdwyer: Fixed.

addons/udes_ldap/models/res_company_ldap.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo import models, fields
+
+
+class CompanyLDAP(models.Model):
+    """LDAP configuration model."""
+
+    _inherit = "res.company.ldap"
+
+    u_ldap_binddn_fmt = fields.Char(
+        string="LDAP binddn format",
```
jrbirch99, line 19: Does this string need to be capitalised: "LDAP Binddn Format"? I understand it is going to be invisible in the view, but maybe just for consistency?

kdwyer: I _think_ the logic behind the casing on this model seems to be rather inconsistent, for example "LDAP Server address", "LDAP base", "Create User", so I'm not sure what's best here.  I'm inclined to leave as-is, but I'm open to persuasion.

addons/udes_ldap/views/res_company_ldap_views.xml
kdwyer: Removed

addons/udes_ldap/models/res_company_ldap.py
ws-mps: Yes, I was suggesting inner loop to drop the keys we don't want.

addons/udes_ldap/models/res_company_ldap.py
ws-mps: Ideally all strings should be properly capitalised, so would be good to do it for new fields even though the old ones are not consistent with it.

kdwyer: Persuaded :smile:  and fixed.

addons/udes_ldap/models/res_company_ldap.py
kdwyer: Done

ws-mps (approved):

adamrpatrick (approved):

PR #563, opened 2022-02-04
Story/606

PeterAlabaster:
fixes to udes_common helper function def batched()
ported edi_notifier

see commit messages for detail


kdwyer (approved):

kdwyer:
Great commit messages :+1:

addons/udes_common/tools/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+    return wrapper
+
+
+def add_even_if_exists(cls):
+    """Patch class to override an existing method"""
+
+    def wrapper(func):
+        # pylint: disable=missing-docstring
+        if hasattr(cls, func.__name__):
+            _logger.warning("%s.%s is being overridden" % (cls.__name__, func.__name__))
```
PeterAlabaster, line 27: self nit: could add parameter to decorator to determine whether we want to log a warning if the func already exists.
Default to true and if set to false explicitly on implementation (i.e for the odoo core func in edi that gets overridden) then it only logs out a .info
```python
@tools.decorators.add_even_if_exists(log_warning=False)
```

adamrpatrick (approved):

PR #564, opened 2022-02-08
[ADD] udes_get_info: Get info, Review usefulness and port

armandcela:
Story: 51

### DIFF
--- a/addons/udes_get_info/models/models.py
+++ b/addons/udes_get_info/models/models.py
@@ -1,28 +1,33 @@
-# -*- coding: utf-8 -*-
-from odoo import models, _
+from odoo import fields, models, _
 from odoo.addons.udes_common.models import add_if_not_exists
 import logging
 
 _logger = logging.getLogger(__name__)
 
 
-# N.B: Basic values should not have any relations as this is to stop the recursion
+# N.B: Basic values that some models have, no need to add them in _get_info_field_names
 BASIC_GET_INFO_VALUES = {"id", "name", "display_name"}
 
 
 @add_if_not_exists(models.BaseModel)
-def _get_info(self, level=0, max_level=None, fields=None):
-
+def _get_info(self, level=1, max_level=None, info_fields=None, extra_fields=None):
+    """
+    Returns the information of the current object. If level equals max_level it will return only
+    BASIC_GET_INFO_VALUES of any relational field of the current object. If level is less than
+    max_level it will return the specified fields of any relational fields of the current object.
+    """
     self.ensure_one()
 
     if not max_level:
         max_level = self._get_info_max_levels
-    if not fields:
-        fields = self._get_info_field_names
-    level_and_max = {"level": level + 1, "max_level": max_level}
+    next_level_and_max = {"level": level + 1, "max_level": max_level}
+    if extra_fields is None:
+        extra_fields = set()
+    if info_fields is None:
+        info_fields = self._get_info_field_names | BASIC_GET_INFO_VALUES | extra_fields
 
     info = {}
-    for field_name in fields:
+    for field_name in info_fields:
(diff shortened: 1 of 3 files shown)


addons/udes_common/models/models.py
```diff
@@ -105,3 +105,59 @@ def selection_display_name(self, selection_field_name):
     # Odoo has a nice function that handles all the weird ways you can define selections!
     selection_values_dict = dict(selection_field._description_selection(self.env))
     return selection_values_dict.get(self[selection_field_name], False)
+
+
+# N.B: Basic values should not have any relations as this is to stop the recursion
```
armandcela, line 110: Will change the comment later , Basic values are not to stop the recursion , max_level stops the recursion

addons/udes_get_info/models/models.py
```diff
@@ -1,46 +1,59 @@
...
 from odoo.addons.udes_common.models import add_if_not_exists
 import logging
 
 _logger = logging.getLogger(__name__)
 
 
-# N.B: Basic values should not have any relations as this is to stop the recursion
-BASIC_GET_INFO_VALUES = {"id", "name", "display_name"}
+# N.B: Basic values that some models have, no need to add them in _get_info_field_names
+BASIC_GET_INFO_VALUES = {"id", "name", "display_name", "code", "barcode"}
```
ws-mps, line 9: Code and barcode are not in all models, why we add them here?

armandcela: Barcode in some model which we use a lot are(location, product), code generally may be used like a identifier in models in odoo, Advantage: will be no need to add in model fields , Disadvantage: Will log in server for the models that is not.

What do you you think is better, happy to remove and to add barcode in location only at the moment

addons/udes_get_info/models/models.py
```diff
@@ -1,46 +1,59 @@
...
+def _get_info(self, level=0, max_level=None, info_fields=None, extra_fields=None):
 
     self.ensure_one()
 
     if not max_level:
         max_level = self._get_info_max_levels
-    if not fields:
-        fields = self._get_info_field_names
     level_and_max = {"level": level + 1, "max_level": max_level}
+    if not extra_fields:
```
ws-mps, line 20: if extra_fields is None:

addons/udes_get_info/models/models.py
```diff
@@ -1,46 +1,59 @@
...
+        extra_fields = set()
+    if not info_fields:
```
ws-mps, line 22: Same as above.

addons/udes_get_info/models/models.py
ws-mps: I prefer to avoid warnings and we add it to the specific models.

caleb-shelton (approved):
LGTM once Miquel's feedback is addressed/resolved.

adamrpatrick (approved):

addons/udes_get_info/models/models.py
```diff
@@ -10,13 +10,16 @@
 
 
 @add_if_not_exists(models.BaseModel)
-def _get_info(self, level=0, max_level=None, info_fields=None, extra_fields=None):
-
+def _get_info(self, level=1, max_level=None, info_fields=None, extra_fields=None):
+    """Returns the information of the current object in the case when level equals to max_level
+    it will return only BASIC_GET_INFO_VALUES of any relational field of current object.
+    In the case level is less than max_level it will return the specified fields of any
+    relational fields of the current object."""
```
adamrpatrick, line 17: Perhaps: "Returns the information of the current object. If level equals max_level
     it will return only BASIC_GET_INFO_VALUES of any relational field of the current object.
     If level is less than max_level it will return the specified fields of any
     relational fields of the current object."

PR #567, opened 2022-02-11
Story #910 Standard Interfaces - Port Item Master

jrbirch99:
I added fields to the product template that are included in UDES 11 and needed for edi item master interface.


addons/udes_stock/data/product_categories.xml
```diff
@@ -0,0 +1,18 @@
...
+    </record>
+
+    <!-- Construct "Speed" product category -->
+    <record id="product_category_speed" model="product.category">
+      <field name="name">Speed</field>
+      <field name="parent_id" ref="product.product_category_all"/>
+    </record>
+
+  </data>
+</odoo>
```
nickcosta85, line 18: Nit:Newline at the end of file

addons/udes_stock/views/product_template.xml
```diff
@@ -0,0 +1,14 @@
...
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']" position="inside">
+                <group string="UDES" name="inventory_udes">
+                </group>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
nickcosta85, line 14: Nit:Newline at the end of file

addons/udes_stock/models/product_template.py
```diff
@@ -14,3 +21,16 @@ class ProductTemplate(models.Model):
...
     type = fields.Selection(default="product")
+
+    u_speed_category_id = fields.Many2one(
+        comodel_name="product.category",
+        domain=_domain_speed_category,
+        string="Product Category Speed",
+        help="Product category speed to match with location speed.",
+    )
+
+    u_height = fields.Float(string="Height", help="Product height in m", default=0.0)
```
PeterC10, line 32: Two suggestions:

1: The unit of measure is important when viewing the field, so I'd suggest setting the string to `Height (m)`.

2: Change the help string, I assume `m` is for metres? Best to say the explicitly in the help to be explicit.

addons/udes_stock/views/product_template.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="view_product_template_form_view_udes" model="ir.ui.view">
+        <field name="name">Changes to product form</field>
```
PeterC10, line 4: Suggestion: `product.template.common.form.udes_stock`

addons/udes_stock/views/product_template.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="view_product_template_form_view_udes" model="ir.ui.view">
```
PeterC10, line 3: Suggestion: `view_product_template_form_view_udes_stock`

addons/udes_stock/views/product_template.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="name">Changes to product form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='inventory']" position="inside">
+                <group string="UDES" name="inventory_udes">
```
PeterC10, line 9: We need to move away from putting `UDES` in the group strings. The system the end user is using is UDES, so they get UDES funtionality. :)

Maybe just `Dimensions`?

addons/udes_stock/models/product_template.py
```diff
@@ -14,3 +21,16 @@ class ProductTemplate(models.Model):
 
     # Default to being a stockable product
     type = fields.Selection(default="product")
+
+    u_speed_category_id = fields.Many2one(
+        comodel_name="product.category",
+        domain=_domain_speed_category,
+        string="Product Category Speed",
+        help="Product category speed to match with location speed.",
```
PeterC10, line 29: This field is added in code but not the view, is it needed?

We could also do with making the help string a bit clearer.

addons/udes_stock/models/product_product.py
```diff
@@ -11,3 +10,29 @@ class ProductProduct(models.Model):
 
     # Add tracking for archiving.
     active = fields.Boolean(tracking=True)
+
+    def sync_active_to_templates(self, activate_templates=True, deactivate_templates=True):
```
PeterC10, line 14: We've added this method but it isn't being called anywhere, is it needed for this story?

addons/udes_stock/data/product_categories.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+
+    <!-- Construct "Height" product category -->
+    <record id="product_category_height" model="product.category">
```
PeterC10, line 6: Q: Why do we have height categories but not length or width?

addons/udes_stock/models/product_template.py
```diff
@@ -14,3 +14,19 @@ class ProductTemplate(models.Model):
 
     # Default to being a stockable product
     type = fields.Selection(default="product")
+
```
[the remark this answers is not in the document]
PeterC10: I can understand the reasoning sometimes. For example the `type` override hasn't go anything to do with the dimension fields added in this story, so I can see the argument for having a line between those.

But certainly `u_height`, `u_length` etc should be one after the other at least.

addons/udes_stock/models/product_template.py
jrbirch99: It is needed, the speed category is assigned in the item master record class. The view should have been included in `product_template.xml`, I just missed it. About rewording the help string - I am not quite sure what the speed refers to? Is it the speed with which the product can be transported across the warehouse?

addons/udes_stock/models/product_product.py
jrbirch99: This is called in the `execute` method inside `UDESItemMasterRecord`.

addons/udes_stock/data/product_categories.xml
jrbirch99: I think it is because in UDES 11 some customers (hds, anniversary) want height categories, but not length/width categories. The height categories are only used by specific customers.

jrbirch99: It is only used for specific customers. Some customer modules use the height category, I assume those customers do not want length or width categories.

jrbirch99:
As discussed:
- The height category is being removed.
- The speed categories are being moved from udes_rockhopper to udes_stock. 
- The README is being updated with new added fields.

addons/udes_stock/data/product_categories.xml
PeterC10: Just for the benefit of anyone who checks after this is removed, the "Height" category is actually meant to reference the pallet height.

But isn't needed in v14 for now, so will be removed as Jack has said elsewhere. :)

armandcela (approved):

PeterC10 (approved):
LGTM, thanks for updating the readme! :)

PR #568, opened 2022-02-11
[ADD] udes_permissions: New module with User Templates model

PeterAlabaster:
And Hierarchical tree view module

User-story: 943


addons/udes_permissions/models/res_groups.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+        UserTemplate = self.env["user.template"]
+
+        if "name" in vals and vals.get("name"):
+            # @Performance: Ideally we would have a link back directly, but because
+            # the relationship is effectively one2one, we would require a funky hack like this one:
+            # http://blog.odoobiz.com/2014/10/openerp-one2one-relational-field-example.html
+            template_for_group = UserTemplate.search([("template_group_id", "=", self.id)])
+            if template_for_group and not self.env.context.get(
+                "bypass_name_propagation_to_template"
```
peterpratt13, line 20: do we want this not condition moved into the upper if to avoid the search?

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+                len(self.TEMPLATE_GROUP_PREFIX) :
+            ]
+        else:
+            # We don't really care if it fails as it is only a name, but log warn it anyway
+            _logger.warning(
+                f"""Could not determine User Template name from Group name '{group_name}'
+                using prefix '{self.TEMPLATE_GROUP_PREFIX}'.
+                Not propagating change to User Template '{self.name}' (id: {self.id})"""
+            )
+            return False
```
peterpratt13, line 219: We return False in the else, and just have None otherwise, is this deliberate. Or do we want to return True at the end?

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Generate the full user template group name, with prefix
+        :param: basename: str() of the template name
+        """
+        return f"{self.TEMPLATE_GROUP_PREFIX}{base_name}"
+
+    def _set_user_template_group_name(self, group_name):
+        """
+        Given a group name which includes the prefix,
+        update the name of the template by removing the prefix from group name,
```
peterpratt13, line 203: nit base name (maybe) as that seems to be the ref else where

addons/udes_permissions/tests/test_user_template.py
```diff
@@ -0,0 +1,250 @@
...
+        # (but we surpress the warn here)
+        template_name_before_invalid_group_name_change = template.name
+        # Surpress warnings when doing this as the warnings are expected
+        with mute_logger("odoo.addons.udes_permissions.models.user_template"):
+            group.name = group.name.replace(template.TEMPLATE_GROUP_PREFIX, "")
+        self.assertEqual(template_name_before_invalid_group_name_change, template.name)
+
+    def test_generated_template_group_id(self):
+        """Ensure a group is generated following the naming scheme (for all of our templates)"""
+        for template in self.all_user_templates:
```
peterpratt13, line 113: nit: subTest

peterpratt13:
first comit looks good, will look at the rest in a bit

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+        :param: groupname: str() of the group name
+                           i.e return value of _get_user_template_group_name()
+        """
+        if group_name.startswith(self.TEMPLATE_GROUP_PREFIX):
+            self.with_context(dict(bypass_name_propagation_to_group=True)).name = group_name[
+                len(self.TEMPLATE_GROUP_PREFIX) :
+            ]
+        else:
+            # We don't really care if it fails as it is only a name, but log warn it anyway
+            _logger.warning(
```
kdwyer, line 214: Technical nit: the f-string will always be created, whereas with `logger.warning('foo %s', 'bar')` the interpolation only occurs if we are actually going to emit a log message.  It's probably not much overhead here - and it's unlikely that warning-level messages would be suppressed - but worth being aware.

addons/udes_hierarchical_tree_view/.gitignore
```diff
@@ -0,0 +1,128 @@
+# Byte-compiled / optimized / DLL files
```
PeterC10, line 1: Should this be here? I haven't checked elsewhere but usually `.gitignore` is only at the root of the repo.

addons/udes_hierarchical_tree_view/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
PeterC10, line 1: Missing newline klaxon.

addons/udes_hierarchical_tree_view/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+        UDES Hierarchical Tree View
+    """,
+    "description": """
+         UDES Hierarchical Tree View
+    """,
+    "author": "Unipart digital",
+    "website": "http://github/unipartdigital/udes-open",
+    "category": "Tools",
+    "version": "0.1",
+    "depends": ["base", "web", "mail", "product"],
```
PeterC10, line 13: Q: Is this `product` dependency temporary? (Considering your comment in `product_views.xml`.

Also a bit unsure if the `mail` dependency is needed?

addons/udes_hierarchical_tree_view/static/src/components/tree_item/TreeItem.js
```diff
@@ -0,0 +1,87 @@
+odoo.define(
+  "udes_hierarchical_tree_view/static/src/components/tree_item/TreeItem.js",
+  function (require) {
+    "use strict";
+    const { Component } = owl;
+    const patchMixin = require("web.patchMixin");
+    const { useState } = owl.hooks;
+    const HierarchicalTreeController = require("udes_hierarchical_tree_view.HierarchicalTreeController")
```
PeterC10, line 8: This isn't being used, is it needed?

addons/udes_hierarchical_tree_view/static/src/components/tree_item/TreeItem.js
```diff
@@ -0,0 +1,87 @@
...
+        let droppedItem = JSON.parse(event.dataTransfer.getData("TreeItem"));
+        if (
+          droppedItem.id == this.props.item.id ||
+          droppedItem.parent_id[0] == this.props.item.id
+        ) {
+          console.log("Drop inside itself or same parent has no effect");
+          return;
+        }
+        if (this.props.item.parent_path.startsWith(droppedItem.parent_path)) {
+          console.log("Oops, drop inside child item is forbidden.");
```
PeterC10, line 63: Q: Are these just for debug? (Basically can we lose them before merging?) If we're going to keep them might be worth rewording a bit.

addons/udes_hierarchical_tree_view/static/src/hierarchical_tree_view/hierarchical_tree_controller.js
```diff
@@ -0,0 +1,89 @@
...
+        if (form_views.length == 1) {
+          await this.model.do_action({
+            "res_id": ev.data.data.id,
+            "res_model": this.model.modelName,
+            "type": "ir.actions.act_window",
+            "views": [[form_views[0].viewID, "form"]],
+            'context': this.model.loadParams.context,
+          })
+        } else {
+          // TODO(palabaster): Log warning/alertbox, no form view defined.
```
PeterC10, line 58: Q: Does this mean a form view has to be defined in the action? (i.e. can't fall back to the default form view, the one with the lowest sequence)

Not an issue at all for this story, but might be worth raising a low priority issue for future. :)

addons/udes_permissions/__manifest__.py
```diff
@@ -11,7 +11,10 @@
     "category": "Uncategorized",
     "version": "0.1",
     # any module necessary for this one to work correctly
-    "depends": ["base"],
+    "depends": [
+        "base",
+        "udes_hierarchical_tree_view",
```
PeterC10, line 16: Might be worth considering making `udes_common` depend on `udes_hierarchical_tree_view`, and adding a dependency to `udes_common` instead.

addons/udes_permissions/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "udes_permissions",
+    "summary": """
+        UDES Permissions
+    """,
+    "description": """
+         UDES Permissions
+    """,
+    "author": "UDES",
```
PeterC10, line 9: `Unipart Digital`

addons/udes_permissions/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "udes_permissions",
```
PeterC10, line 2: `UDES Permissions`

addons/udes_permissions/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "summary": """
```
PeterC10, line 3: Q: Do we need both summary and description? I can't remember... will come back in the morning. :P

Could also borrow a few words from the readme to use.

addons/udes_permissions/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        UDES Permissions
+    """,
+    "description": """
+         UDES Permissions
+    """,
+    "author": "UDES",
+    "website": "",
+    "category": "Uncategorized",
```
PeterC10, line 11: We should probably set this to something else, lowest of low priorities though.

addons/udes_permissions/models/res_groups.py
```diff
@@ -0,0 +1,24 @@
...
+        If we change the name of a group directly linked to a User Template,
+        try to propagate the name change to the User Template
+        """
+        UserTemplate = self.env["user.template"]
+
+        if "name" in vals and vals.get("name"):
+            # @Performance: Ideally we would have a link back directly, but because
+            # the relationship is effectively one2one, we would require a funky hack like this one:
+            # http://blog.odoobiz.com/2014/10/openerp-one2one-relational-field-example.html
+            template_for_group = UserTemplate.search([("template_group_id", "=", self.id)])
```
PeterC10, line 18: Might be worth considering adding a field to link directly to the template from the group, not a big deal at this early stage though.

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+
+    If you are utilizing this model, you should only need to use the following functions;
+    - def get_template_groups()
+    - def get_template_descendants()
+    instead of directly access the fields on here.
+    """
+
+    _name = "user.template"
+    _description = "User Template"
+    _rec_name = "complete_name"
```
PeterC10, line 28: Q: Is `rec_name` set to `complete_name` to help with the hierarchy view? If you have more than a few levels then this can get difficult to read pretty quickly.

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+        groups here, along with all child templates groups will
+        be considered in get_template_groups()""",
+    )
+    # NB: We don't cascade on parent as it won't call the unlink() hook where cascaded
+    parent_id = fields.Many2one(
+        "user.template",
+        index=True,
+        string="Parent",
+        help="""The Parent Template of this User Template.
+        Allows us to create a hierarchy of templates.
```
PeterC10, line 66: Suggestion: `Allows for a hierarchy of templates to be created`

addons/udes_permissions/models/user_template.py
```diff
@@ -0,0 +1,303 @@
...
+        This can be used as an interim until more granular groups are available.
+        The group set here will be bundled in with all other groups as part of get_template_groups()""",
+    )
+
+    # The three following computes all use the same method to save searching child_of
+    # on three separate occasions, as they all hinge off the same initial 'family'
+    family_descendant_ids = fields.Many2many(
+        "user.template",
+        compute="_compute_family_column_ids",
+        store=False,  # We cannot store as function won't recompute if indirect descendants groups change
```
PeterC10, line 102: I don't mind being explicit, but `store=False` is the default. Is this to make sure no one tries this? :D

addons/udes_permissions/views/user_template_view.xml
```diff
@@ -0,0 +1,99 @@
...
+                <form string="User Template">
+                    <sheet>
+                        <!-- Here we have a set of smart buttons that use the 'recursive_child_*' columns
+                        in combination with actions defined on user.template to display the relevant information -->
+                        <div name="button_box" class="oe_button_box">
+                            <button name="action_view_child_templates" type="object" class="oe_stat_button" icon="fa-sitemap">
+                                <field name="family_descendant_count" widget="statinfo" string="Child Templates" />
+                            </button>
+                            <button name="action_view_child_groups" type="object" class="oe_stat_button" icon="fa-users">
+                                <field name="family_group_count" widget="statinfo" string="Implied Groups" />
```
PeterC10, line 17: (Just a comment to remind myself)

Not 100% sure on using `Implied`. Obviously it is implied for the template, but it isn't necessarily implied for the user that the template is applied to. Super low priority though and I can't think of an alternative atm anyway. :)

addons/udes_permissions/views/user_template_view.xml
```diff
@@ -0,0 +1,99 @@
...
+                            <label for="name" class="oe_edit_only" />
+                            <h1>
+                                <field name="name" placeholder="e.g. Stock Manager" />
+                            </h1>
+                        </div>
+                        <!-- Main fields for user template -->
+                        <group>
+                            <group name="basic">
+                                <!-- This field is required at the model level, but we set it to not required
+                                to prevent form validation kicking in before one can be generated and set -->
```
PeterC10, line 34: Would have thought there'd be a nicer way to get around this by now. The workaround isn't too bad, but does always look a bit odd.

PeterC10:
Liking what I see so far very much, will take a closer look tomorrow but great job so far. :)

addons/udes_permissions/models/user_template.py
PeterAlabaster: Good spot thanks, i'll remove the return

addons/udes_permissions/models/user_template.py
PeterAlabaster: so `_get_user_template_group_name()` takes the template name i.e 'Administrator', whereas `_set_user_template_group_name()` takes the template with prefix i.e `User Template Group: Administrator` which is why i've called this param group_name rather than base_name as they are intentionally different things

PeterAlabaster: this has made me realise i need to update the function name to `_set_user_template_name()` :+1:

addons/udes_hierarchical_tree_view/__manifest__.py
PeterAlabaster: product is temporary, and you're right looks like mail is not necessary as a dep

addons/udes_hierarchical_tree_view/static/src/components/tree_item/TreeItem.js
PeterAlabaster: I'll look at rewording any console logs, these were unchanged from the original module

addons/udes_hierarchical_tree_view/static/src/hierarchical_tree_view/hierarchical_tree_controller.js
PeterAlabaster: We could always change that to make an RPC to get the default form on view initialization, but i didn't know off the top of my head how that is done

addons/udes_permissions/models/res_groups.py
PeterAlabaster: I did give it a go but it just felt ugly doing the hack specified in the comment. Unless i'm being dumb and forgetting how to do a o2o properly

addons/udes_permissions/models/user_template.py
PeterAlabaster: Yeah it's basically to make display_name based on the full hierarchy. Happy to change it back to just name if we would prefer

addons/udes_permissions/models/user_template.py
PeterAlabaster: yeah it's more to explicitly prevent anyone from ever trying it in future because it will not work :+1:

addons/udes_hierarchical_tree_view/static/src/hierarchical_tree_view/hierarchical_tree_controller.js
PeterAlabaster: i've added the error now, only developers should ever see it, if we add it to a new view and we don't do the above in xml they will get a warn in the ui, but adding to the xml should always result in one being found. the RPC thing could be a future improvement though so we don't have to do it

addons/udes_permissions/__manifest__.py
PeterC10: Sorry still needs changing, `Digital` instead of `digital`. 😅

addons/udes_permissions/models/res_groups.py
PeterC10: Let's leave it for now then, can always address in the following story if actually needed. :)

addons/udes_permissions/views/user_template_view.xml
PeterC10: Let's leave it as if... if you're managing user templates then you should be knowing this. And at the moment the parent templates inherit all groups the child has, so should be fine. :)

addons/udes_permissions/models/user_template.py
PeterAlabaster: we can do that for the tree view as we can just display a different column, but not the hierarchical tree view as that uses display_name in the widget

PeterC10: Okay understood. Think there'll be some cases we do just want the name to display, but we can tackle that in the relevant future stories. :)

adamrpatrick (approved):

PR #570, opened 2022-02-14
Story/934 unlink

kdwyer:


peterpratt13 (approved):

addons/udes_stock/tests/test_product_template.py
```diff
@@ -0,0 +1,26 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        cls.apple_template = cls.apple.product_tmpl_id
+        cls.red_apple = cls.apple
+
+    def test_raises_error_on_deletion(self):
+        """Unlinking a product template should raise an error."""
+        expected_message = "Products may not be deleted. Please archive them instead."
```
armandcela, line 20: Nit: I tried to make a suggestion the only one though maybe was to make the expected message a constant and to import in unit tests. In case that the message will change no need to change it in unit tests too.

kdwyer: I think it's better that the message is defined separately in the tests.  That way we can be sure that any change to the message is definitely intentional.

armandcela: Agree, thats happen when i hardly try to add a thread :)

PeterC10:
Closed, please see: #572

PeterC10:
See #572

PR #572, opened 2022-02-15
Story/934 product enhancements

PeterC10:
Raise an error and recommend archiving rather than deletion if an attempt is made to delete a product.

Set "active_test" to False for "product_variant_ids". This means related fields on product.template (like barcode) will be preserved even when the template and product are archived.

### DIFF
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -121,3 +121,8 @@ class ProductProduct(models.Model):
         if activate_templates:
             templates_to_activate = templates.filtered(lambda t: t.product_variant_ids)
             templates_to_activate.write({"active": True})
+
+    @api.multi
+    def unlink(self):
+        """Override superclass to prevent deletion."""
+        raise ValidationError(_("Products may not be deleted. Please archive them instead."))
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -102,24 +102,24 @@ class ProductProduct(models.Model):
         Product templates will be active if any of their product variants are
         active, including product.products not present in self.
         """
-        ProductTemplate = self.env["product.template"]
-
         self = self.with_context(prefetch_fields=False)
         templates = self.mapped("product_tmpl_id")
 
         # Prefetch fields
         templates.mapped("product_variant_ids")
 
-        # Activate or deactivate templates based on product_variant_ids.
-        # product_variant_ids only contains active products because Odoo
-        # automatically adds active=True to its domain.
+        # Activate or deactivate templates based on whether it has active product_variant_ids.
 
         if deactivate_templates:
-            templates_to_deactivate = templates.filtered(lambda t: not t.product_variant_ids)
+            templates_to_deactivate = templates.filtered(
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/product_product.py
```diff
@@ -102,24 +102,24 @@ def sync_active_to_templates(self, activate_templates=True, deactivate_templates
...
-            templates_to_deactivate = templates.filtered(lambda t: not t.product_variant_ids)
+            templates_to_deactivate = templates.filtered(
+                lambda t: not t.product_variant_ids.filtered("active")
+            )
             templates_to_deactivate.write({"active": False})
 
         if activate_templates:
-            templates_to_activate = templates.filtered(lambda t: t.product_variant_ids)
+            templates_to_activate = templates.filtered(
+                lambda t: t.product_variant_ids.filtered("active")
```
peterpratt13, line 121: Does this mean if the product variants are n't active, then we cannot activate a template. Should we do it so that if the template has variants, we activate the template and all its variants?

armandcela: I think that it means that in order to activate a template should at least have a variant activated

PeterC10: This is how it works already I think, the second test here: https://github.com/unipartdigital/udes-open/blob/master/addons/udes_stock/tests/test_product.py

But let me know if I misunderstood. :)

But as far as I know if the product template doesn't have an active variant, then it shouldn't be active.

armandcela (approved):

armandcela (approved):

addons/udes_stock/models/product_product.py
peterpratt13: ah ok - so no template can be deactivated if a variant is active. And vice versa. 

I was wondering if a template and its variants could be deactivated and then when reactivating you couldn't because no variants are active

PeterC10: Yep exactly.

And this method is only called as a safety measure in places where it is needed (EDI for example).

But yeah you need at least 1 active variant for the template to be active in this check.

FWIW our code on the v11 branches basically assumes 1 template to 1 variant, a bit like UoM. But for 14 it is something that needs more consideration. Thankfully from what Kevin found, it seems like barcodes are handled a bit more nicely in v14.

addons/udes_stock/tests/__init__.py
```diff
@@ -33,3 +33,4 @@
 from . import test_stock_location_picking_zone
 from . import test_location_countable
 from . import test_location_state
+from . import test_product_template  # noqa: F401
```
PeterAlabaster, line 36: don't all of the other `from . import ...` get imported but unused? is there a reason it's only on test_product_template?

adamrpatrick (approved):

PeterAlabaster (approved):

addons/udes_stock/tests/__init__.py
kdwyer: I only changed (added) that line; changing all the others seems like a lot of noise, especially as we don't have a consistent linting policy atm.

PR #573, opened 2022-02-25
Story/1088 - Rename packaging module and tidy up module manifests/Python files

PeterC10:
"Advanced" terminology dropped on packaging module, renamed to fit in with other stock modules.

Module manifests updated so all authors now say "Unipart Digital".

Removed utf-8 coding at top of Python files where set, as this is not needed.


addons/edi_notifier/__init__.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
kdwyer, line 1: Bit of a nit, but an initial blank line in a file seems wrong somehow (this applies to a number of files)?

kdwyer (approved):
One nit but LGTM

LoKKeR96 (approved):
LGTM

adamrpatrick (approved):

addons/edi_notifier/__init__.py
PeterC10: Yes agreed, I used a reg-ex that would remove the line with utf-8, not just empty it. But I didn't consider that some files with this set also left a blank line underneath. Have removed.

PR #575, opened 2022-02-28
[ADD] udes_stock: Add method to set orig picking on pickings

PeterC10:
Added "_set_original_picking_on_historical_pickings" which identifies
any pickings in the system that don't have "u_original_picking_id" set
and sets the value.

Method call in XML which is run when udes_stock is installed or updated.

Story/907

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -3,7 +3,7 @@ import inspect
 from collections import OrderedDict, defaultdict
 from lxml import etree
 
-from odoo import api, fields, models, _
+from odoo import api, fields, models, SUPERUSER_ID, _
 from odoo.exceptions import ValidationError, UserError
 
 from ..common import check_many2one_validity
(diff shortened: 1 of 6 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -2753,3 +2753,46 @@ def _get_pick_details(self):
             "submitted_value": self.origin,
         }
         return pick_details
+
+    @api.model
+    def _set_original_picking_on_historical_pickings(self):
```
PeterC10, line 2758: Suggestions welcome on a better name for this method!

addons/udes_stock/models/stock_picking.py
```diff
@@ -2753,3 +2753,46 @@ def _get_pick_details(self):
...
+            update_datetime = fields.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S.%f")
+
+            update_query = """
+            UPDATE stock_picking
+            SET
+            write_uid = {user_id},
+            write_date = '{update_datetime}',
+            u_original_picking_id =
+            CASE
+            WHEN backorder_id IS NOT null THEN backorder_id
```
peterpratt13, line 2786: What happens in the recursive hole? Pick 3 backorder of pick 2 backorder of pick 1?

addons/udes_stock/models/stock_picking.py
```diff
@@ -2753,3 +2753,46 @@ def _get_pick_details(self):
...
+        if picking_count:
+            _logger.info("Setting Original Picking value on %s pickings", picking_count)
+
+            update_datetime = fields.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S.%f")
+
+            update_query = """
+            UPDATE stock_picking
+            SET
+            write_uid = {user_id},
+            write_date = '{update_datetime}',
```
kdwyer, line 2783: `write_date = now() at time zone 'UTC'` would be consistent with how [Odoo updates write date](https://github.com/unipartdigital/odoo/blob/ee08a6f63711c950bc6c3eae61fd7f959824c0c4/odoo/models.py#L3504) (unless we specifically want the application time, instead of commit time?).

PeterC10: Originally I did it like this to make sure all pickings updated had the same datetime for the write date, but at that point pickings were being updated in batches.

With them all now being updated in one query I should be able to use this, will let you know how I get on. :)

kdwyer: `now() at ...` and similar will always be the same for all queries within a single transaction - see the send of [this section](https://www.postgresql.org/docs/14/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT)

addons/udes_stock/models/stock_picking.py
```diff
@@ -2753,3 +2753,46 @@ def _get_pick_details(self):
...
+            ELSE id
+            END
+            WHERE
+            id IN ({picking_ids})
+            """.format(
+                user_id=SUPERUSER_ID,
+                update_datetime=update_datetime,
+                picking_ids=" ,".join(str(p_id) for p_id in picking_ids),
+            )
+            self.env.cr.execute(update_query)
```
kdwyer, line 2796: We can restructure slightly to follow best practice (not using string formatting for values):

```python
update_query = """
UPDATE stock_picking
SET
write_uid = %(user_id)s,
write_date = now() at time zone 'UTC'
u_original_picking_id =
CASE
WHEN backorder_id IS NOT null THEN backorder_id
ELSE id
END
WHERE
id IN %(picking_ids)s
"""
values = {'user_id': SUPERUSER_ID, 'picking_ids': tuple(picking_ids)} 
self.env.cr.execute(update_query, values)
```

`psycopg2` will convert the tuple of ids into something `IN` can handle.

addons/udes_stock/models/stock_picking.py
PeterC10: Bad miss from me. Good spot. :)

Now using a recursive CTE so I can get the picking that is top of the tree, for those pickings that are backorders. Tested against the value set against newly created pickings on the system.

Timed at 1 minute 45 seconds for a database with roughly 800k pickings but open to suggestions on performance if any, the query can take a little while to get used to at first.

addons/udes_stock/models/stock_picking.py
PeterC10: Have removed `historical`, will leave open incase someone has a better name.

addons/udes_stock/models/stock_picking.py
PeterC10: Updated now, good spot. :)

addons/udes_stock/models/stock_picking.py
PeterC10: Yep fair point, have now updated the method to be a bit more proper. Although the query has changed somewhat due to backorder fun...!

PeterAlabaster (approved):

PeterC10:
Add a few more commits to this branch to cover the following as part of this story:

1. Actually a bug in UDES11. We don't use instant transfers in UDES11, but it is the Odoo default. We added overrides so that only planned pickings are created. But we're missing some context when duplicating an existing picking, which means it automatically gets confirmed as an instant transfer.

2. Origins no longer copied across when a picking is duplicated. Need to double check this one, but from a user perspective in the UI it means the origin is blank but they need to set it before they can save (although it does already exist in the db at this point), edit, process etc the picking.

3. Partner, Scheduled Date and Origin are now readonly, unless the picking is in draft. This has been locked down so we have better traceability of the process of a chain of pickings through the system. E.g. if we print any docs or have any previously run reports referencing the current scheduled date/origin, it would be hard to tie these up if they can be edited at any point.

PR #576, opened 2022-03-02
[FIX] udes_stock: Added readonly attr when not in draft

nickcosta85:
Added a readonly attribute for destination_location on when the state is not in draft to prevent circumventing the he checks already  in place by changing the destination-location on a check peacemeal.

Story/965

Signed-off-by: Nick Costa <nicholas.costa@unipart.io>

### DIFF
--- a/addons/udes_stock/views/stock_picking.xml
+++ b/addons/udes_stock/views/stock_picking.xml
@@ -87,7 +87,7 @@
 
             <!-- Destination location field changed to read-only for goods-in -->
             <xpath expr="//field[@name='location_dest_id']" position="attributes">
-                <attribute name="attrs">{'readonly': [('picking_type_code', '=', 'incoming')]}</attribute>
+                <attribute name="attrs">{'readonly': ['|',('picking_type_code', '=', 'incoming'),('state', 'not in', 'draft')]}</attribute>
             </xpath>
 
             <!-- origin (Source Document  field changed to required for goods-in -->


addons/udes_stock/views/stock_picking.xml
```diff
@@ -87,7 +87,7 @@
 
             <!-- Destination location field changed to read-only for goods-in -->
             <xpath expr="//field[@name='location_dest_id']" position="attributes">
-                <attribute name="attrs">{'readonly': [('picking_type_code', '=', 'incoming')]}</attribute>
+                <attribute name="attrs">{'readonly': ['|',('picking_type_code', '=', 'incoming'),('state', 'not in', 'draft')]}</attribute>
```
PeterC10, line 90: Q: Do we need to consider the destination location on the move lines? This would stop the destination location being changed on the main picking record but not the individual lines.

nickcosta85: As far as I know, it's not a problem as the individual move lines can be edited but you cannot set them to anything that is not a child of the picking destination location.
But let me know if you think I have missed something

PeterAlabaster: I don't think we need to change anything for the move lines as their domain is based on children of the pickings location_dest_id, the problem (as i understand it) is that the location_dest_id could be changed on the pick thus releiving the domain on the move lines, and if the pick is waiting/ready we still want to be able to change what location stuff is being moved to

PeterAlabaster (approved):

addons/udes_stock/views/stock_picking.xml
ws-mps: Agree not needed.

ws-mps (approved):

PR #577, opened 2022-03-03
[IMP] udes_stock: origin in stock.move not set correctly

LoKKeR96:
The tests here use an helper method to create a picking instead
of using for example create_picking() in stock.picking model.
Bebcause of this there can be differences in the tests in terms
of how a picking is cretated which can mean a different behaviour
compared to UDES is tested.

This change aligns better how a picking is created compared to UDES.

User-story: 840

### DIFF
--- a/addons/udes_stock/tests/common.py
+++ b/addons/udes_stock/tests/common.py
@@ -1,7 +1,4 @@
-# -*- coding: utf-8 -*-
-
 from odoo.tests import common
-from collections import namedtuple
 
 SECURITY_GROUPS = [
     ('inbound', 'udes_stock.group_inbound_user'),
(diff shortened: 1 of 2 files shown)


addons/udes_stock/tests/common.py
```diff
@@ -291,7 +288,7 @@ def create_picking(cls, picking_type, products_info=False,
 
         if products_info:
             for product_info in products_info:
-                product_info.update(picking=picking)
+                product_info.update(picking=picking, origin=picking.origin)
```
armandcela, line 291: Not sure if you need to pass separately origin as a parameter, you are passing the whole picking. I think on create move you could use "origin": picking.origin rather that "origin": origin if that was the reason of passing a new origin parameter

LoKKeR96: Good point, I made changes to follow your suggestion.

nickcosta85:
LGTM, adding my thumb up even though you have 3 for the updated test part :)

armandcela (approved):

nickcosta85 (approved):

PR #581, opened 2022-03-14
Story/645

kdwyer:


addons/udes_stock/views/stock_picking.xml
```diff
@@ -104,6 +104,20 @@
...
     </record>
 
+    <!-- Stock Picking Type View -->
+    <record id="view_picking_type_form_udes_stock" model="ir.ui.view">
+      <field name="name">stock.picking.type.form.udes_stock</field>
+      <field name="model">stock.picking.type</field>
+      <field name="inherit_id" ref="stock.view_picking_type_form"/>
+      <field name="arch" type="xml">
+        <xpath expr="//form/sheet/group[last()]" position="after">
+          <group string="Empty Pickings" groups="base.group_no_one">
```
kdwyer, line 114: Can anyone think of a better name for this feature?

addons/udes_stock/README.md
```diff
@@ -151,12 +152,18 @@ This is essentially a collection of products that are to be moved from one locat
 | _create_batch | Creates a batch |
 | create_picking | Create and return a picking for the given picking_type. For multiple pickings a list of lists of dicts of product_info should be passed, and pickings with the same picking_type and other kwargs are the same. The kwargs are applied to pickings, not moves. If needed, the moves can be created outside of create_pickings with _create_moves. |
 | _create_move | Create and return move(s) for the given move_values. Should be used in conjunction with _prepare_move to obtain move_values |
+| get_is_empty_pickings | Find empty pickings that will be deleted|
```
peterpratt13, line 155: Think this method is `get_empty_pickings`? The description doesn't need to be about deletes, might be better to know it is a gen and uses u_is_empty not a check on the mls/moves?

kdwyer: Changed, please review.

peterpratt13 (approved):

addons/udes_stock/views/stock_picking.xml
PeterC10: Maybe just `Pickings`? We're putting this in the `udes_stock` module so everything here is pretty core and generic. It's likely we'll have other config later on to put in that doesn't nicely fall into other areas.

PeterC10 (approved):

PR #582, opened 2022-03-22
[IMP] udes_stock: Port some configuration fields for workflows

LoKKeR96:
User-story: 1168


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -2,9 +2,43 @@
...
+    u_over_receive = fields.Boolean(
+        string="Over Receive",
+        default=True,
+        help="If True, allow additional items not in the ASN, or over the expected quantity, "
+        "to be added at goods in.",
+    )
+    # TODO(ale): misleading, as this also depends on `u_user_scans` and
+    # and `u_target_storage_format`; consider using an enumeration
+    u_scan_parent_package_end = fields.Boolean(
+        string="Scan parent package at the end",
```
jrbirch99, line 38: Should capitalise the string: https://capitalizemytitle.com/

adamrpatrick (approved):

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -2,9 +2,43 @@
...
+]
+
+
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
+    u_user_scans = fields.Selection(
+        [("pallet", "Pallets"), ("package", "Packages"), ("product", "Products")],
+        string="What the User Scans",
+        help="What the user scans when asked to " "scan something from pickings of this type",
```
PeterC10, line 19: Don't need the extra quotes in the middle, this was probably a black formatting thing in UDES 11 originally.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -2,9 +2,43 @@
...
+        default=False,
+        help="If True, allow less items than the expected quantity in a move line.",
+    )
+    u_over_receive = fields.Boolean(
+        string="Over Receive",
+        default=True,
+        help="If True, allow additional items not in the ASN, or over the expected quantity, "
+        "to be added at goods in.",
+    )
+    # TODO(ale): misleading, as this also depends on `u_user_scans` and
```
PeterC10, line 35: We can remove this comment.

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -6,8 +6,37 @@ class StockWarehouse(models.Model):
...
+    def _domain_u_damaged_location_id(self):
+        """
+        Domain for locations outside Stock
+        """
+        Location = self.env["stock.location"]
+        stock_children_locations = Location.search(
+            [("id", "child_of", self.env.ref("stock.stock_location_stock").id)]
+        ).mapped("id")
+
+        return [("id", "not in", stock_children_locations)]
```
PeterC10, line 18: Suggestion:

`return ["!", ("id", "child_of", self.env.ref("stock.stock_location_stock").id)]`

This should do the same thing, but one query instead of two.

addons/udes_stock/models/stock_picking_type.py
PeterC10: image

LoKKeR96: I didn't realize this website actually checks which letters to capitalise instead of capitalising all initial letters. Thanks for reminding, I will start using it now 😆

addons/udes_stock/models/stock_warehouse.py
LoKKeR96: Happy to change but the story for this commit does not include unit tests so changing code might be a risk at this moment in time. Unless people agree it's fine.

PeterC10: The test cases for the story says unit tests, so it would be within scope to add a unit test to check this if we wanted.

We should have plenty of stock and non-stock locations in our `udes_stock` tests to check.

LoKKeR96: Sorry, you're right!

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -1,10 +1,41 @@
...
+    )
+    u_under_receive = fields.Boolean(
+        string="Under Receive",
+        default=False,
+        help="If True, allow less items than the expected quantity in a move line.",
+    )
+    u_over_receive = fields.Boolean(
+        string="Over Receive",
+        default=True,
+        help="If True, allow additional items not in the ASN, or over the expected quantity, "
```
PeterC10, line 31: Suggestion: The help string from v11 is referencing ASN and Goods In, but this setting applies for other types too.

I suggest just:

`If True, allow more items than the expected quantity in a move line.`

addons/udes_stock/models/stock_warehouse.py
LoKKeR96: As per discussions, unit tests are not required and your suggestion has been implemented and tested manually.

PeterC10 (approved):

PR #583, opened 2022-03-22
[ADD] udes_stock: Add in "scanned by" functionality

caleb-shelton:
Add tracking data of who did the picking and the done datetime.
Add unit tests covering this.

User-story: 1083

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -1,12 +1,32 @@
-from odoo import models, fields, api, _
 from collections import defaultdict
+from datetime import datetime
+import logging
+
+from odoo import models, fields, api, _
 from odoo.tools.float_utils import float_compare, float_round
 from odoo.osv import expression
 from odoo.exceptions import ValidationError
 
+_logger = logging.getLogger(__name__)
+
 
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
+
+    u_done_by_id = fields.Many2one(
+        "res.users",
+        "Scanned by",
+        help="ID of the user to complete the pack op.",
+        index=True,
+        copy=False,
+    )
+
+    u_done_datetime = fields.Datetime(
+        "Completion datetime",
+        help="Date and time the operation was completed.",
+        index=True,
+        copy=False,
+    )
 
     u_picking_type_id = fields.Many2one(
         "stock.picking.type",
@@ -15,6 +35,30 @@ class StockMoveLine(models.Model):
         store=True,
         readonly=True,
     )
+
+    @staticmethod
+    def _now_for_tracking_data():
+        """Can be overwritten to change the tracking resolution"""
+        return datetime.now()
+
+    @api.model
+    def _add_user_tracking_data(self, vals):
+        """Inject/overwrite user tracking data into vals"""
+        vals["u_done_by_id"] = self.env.uid
+        vals["u_done_datetime"] = self._now_for_tracking_data()
+        return vals
+
+    @api.model
+    def create(self, vals):
+        """Extend to track the requester user"""
+        vals = self._add_user_tracking_data(vals)
+        return super().create(vals)
+
+    @api.model
+    def write(self, vals):
+        """Extend to track the requester user"""
+        vals = self._add_user_tracking_data(vals)
+        return super().write(vals)
 
     def _log_message(self, record, move, template, vals):
         """
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -38,7 +38,7 @@ class StockMoveLine(models.Model):
 
     @staticmethod
     def _now_for_tracking_data():
-        """Can be overwritten to change the tracking resolution"""
+        """Override to change the tracking resolution"""
         return datetime.now()
 
     @api.model
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -15,14 +15,14 @@ class StockMoveLine(models.Model):
 
     u_done_by_id = fields.Many2one(
         "res.users",
-        "Scanned by",
-        help="ID of the user to complete the pack op.",
+        "Scanned By",
+        help="User who completed the product move",
         index=True,
         copy=False,
     )
 
     u_done_datetime = fields.Datetime(
-        "Completion datetime",
+        "Completion Datetime",
         help="Date and time the operation was completed.",
         index=True,
         copy=False,
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -44,8 +44,9 @@ class StockMoveLine(models.Model):
     @api.model
     def _add_user_tracking_data(self, vals):
         """Inject/overwrite user tracking data into vals"""
-        vals["u_done_by_id"] = self.env.uid
-        vals["u_done_datetime"] = self._now_for_tracking_data()
+        if vals.get("qty_done", 0) > 0:
+            vals["u_done_by_id"] = self.env.uid
+            vals["u_done_datetime"] = self._now_for_tracking_data()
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -1,13 +1,33 @@
...
 
+_logger = logging.getLogger(__name__)
+
 
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
+    u_done_by_id = fields.Many2one(
+        "res.users",
+        "Scanned by",
```
PeterC10, line 18: `Scanned By`

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1,13 +1,33 @@
...
+        help="ID of the user to complete the pack op.",
```
PeterC10, line 19: Suggestion: `User who completed the product move.`

addons/udes_stock/tests/test_scanned_by.py
```diff
@@ -0,0 +1,66 @@
...
+        cls.create_quant(cls.banana.id, cls.test_stock_location_02.id, 20)
+        picking = cls.create_picking(cls.picking_type_internal)
+        move = cls.create_single_move(cls.apple, 10, picking)
+        move = cls.create_single_move(cls.banana, 20, picking)
+        picking.action_confirm()
+        picking.action_assign()
+
+        cls.picking = picking.with_env(test_user_env)
+
+    # TODO: Does this need enabling even with real_uid removed now?
```
PeterC10, line 39: Yes this is still needed, we still want to be sure that `sudo` calls do not change who completed the operation.

addons/udes_stock/models/stock_move_line.py
caleb-shelton: I have https://capitalizemytitle.com/ book marked as well! :cry:

addons/udes_stock/models/stock_move_line.py
```diff
@@ -16,6 +36,30 @@ class StockMoveLine(models.Model):
         readonly=True,
     )
 
+    @staticmethod
+    def _now_for_tracking_data():
+        """Can be overwritten to change the tracking resolution"""
```
kdwyer, line 41: I'd make this "Override to change ..."

addons/udes_stock/tests/test_scanned_by.py
```diff
@@ -0,0 +1,64 @@
...
+        cls.create_quant(cls.banana.id, cls.test_stock_location_02.id, 20)
+        picking = cls.create_picking(cls.picking_type_internal)
+        move = cls.create_single_move(cls.apple, 10, picking)
+        move = cls.create_single_move(cls.banana, 20, picking)
+        picking.action_confirm()
+        picking.action_assign()
+
+        cls.picking = picking.with_env(test_user_env)
+
+    def test_records_real_id_on_move_line_update(self):
```
kdwyer, line 39: I'd name the method `test_records_original_user_id_on_move_line_update`, "real" doesn't seem to mean anything here.
In the docstring 
- u_done_by -> u_done_by_id?
- required -> used?
- and be reported on -> can be reported on?

addons/udes_stock/tests/test_scanned_by.py
```diff
@@ -0,0 +1,64 @@
...
+        """The original user should be in u_done_by so that when sudo is required for
+        permission access the correct user and be reported on
+        """
+        sudo_picking = self.picking.sudo()  # Switch to admin to get around any access rights
+        move_line = sudo_picking.move_line_ids[0]
+        move_line.write({"qty_done": move_line.product_uom_qty})
+        self.assertEqual(move_line.u_done_by_id.id, self.test_user.id)
+
+    def test_multiple_users_move_line_update(self):
```
kdwyer, line 48: I think this method name could be a little more descriptive; maybe something like:
`test_records_correct_user_if_different_user_completes_picking`

(I'm not convinced about "correct_user" either - ideally we want to convey that it's the user who completes the picking.  Perhaps you can come up with something better)

addons/udes_stock/tests/test_scanned_by.py
```diff
@@ -0,0 +1,64 @@
...
+        """Check that when two users act on the same picking that they are recorded seperately"""
```
kdwyer, line 49: > Check that ... they are recorded separately

We seem to be testing that the "second" user is recorded in `u_done_by_id` - the docstring should accurately reflect what we are testing.

addons/udes_stock/tests/test_scanned_by.py
caleb-shelton: I think 'correct' works.

kdwyer: mmm, but the reader needs to understand what "correct" means in this context.  I'd rather find something more explicit, if we could.

addons/udes_stock/tests/test_scanned_by.py
kdwyer: Apologies, an even better name would be

`test_records_original_user_id_on_move_line_update_with_sudo`

caleb-shelton: fixed

kdwyer (approved):

addons/udes_stock/tests/common.py
```diff
@@ -391,6 +391,23 @@ def create_move(cls, pickings, products_info, **kwargs):
         moves = Picking._create_move(move_values)
         return moves
 
+    @classmethod
+    def create_single_move(cls, product, qty, picking, **kwargs):
```
peterpratt13, line 395: Q: Why do we do this when create_move exists?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -16,6 +36,30 @@ class StockMoveLine(models.Model):
...
+    def _add_user_tracking_data(self, vals):
+        """Inject/overwrite user tracking data into vals"""
+        vals["u_done_by_id"] = self.env.uid
+        vals["u_done_datetime"] = self._now_for_tracking_data()
+        return vals
+
+    @api.model
+    def create(self, vals):
+        """Extend to track the requester user"""
+        vals = self._add_user_tracking_data(vals)
```
peterpratt13, line 54: Does this mean the u_done_by on the move line is set on creation? When the qty_done is zero

addons/udes_stock/tests/common.py
peterpratt13: Can we not just use create_move, which uses the functionality emplyed by the code outside of tests?

addons/udes_stock/models/stock_move_line.py
caleb-shelton: Does this have a potential to break something?

addons/udes_stock/tests/common.py
caleb-shelton: fixed

addons/udes_stock/models/stock_move_line.py
peterpratt13: qty_done > 0 i believe

addons/udes_stock/tests/common.py
```diff
@@ -391,6 +391,23 @@ def create_move(cls, pickings, products_info, **kwargs):
...
+            "name": product.name,
+            "product_uom": product.uom_id.id,
+            "product_uom_qty": qty,
+            "location_id": picking.location_id.id,
+            "location_dest_id": picking.location_dest_id.id,
+            "picking_id": picking.id,
+            "priority": picking.priority,
+        }
+        vals.update(kwargs)
+        return Move.create(vals)
```
peterpratt13, line 409: Can delete this now

addons/udes_stock/models/stock_move_line.py
caleb-shelton: fixed

peterpratt13 (approved):

PR #585, opened 2022-03-25
Story #1384 See who is assigned to a batch from the picking screens

jrbirch99:
[IMP] udes_stock: Add batch user_id to the picking list view.

- Add the user responsible for the batch, to which the picking belongs, to the picking list view.

Story/1384

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -105,6 +105,13 @@ class StockPicking(models.Model):
     # override batch_id to be copied
     batch_id = fields.Many2one("stock.picking.batch", copy=True)
 
+    u_batch_user_id = fields.Many2one(
+        string="Batch User",
+        related="batch_id.user_id",
+        store=False,
+        help="User responsible for the batch",
+    )
+
     u_quantity_done = fields.Float(
         "Quantity done",
         compute="_compute_picking_quantities",
--- a/addons/udes_stock/tests/test_picking.py
+++ b/addons/udes_stock/tests/test_picking.py
@@ -606,3 +606,40 @@ class TestPickingLocked(common.BaseUDES):
         self.assertTrue(
             self.picking.is_locked, "Picking should have been locked after its moves are  confirmed"
         )
+
+class TestBatchUserName(common.BaseUDES):
+    """Test Batch User takes value expected and changes when expected"""
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestBatchUserName, cls).setUpClass()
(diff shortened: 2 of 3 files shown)


addons/udes_stock/tests/test_picking.py
```diff
@@ -606,3 +606,40 @@ def test_confirming_a_move_locks_the_picking(self):
         self.assertTrue(
             self.picking.is_locked, "Picking should have been locked after its moves are  confirmed"
         )
+
+class TestBatchUserName(common.BaseUDES):
+    """Test that unlocked pickings are locked when confirmed or completed"""
```
LoKKeR96, line 611: This is not relevant right?

addons/udes_stock/models/stock_picking.py
```diff
@@ -105,6 +105,14 @@ class StockPicking(models.Model):
...
     batch_id = fields.Many2one("stock.picking.batch", copy=True)
 
+    u_batch_user_id = fields.Many2one(
+        "res.users",
+        "Batch User",
+        compute="_compute_batch_user",
+        store=False,
+        help="User responsible for the batch",
+    )
+
```
adamrpatrick, line 115: We can make use of the `related` parameter to avoid having to compute the field e.g. 
```
u_batch_user_id = fields.Many2one(related="batch_id.user_id", string=BLAH ETC)
```

addons/udes_stock/views/stock_picking.xml
```diff
@@ -193,6 +193,10 @@
                 <field name="u_location_category_id"/>
             </xpath>
 
+            <xpath expr="//field[@name='batch_id']" position="after">
+                <field name="u_batch_user_id"/>
```
adamrpatrick, line 197: Do Product want this adding to the form view as well as the tree view?

addons/udes_stock/models/stock_picking.py
```diff
@@ -307,6 +314,7 @@ def _compute_is_locked(self):
         """Derive is_locked from the picking's state."""
         self.is_locked = self.state != "draft"
 
+
```
adamrpatrick, line 317: Tiniest of nits: There's a random extra new line here

addons/udes_stock/views/stock_picking.xml
jrbirch99: Spoke to Nuno, he said he doesn't want it.

addons/udes_stock/models/stock_picking.py
jrbirch99: fixed

adamrpatrick (approved):

PR #586, opened 2022-03-25
Story #1384 See who is assigned to a batch from the picking screens

jrbirch99:
[IMP] udes_stock: Add batch user_id to the picking list view.

- Add the user responsible for the batch, to which the picking belongs, to the picking list view.

Story/1384

### DIFF
--- a/addons/udes_stock/tests/test_stock_picking.py
+++ b/addons/udes_stock/tests/test_stock_picking.py
@@ -835,3 +835,43 @@ class TestStockPickingUnlinkingEmpties(common.BaseUDES):
         Picking.unlink_empty()
 
         self.assertTrue(pick.exists())
+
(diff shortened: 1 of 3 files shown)


addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -835,3 +835,43 @@ def test_does_not_unlink_empty_pickings_if_picking_type_not_configured(self):
         Picking.unlink_empty()
 
         self.assertTrue(pick.exists())
+
+class TestBatchUserName(common.BaseUDES):
+    """Test that unlocked pickings are locked when confirmed or completed"""
```
LoKKeR96, line 840: This is not relevant, right?

LoKKeR96 (approved):

adamrpatrick (approved):

PR #587, opened 2022-03-29
[IMP] udes_stock: Adding support for finding packages

LoKKeR96:
User-story: 1169

The find_move_lines() function in stock.quant.package is not checking
for move lines that use the package as a destination package. This
change ensures this check is done.

The get_or_create() fucntion in the mixin stock model does not allow
for returning an empty result without raising an issue. Adding here
the support for such use case.

### DIFF
--- a/addons/udes_stock/models/mixin_stock_model.py
+++ b/addons/udes_stock/models/mixin_stock_model.py
@@ -30,7 +30,7 @@ class MixinStockModel(models.AbstractModel):
 
         return domain
 
-    def get_or_create(self, identifier, create=False, aux_domain=None):
+    def get_or_create(self, identifier, create=False, aux_domain=None, return_emtpy=False):
         """Gets an object of the model from the identifier. In case that no results
             are found, creates a new object of the model depending on the create
             parameter and the MSM_CREATE setting.
@@ -40,6 +40,10 @@ class MixinStockModel(models.AbstractModel):
         :kwargs:
             - create: Boolean
                 If true, and MSM_CREATE is true, a new object is created if needed
+            - aux_domain: list
+                An additional domain to add to the search
+            - return_emtpy: Boolean
+                Allow empty/False results to be returned without raising an exception
         :returns:
             Object of the model queried
         """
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/mixin_stock_model.py
```diff
@@ -40,6 +40,10 @@ def get_or_create(self, identifier, create=False, aux_domain=None):
         :kwargs:
             - create: Boolean
                 If true, and MSM_CREATE is true, a new object is created if needed
+            - aux_domain: list
+                An additional domain to add to the search
+            - no_results: Boolean
```
armandcela, line 45: Nit: Believe the name no_results can be changed as it is not understand easily that when no_results is False to raise error, how about smth like no_results_exception, you may need to change the logic in if condition and where is called.
if no_results_exception:
    raise ValidationError...

If you want by default to raise validation error may change the default to True

LoKKeR96: I will ask Miquel, I worked this out with him. We took the logic from another function that does the same thing. I think that he will reply that we want to keep the logic of this function this way most probably. Changing the name is probably not an issue. Will update you here later.

addons/udes_stock/models/stock_quant_package.py
```diff
@@ -88,7 +88,11 @@ def find_move_lines(self, aux_domain=None):
 
         if aux_domain is None:
             aux_domain = [("state", "not in", ["done", "cancel"])]
-        domain = [("package_id", "in", self.ids)] + aux_domain
+        domain = aux_domain + [
+            "|",
+            ("result_package_id", "in", self.ids),
```
jrbirch99, line 93: Q: Is `result_package_id` a stock.quant.package field? I can only see it being defined as a move_line field in the codebase.

LoKKeR96: It is a move_line field as we are searching for move lines that use this package which have multiple links to the package model hence the 2 fields.

peterpratt13:
Can we update the README for that change as well please - it needs updating on that method to say it raises an error anyway. :)

addons/udes_stock/models/stock_quant_package.py
jrbirch99: oh yeah, thought it was searching packages for some reason

adamrpatrick (approved):

jrbirch99 (approved):

PR #589, opened 2022-03-30
[FIX] udes_stock: Stop 'reserve stock' error when product has no barcode

caleb-shelton:
The error was caused by searching for products by their barcode, when barcode
is not a required field. The fix involves searching for products via their name
instead.

User-story: 903


PeterAlabaster (approved):

jrbirch99 (approved):

armandcela:
I think story description and DO should be changed default_code is Internal Reference not barcode

armandcela (approved):

PR #592, opened 2022-04-05
[IMP] udes_simple_location_blocking: Add new module for blocking locations

LoKKeR96:
[IMP] udes_simple_location_blocking: Add new module for blocking locations

User-story: 1277

Add a new module from UDES11 to block locations from being used by pickings,
stock moves, stock move lines and stock inventory adjustments.
Also avoid stock quantities that are in blocked location from being used.

Additionally, this module is now a dependency of udes_stock and udes_suggest_location.

### DIFF
--- a/addons/udes_simple_location_blocking/models/stock_location.py
+++ b/addons/udes_simple_location_blocking/models/stock_location.py
@@ -1,0 +1,80 @@
+from odoo import api, models, fields, _
+from odoo.exceptions import ValidationError
+
+
+class StockLocation(models.Model):
+    _inherit = "stock.location"
+
+    u_blocked = fields.Boolean(
+        string="Is Blocked?",
+        help=_("Check this box to prevent stock picks from this location"),
+        index=True,
+    )
+
+    u_blocked_reason = fields.Char(string="Reason for Block:", required=False)
+
+    def check_blocked(self, prefix="", extra=[]):
+        """
+        Checks if any of the locations in self are blocked and if blocked raise a ValidationError.
+
+        :param prefix: string
+        :param prefix: list: names of Boolean fields to use to filter blocked locations
+        """
+        if not isinstance(prefix, str):
+            raise ValidationError(_("Prefix parameter for check_blocked should be string"))
+        blocked_locations = self.filtered(lambda x: x.u_blocked)
+        if not isinstance(extra, list):
+            raise ValidationError(_("Extra parameter for check_blocked should be list"))
+        for field in extra:
+            blocked_locations = self.filtered(lambda x: getattr(x, field))
+        if blocked_locations:
+            raise ValidationError(
+                _("%s %s Please speak to a team leader to resolve the issue.")
+                % (prefix, "".join(blocked_locations._prepare_blocked_msg()))
+            )
+
+    def _prepare_blocked_msg(self):
+        """
+        Prepares a message for the locations depending on if it is blocked or not.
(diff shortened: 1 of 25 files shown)


adamrpatrick (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -177,7 +177,7 @@ def get_empty_locations(self, limit=None, sort=True):
         :returns: Move lines of picking
         """
         locations = self._get_child_dest_locations(
-            aux_domain=[("barcode", "!=", False), ("quant_ids", "=", False)], limit=limit
+            [("u_blocked", "=", False), ("barcode", "!=", False), ("quant_ids", "=", False)], limit=limit
```
adamrpatrick, line 180: Actually, can we add a function to get the domain to use in this search. Then we can extend it in udes_simple_blocking to add the `u_blocked` parameter and remove the need for udes_stock to depend upon udes_simple_blocking. It fits more nicely to have udes_simple_blocking depend upon udes_stock instead.

adamrpatrick (changes requested):

addons/udes_stock/models/stock_picking.py
LoKKeR96: Wouldn't that create a circular dependency?

adamrpatrick: No because udes_stock wouldn't depend upon blocked locations. Blocked locations would just override the domain.

LoKKeR96: Ah, sorry, I see what you mean. At the moment it's udes_stock that depends on simple_location_blocking but you want to move the functionality hence the dependency. Makes sense. Will do.

addons/udes_simple_location_blocking/models/stock_location.py
```diff
@@ -0,0 +1,75 @@
...
+                msg.append(_("Location %s is blocked %s.") % (location.name, reason))
+            else:
+                msg.append(_("Location %s is not blocked.") % location.name)
+        msg = " ".join(msg)
+        return msg
+
+    @api.onchange("u_blocked")
+    def onchange_u_blocked(self):
+        """Empty blocked reason when locations are unblocked"""
+        if not self.u_blocked:
```
peterpratt13, line 50: I always forget - is self always singleton here? Or should we loop?

LoKKeR96: I looked at the api.py but it doesn't specify. I would assume since this is called when the record is modified on the GUI that it returns a single record? I don't think you can modify a field for multiple records at once in the GUI.

LoKKeR96: I checked with Armand and he also thinks that it's always a singleton but he didn't say for sure. I cannot find a definitive answer.

peterpratt13: haven't looked, do we have an option to unblock multiple locations at once? With the select, and unblock or is it just going into each location (or parent loc) and unblocking it?

LoKKeR96: There is a field on the form for each location. I cannot see an additional button to unblock multiple locations at once.

peterpratt13 (approved):

adamrpatrick (approved):

PR #594, opened 2022-04-06
[IMP] udes_stock: Port utils.py from UDES11 and add tests

LoKKeR96:
User-story: 1179


addons/udes_stock/tests/test_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        label = "quantity"
+        value = "5"
+        input_dict = {"label": label, "value": value}
+
+        result = utils.format_dict_for_display_list_component(input_dict)
+        test_result = ["**label**: " + label, "**value**: " + str(value)]
+        self.assertEqual(result, test_result)
+
+        label = "quantity"
+        value = "5"
```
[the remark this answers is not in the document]
LoKKeR96: If happy can you resolve? @caleb-shelton :)

addons/udes_stock/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        "id": location.id,
+        # "blocked": {"is_blocked": location.u_blocked, "reason": reason},
+        # TODO: enable when u_storage_format is implemented
+        # "storage_format": location.u_storage_format,
+    }
+    loc.update(extras)
+    return loc
+
+
+def product_to_dict(product, **extras):
```
armandcela, line 27: I think this method can be removed and used _get_info instead which is defined in udes_get_info module.

addons/udes_stock/utils.py
```diff
@@ -0,0 +1,117 @@
+from urllib.parse import urljoin
+from odoo import _
+
+BASE_PRODUCT_IMAGE_URL = "/web/image/product.product/%i"
+
+
+def location_to_dict(location, **extras):
```
armandcela, line 7: I think this method can be removed and used _get_info instead which is defined in udes_get_info module.

addons/udes_stock/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        for product, product_lines in package_lines.groupby("product_id", order):
+            contents.append(
+                package_product_quantity_label(
+                    package, product, int(sum(product_lines.mapped("product_uom_qty")))
+                )
+            )
+    return contents
+
+
+def batch_to_dict(batch):
```
armandcela, line 62: I think this method can be removed and used _get_info instead which is defined in udes_get_info module.

addons/udes_stock/utils.py
LoKKeR96: You're right, although in UDES11 this function is used by workflows with this specific dictionary structure. I am happy to remove it which means when workflows are ported they will need to be adapted or change it to use __get_info to expand.

addons/udes_stock/utils.py
LoKKeR96: Yes I think it would be best if it's removed or adapted, the only thing I'm worried about is the field "`"name": pick.batch_id.u_original_name or pick.batch_id.name` which __get_info wouldn't exactly replicate the same way. Happy to remove or expand to use __get_info.
This function is used by workflows.

addons/udes_stock/utils.py
LoKKeR96: This one can be completely removed as it uses the same field names as the product model. Will remove.

addons/udes_stock/utils.py
armandcela: Can be put into extras i think and on location just add batch_id or in _get_info_field_names.
Would be good to use get_info module for this kind of dict converts

addons/udes_stock/utils.py
armandcela: I think yes, is better to use it in that way because thats why we moved get_info and have extras on that so we can call by different variables from different places.

addons/udes_stock/utils.py
LoKKeR96: As per discussion with Miquel, this will be removed and when workflows are ported a better way will be implemented to get this info. Most surely using __get_info() and not having a utils function just for this.

addons/udes_stock/utils.py
LoKKeR96: As per discussion with Miquel, this will be removed and when workflows are ported a better way will be implemented to get this info. Most surely using __get_info() and not having a utils function just for this.

addons/udes_stock/utils.py
LoKKeR96: As per discussion with Miquel, this will be removed and when workflows are ported a better way will be implemented to get this info. Most surely using __get_info() and not having a utils function just for this.

armandcela (approved):

PR #596, opened 2022-04-11
Story/1215

peterpratt13:


addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -71,6 +71,30 @@ def _assert_picking_related_pickings_match_expected_values(self, pickings, expec
...
 class TestStockPicking(TestStockPickingCommon):
+    def test_picking_naming_convention(self):
+        """
+        Test that backorders get created with a -001 suffix.
+        
+        Create a picking and backorder twice to test it holds for chained
+        pickings.
+        """
+        self.create_quant(self.fig.id, self.test_stock_location_02.id, 50)
+        pick = self.Picking.create_picking(
```
peterpratt13, line 82: self.create_picking

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -71,6 +71,30 @@ def _assert_picking_related_pickings_match_expected_values(self, pickings, expec
...
+        )
+        pick.move_lines[0].quantity_done = 10
+        pick._action_done()
+        bk_picking_1 = pick.u_created_backorder_ids
+        bk_picking_1.move_lines[0].quantity_done = 10
+        bk_picking_1._action_done()
+        bk_picking_2 = bk_picking_1.u_created_backorder_ids
+
+        # Check naming convention
+        self.assertEqual(bk_picking_1.name, pick.name + "-001")
```
peterpratt13, line 95: check that the original name does not have `-`

addons/udes_stock/models/common.py
```diff
@@ -0,0 +1,46 @@
...
+        seq_ids = IrSequence.search([('code', '=', code), ('company_id', 'in', [force_company, False])], order='company_id')
+        if not seq_ids:
+            return False
+        ir_sequence = seq_ids[0]
+    else:
+        ir_sequence = sequence
+    
+    # Name pattern for continuation object.
+    # Is three digits enough? Model configurable?
+    name_pattern = r"({}\d+)-(\d{{3}})".format(re.escape(ir_sequence.prefix))
```
kdwyer, line 36: For regex pernickityness, I'd use `[0-9]` for the dgits, or set `re.ASCII` in the call to `re.match`.

addons/udes_stock/models/common.py
```diff
@@ -0,0 +1,46 @@
...
+
+    match = re.match(name_pattern, obj.name)
+    if match:
+        root = match.group(1)
+        new_sequence = int(match.group(2)) + 1
+    else:
+        # This must be the original object.
+        root = obj.name
+        new_sequence = 1
+    return "{}-{:0>3}".format(root, new_sequence)
```
kdwyer, line 46: We can use an f-string here?

addons/udes_stock/models/stock_picking.py
```diff
@@ -89,14 +90,28 @@ class StockPicking(models.Model):
         help="Pickings that are unused after refactoring are empty and ready to be deleted",
     )
 
-    # User responsible for Batch 
+    # User responsible for Batch
     u_batch_user_id = fields.Many2one(
```
kdwyer, line 94: Stripping trailing whitespace :+1:

addons/udes_stock/models/stock_picking.py
```diff
@@ -89,14 +90,28 @@ class StockPicking(models.Model):
...
+        Override the method in core stock to customise the name generation.
+        
+        Backorders have a naming pattern of adding `-001` to it,
+        so they are more visible to users. 
+        """
+        if vals.get("backorder_id"):
+            ir_sequence = picking_type.sequence_id
+            # Specify sequence as it is picking type specific
+            return get_next_name(self, "stock.picking", sequence=ir_sequence)
+        else:
```
kdwyer, line 112: We don't need this `else`, we can dedent the following line.

kdwyer (approved):
I've left a few comments, but they're all nits.  Feel free to address to resolve.

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -98,6 +100,54 @@ def test_picking_naming_convention(self):
         self.assertEqual(bk_picking_1.name, pick.name + "-001")
         self.assertEqual(bk_picking_2.name, pick.name + "-002")
 
+    @unittest.skip("Takse too long")
```
kdwyer, line 103: s/Takse/Takes

addons/udes_stock/models/common.py
kdwyer: There's two \ds :smiley:

addons/udes_stock/tests/test_stock_move.py
```diff
@@ -32,7 +30,7 @@ def _get_expected_move_line_values(self, move, qty, **kwargs):
         expected_move_values.update(kwargs)
         return expected_move_values
 
-    def test01_split_out_move_lines_raise_error(self):
+    def test_split_out_move_lines_raise_error(self):
```
PeterAlabaster, line 33: <3

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -71,6 +74,80 @@ def _assert_picking_related_pickings_match_expected_values(self, pickings, expec
...
+        bk_picking_1.move_lines[0].quantity_done = 10
+        bk_picking_1._action_done()
+        bk_picking_2 = bk_picking_1.u_created_backorder_ids
+
+        # Check naming convention
+        self.assertNotRegex("-", pick.name)
+        self.assertEqual(bk_picking_1.name, pick.name + "-001")
+        self.assertEqual(bk_picking_2.name, pick.name + "-002")
+
+    @unittest.skip("Takse too long")
```
PeterAlabaster, line 103: nit: typo: `Takes`

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -71,6 +74,80 @@ def _assert_picking_related_pickings_match_expected_values(self, pickings, expec
...
+    def test_picking_naming_convention_exceeds_limit(self):
```
PeterAlabaster, line 104: Can we replicate the same test but with fewer picks by chaanging the padding to `00` instead of `000`?

PeterAlabaster:
Nice, thumbing but going to leave my last comment open for a TL to check on :)

PeterAlabaster (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -235,43 +249,6 @@ def get_move_lines(self, done=None):
             return self.move_line_ids.get_lines_incomplete()
         return mls
 
-    def _backorder_move_lines(self, mls=None):
```
ws-mps, line 238: I don't think we should be deleting this function, only reversing which move lines we move into the backorder.

peterpratt13: I thought it might be nice just to rely on action_done, so we have one behaviour.  I don't see how we would want to split one set of move lines into one picking then another, because i guess refactoring should do that?

addons/udes_stock/models/common.py
```diff
@@ -0,0 +1,58 @@
+import re
```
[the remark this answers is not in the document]
PeterC10: I think we're okay with this in 'udes_stock' for now at least. It's something we can keep under review going forwards though.

It may be in future we'd want a similar system in place for other areas, but without that being thought about yet I think it's better to keep this where it as and if we need to we can move it later.

peterpratt13:
Update:
* _backorder_move_lines is kept. 
* It now moves all incomplete working into a new move whilst trying to preserve the move lines if possible and the quants reserved etc.
* Calls a new method that is the reverse split_out_move_lines, delibaretly separate as the cases become less clear when merged.
* tests added

caleb-shelton (commented):
{typo comment}

PR #597, opened 2022-04-11
[IMP] udes_stock_picking_batch: Implement Batch States

LoKKeR96:
User-story: 1172

Implement additional batch states from UDES11 and relative
functions.

### DIFF
--- a/addons/udes_stock_picking_batch/__manifest__.py
+++ b/addons/udes_stock_picking_batch/__manifest__.py
@@ -10,8 +10,11 @@
     "depends": [
         "base",
         "stock",
+        "stock_picking_batch",
     ],
     "external_dependencies": {},
-    "data": [],
+    "data": [
+        "views/stock_picking_batch.xml"
+    ],
     "demo": [],
 }
--- a/addons/udes_stock_picking_batch/models/stock_picking.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking.py
@@ -1,5 +1,53 @@
-from odoo import models
+from odoo import models, api
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
+
+    @api.constrains("batch_id")
+    def _trigger_batch_confirm_and_remove(self):
+        """
+       Batch may require new pickings to be auto confirmed or removed
+       Note: this function is very fragile, any change might brake its functionality
+        """
+        origin_batches = self.env.context.get("orig_batches")
+        batches = self.batch_id
+        if origin_batches:
+            batches += origin_batches
+        batches._assign_picks()
+        batches._remove_unready_picks()
+        batches._compute_state()
+
+    @api.constrains("state")
+    def _trigger_batch_state_recompute(self):
+        """
+        Changes to picking state cause batch state recompute, may also cause
+        unready pickings to be removed from the batch
+        Note: this function is very fragile, any change might brake its functionality
+        """
+        origin_batches = self.env.context.get("orig_batches")
+        batches = self.batch_id
+        if origin_batches:
+            batches += origin_batches
+        batches._remove_unready_picks()
+        batches._compute_state()
+
+    def action_assign(self):
+        """
+        Recompute batch state. In theory this is not necessary
+        but the constraint on state has not proven to work correctly.
+        """
+        super(StockPicking, self).action_assign()
+        self.batch_id._compute_state()
+
+    def write(self, vals):
+        """
+        If writing picking, check if previous batch is now complete.
+        This will be used to trigger recompute of the batch state
+        we can't relate on the state after as batch_id might be
+        removed during write.
+        """
+
+        if self.batch_id:
+            self = self.with_context(orig_batches=self.batch_id)
+        return super(StockPicking, self).write(vals)
--- a/addons/udes_stock_picking_batch/models/stock_picking_batch.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking_batch.py
@@ -1,5 +1,202 @@
-from odoo import models
+from odoo import models, fields, _, api
+from odoo.exceptions import UserError, ValidationError
+import logging
+
+_logger = logging.getLogger(__name__)
 
 
 class StockPickingBatch(models.Model):
     _inherit = "stock.picking.batch"
+
+    user_id = fields.Many2one(
+        "res.users",
+        states={
+            "draft": [("readonly", False)],
+            "waiting": [("readonly", False)],
+            "ready": [("readonly", False)],
+            "in_progress": [("readonly", False)],
+        },
+    )
+    # Note: state field has been found to recompute itself everytime it was accessed even with store=True
+    # lock_batch_state has been put in places to ensure correct behaviour.
+    state = fields.Selection(
+        selection=[
+            ("draft", "Draft"),
+            ("waiting", "Waiting"),
+            ("ready", "Ready"),
+            ("in_progress", "Running"),
+            ("done", "Done"),
+            ("cancel", "Cancelled"),
+        ],
+        compute="_compute_state",
+        store=True,
+    )
+    picking_ids = fields.One2many(
+        "stock.picking",
+        "batch_id",
+        states={
+            "draft": [("readonly", False)],
+            "waiting": [("readonly", False)],
+            "ready": [("readonly", False)],
+            "in_progress": [("readonly", False)],
+        },
+    )
+
+    @api.constrains("user_id")
+    def _compute_state(self):
+        """ Compute the state of a batch
+            waiting     : At least some picks are not ready
+            ready       : All picks are in ready state (assigned)
+            in_progress : All picks are ready and a user has been assigned
+            done        : All picks are complete (in state done or cancel)
+
+            the other two states are draft and cancel are manual
+            to transitions from or to the respective state.
+        """
+        if self.env.context.get("lock_batch_state"):
+            # State is locked so don't do anything
+            return
+
+        for batch in self:
+            if batch.state in ["draft", "cancel"]:
+                # Can not do anything with them don't bother trying
+                continue
+
+            if batch.picking_ids:
+
+                ready_picks = batch.ready_picks()
+                done_picks = batch.done_picks()
+                unready_picks = batch.unready_picks()
+
+                # Figure out state
+                if ready_picks and not unready_picks:
+                    if batch.user_id:
+                        batch.state = "in_progress"
+                    else:
+                        batch.state = "ready"
+
+                if ready_picks and unready_picks:
+                    if batch.user_id:
+                        batch.state = "in_progress"
+                    else:
+                        batch.state = "waiting"
+
+                if done_picks and not ready_picks and not unready_picks:
+                    batch.state = "done"
+            else:
+                batch.state = "done"
+
+    def done_picks(self):
+        """ Return done picks from picks or self.picking_ids """
+        picks = self.picking_ids
+        return picks.filtered(lambda pick: pick.state in ["done", "cancel"])
+
+    def ready_picks(self):
+        """ Return ready picks from picks or self.picking_ids """
+        picks = self.picking_ids
+        return picks.filtered(lambda pick: pick.state == "assigned")
+
+    def unready_picks(self):
+        """ Return unready picks from picks or self.picking_ids """
+        picks = self.picking_ids
+        return picks.filtered(lambda pick: pick.state in ["draft", "waiting", "confirmed"])
+
+    def mark_as_todo(self):
+        """Mark as to-do will change the state from draft to waiting.
+
+        This is done without calling action assign.
+        """
+        _logger.info("User %r has marked %r as todo.", self.env.uid, self)
+        not_draft = self.filtered(lambda b: b.state != "draft")
+        if not_draft:
+            raise UserError(
+                _(
+                    "The following batches can not be marked as Todo as they are not in the draft state: %s"
+                )
+                % not_draft.mapped("name")
+            )
+        self.write({"state": "waiting"})
+        self._compute_state()
+
+    def action_confirm(self):
+        """Overwrite method action_confirm
+           Move batch from draft to waiting.
+           Raise error if not in draft and rollback to draft on error in action_assign.
+        """
+        self.ensure_one()
+        if self.state != "draft":
+            raise ValidationError(
+                _("Batch (%s) is not in state draft can not perform " "action_confirm")
+                % ",".join(self.name)
+            )
+        if not self.picking_ids:
+            raise UserError(_("You have to set some pickings to batch."))
+        self._check_company()
+
+        pickings_todo = self.picking_ids
+        # Set batch to waiting
+        self.write({"state": "waiting"})
+
+        try:
+            p = pickings_todo.with_context(lock_batch_state=True).action_assign()
+            self._compute_state()
+            return p
+        except:
+            # Return all to draft
+            self.write({"state": "draft"})
+            raise
+
+    @api.constrains("picking_ids")
+    def _assign_picks(self):
+        """If configured, attempt to assign all the relevant pickings in self"""
+        if self.env.context.get("lock_batch_state"):
+            # State is locked so don't do anything
+            return
+
+        # Get active batches with pickings, apply lock_batch_state here as we do not
+        # want to recompute the state just yet as we want to mark pickings assigned first.
+        batches = self.with_context(lock_batch_state=True).filtered(
+            lambda b: (
+                b.state in ["waiting", "in_progress"]
+                and b.picking_ids
+                and any(b.mapped("picking_type_id.u_auto_assign_batch_pick"))
+            )
+        )
+
+        for batch in batches:
+            picks_to_assign = batch.picking_ids.filtered(
+                lambda p: p.state == "confirmed"
+                and p.picking_type_id.u_auto_assign_batch_pick
+                and p.move_lines.filtered(
+                    lambda move: move.state not in ("draft", "cancel", "done")
+                )
+            )
+            if picks_to_assign:
+                picks_to_assign.with_context(lock_batch_state=True).action_assign()
+                batch._compute_state()
+
+    def _remove_unready_picks(self):
+        """ Remove unready picks from running batches in self, if configured """
+        if self.env.context.get("lock_batch_state"):
+            # State is locked so don't do anything
+            return
+
+        # Get unready picks in running batches
+        unready_picks = (
+            self.with_context(lock_batch_state=True)
+            .filtered(lambda b: b.state in ["waiting", "in_progress"])
+            .unready_picks()
+        )
+
+        if not unready_picks:
+            # Nothing to do
+            return
+
+        # Remove unready pick, if configured.
+        unready_picks.filtered(lambda p: p.picking_type_id.u_remove_unready_batch)
+
+        # TODO - implement when u_reserved_pallet is implemented
+        # unready_picks.write(
+        #     {"batch_id": False, "u_reserved_pallet": False}
+        # )
+        unready_picks.write({"batch_id": False})
--- a/addons/udes_stock_picking_batch/models/stock_picking_type.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking_type.py
@@ -1,5 +1,17 @@
-from odoo import models
+from odoo import models, fields
 
 
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
+
+    u_auto_assign_batch_pick = fields.Boolean(
+        string="Auto Assign Running Batch Picks",
+        help="Reserve automatically stock to picks when added to a running batch",
+        default=False,
+    )
+
+    u_remove_unready_batch = fields.Boolean(
+        string="Auto Remove Running Batch Unready Picks",
+        help="Remove automatically unready picks from a running batch when batch assigned or pick state changed",
+        default=True,
+    )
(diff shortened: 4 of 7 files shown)


addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+            <field name="arch" type="xml">
+                <xpath expr="//filter[@name='draft']" position="after">
+                    <filter name="ready" string="Ready"
+                            domain="[('state','=', 'ready')]"
+                            help="Batch Pickings Ready"/>
+                    <filter name="waiting" string="Waiting"
+                            domain="[('state','=', 'waiting')]"
+                            help="Batch Pickings Waiting"/>
+                </xpath>
+                <xpath expr="//filter[@name='in_progress']" position="replace">
```
armandcela, line 40: Can change what you need to change i think with position="attributes". I think replacing should be used as last resort.

example:

`<xpath expr="//filter[@name='in_progress']" position="attributes">`
    `<attribute name="help">Batch Transfers not finished</attribute>`
    `<attribute name="domain">[]</attribute>`
`</xpath>`

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+                </xpath>
+                <xpath expr="//filter[@name='in_progress']" position="after">
+                    <filter name="cancel" string="Cancelled"
+                            domain="[('state','=', 'cancel')]"
+                            help="Batch Pickings Cancelled"/>
+                </xpath>
+            </field>
+        </record>
+
+        <!-- Stock picking batch action -->
```
armandcela, line 52: Nit: Is an act_window what you are overriding, i think the comment should be improved

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
LoKKeR96: What I wanted to change was the string, for some reason the code is still the same so I need to change the name. Happy to try your way anyway.

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+                </xpath>
+                <xpath expr="//button[@name='action_cancel']" position="replace">
+                    <button name="action_cancel" string="Cancel" type="object" states="ready,waiting,in_progress"/>
+                </xpath>
+            </field>
+        </record>
+
+        <!-- Stock picking batch filter -->
+        <record id="stock_picking_batch_filter" model="ir.ui.view">
+            <field name="name">udes_stock.stock.picking.batch.filter</field>
```
armandcela, line 27: Nit: udes_stock i think was a copy paste from udes_stock module. Not to get confused i think would be better to change the name.
```suggestion
            <field name="name">stock.picking.batch.filter.inherit</field>
```
I would use like this as the name of the module is very long

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,93 @@
...
+
+    def mark_as_todo(self):
+        """Changes state from draft to waiting.
+
+        This is done without calling action assign.
+        """
+        _logger.info("User %r has marked %r as todo.", self.env.uid, self)
+        not_draft = self.filtered(lambda b: b.state != "draft")
+        if not_draft:
+            raise UserError(_('Only draft batches may be marked as "todo": %s') % not_draft.ids)
```
armandcela, line 91:
```suggestion
            raise UserError(_('Only draft batches may be marked as "todo": %s') % not_draft.mapped("name"))
```
As ids may be confusing for users to understand which batches are not in draft

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
PeterC10: Personally I think it is good practice to include the module name (although better at the end imo, so it is appended to the original name)

Helps with debugging when you have an issue with views.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
PeterC10: The message could be clearer too, i.e.

`The following batches can not be marked as ToDo as they are not in the draft state:`

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,93 @@
...
+            ("ready", "Ready"),
+            ("in_progress", "Running"),
+            ("done", "Done"),
+            ("cancel", "Cancelled"),
+        ],
+        compute="_compute_state",
+        store=True,
+    )
+
+    @api.constrains("user_id")
```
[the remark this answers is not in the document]
PeterC10: Yes, seems like a mistake...

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
PeterC10: Yes, better to change an individual attribute than replace the whole thing.

The original code has string as `In Progress`, but you've overridden it to still say `In Progress`. Am I missing something?

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+            <field name="model">stock.picking.batch</field>
+            <field name="inherit_id"
+                   ref="stock_picking_batch.stock_picking_batch_filter"/>
+            <field name="arch" type="xml">
+                <xpath expr="//filter[@name='draft']" position="after">
```
PeterC10, line 32: Can we add cancelled too? I know it is an original state but it doesn't have a filter from what I remember when we looked over it earlier.

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+                    <filter name="ready" string="Ready"
+                            domain="[('state','=', 'ready')]"
+                            help="Batch Pickings Ready"/>
```
PeterC10, line 35: Suggestion: `Ready Batch Pickings` - reads a bit more clearer

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
PeterC10: You've already done this, ignore, sorry. :)

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+                    <filter name="waiting" string="Waiting"
+                            domain="[('state','=', 'waiting')]"
+                            help="Batch Pickings Waiting"/>
+                </xpath>
+                <xpath expr="//filter[@name='in_progress']" position="replace">
+                    <filter name="in_progress" string="In Progress" domain="[('state', '=', 'in_progress')]"
+                            help="Batch Transfers not finished"/>
+                </xpath>
+                <xpath expr="//filter[@name='in_progress']" position="after">
+                    <filter name="cancel" string="Cancelled"
```
PeterC10, line 45: Q: Does this make the filter show at the bottom of all the states?

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+        </record>
+
+         <!-- 'Mark as To-do' in action dropdown -->
+        <record id="action_mark_activities_done" model="ir.actions.server">
+            <field name="name">Mark as Todo</field>
+            <field name="model_id" ref="model_stock_picking_batch"/>
+            <field name="binding_model_id" ref="udes_stock_picking_batch.model_stock_picking_batch"/>
+            <field name="state">code</field>
+            <field name="code">
+                for rec in records:
```
jrbirch99, line 69: Q: Have you tested that this code runs and doesn't error? When I wrote a cron job in 11 in this format (mutli-line) it didn't run for me.

LoKKeR96: I tried and it does work.

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
PeterC10: Could we put Cancelled underneath done?

It's not an expected state in the "happy path" flow if that makes ense. :)

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
LoKKeR96: @PeterC10 I wanted to change it to Running instead of "In Progress" since the display string for this state is "Running". Would that be a good idea?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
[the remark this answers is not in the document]
LoKKeR96: Could it be that who wrote it originally was also trying to change the state of the batch based on whether a user_id is assigned? I wouldn't say that's correct either but just a thought.

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
LoKKeR96: If I change it that way then I need to do the same of all the others?

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -0,0 +1,75 @@
...
+    <data>
+
+        <!-- Stock picking batch form -->
+        <record id="stock_picking_batch_form" model="ir.ui.view">
+            <field name="name">udes_stock.stock.picking.batch.form</field>
+            <field name="model">stock.picking.batch</field>
+            <field name="inherit_id"
+                   ref="stock_picking_batch.stock_picking_batch_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='state']" position="replace">
```
PeterC10, line 12: Similar to this thread: https://github.com/unipartdigital/udes-open/pull/597/files#r847543615

Just replace the `statusbar_visible` attribute, same goes for the `action_cancel` button too (just change `states` in this one)

addons/udes_stock_picking_batch/models/stock_picking_batch.py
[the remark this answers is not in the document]
PeterC10: I really don't know, has been like that in v11 since 2018!

For v14 I would suggest changing this to:

```
@api.depends("user_id", "picking_ids", "picking_ids.batch_id")
```

That ensures the state will be recomputed in the following cases:

* The user is changed
* The pickings are changed on the batch
* The batch is changed via the picking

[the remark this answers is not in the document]
armandcela: @PeterC10 Resolving the thread as is addressed to your suggestion

armandcela (approved):

PeterC10 (approved):

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+       Batch may require new pickings to be auto confirmed or removed
+       Note: this function is very fragile, any change might brake its functionality
+        """
+        orig_batches = self.env.context.get("orig_batches")
+        if orig_batches:
+            for batch in orig_batches:
+                batch._assign_picks()
+                batch._remove_unready_picks()
+                batch._compute_state()
+        for batch in self.mapped("batch_id"):
```
armandcela, line 19: Nit: 
```suggestion
        for batch in self.batch_id:
```

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+        Changes to picking state cause batch state recompute, may also cause
+        unready pickings to be removed from the batch
+        Note: this function is very fragile, any change might brake its functionality
+        """
+        orig_batches = self.env.context.get("orig_batches")
+        if orig_batches:
+            for batch in orig_batches:
+                batch._remove_unready_picks()
+                batch._compute_state()
+        for batch in self.mapped("batch_id"):
```
armandcela, line 36: Nit: 
```suggestion
        for batch in self.batch_id:
```

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+
+    @api.constrains("batch_id")
+    def _trigger_batch_confirm_and_remove(self):
+        """
+       Batch may require new pickings to be auto confirmed or removed
+       Note: this function is very fragile, any change might brake its functionality
+        """
+        orig_batches = self.env.context.get("orig_batches")
+        if orig_batches:
+            for batch in orig_batches:
```
armandcela, line 15: Question: Do we need to enter in loop , as from what i see methods inside handles the multi records.

So probably could do direct :  `batches._assign_picks() `etc and just need to find batches first
`batches = orig_batches or self.batch_id `

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+        """
+        super(StockPicking, self).action_assign()
+        self.batch_id._compute_state()
+
+    def write(self, vals):
+        """If writing picking, check if previous batch is now complete"""
+        # This will be used to trigger recompute of the batch state
+        # we can't relate on the state after as batch_id might be
+        # removed in the write
+        batches = self.batch_id
```
armandcela, line 53: Nit: Docstring and comment can be merged into docstring into one whole docstring i think

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+        context_vals = {"orig_batches": batches} if batches else {}
```
armandcela, line 54:
```
if batches:
    self= self.with_context(orig_batches=batches)
return super(StockPicking, self).write(vals)
```

I think looks cleaner and no need to use kwargs for a single context variable

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: I don't think so but very simple changes in this function were causing unexpected behaviour so I avoided touching it too much. I am going to push a simplified version of it that seems to work now that I have implemented the recompute state on the action_assign on picks.

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,5 +1,55 @@
...
+            batch._compute_state()
+
+    @api.constrains("state")
+    def _trigger_batch_state_recompute(self):
+        """
+        Changes to picking state cause batch state recompute, may also cause
+        unready pickings to be removed from the batch
+        Note: this function is very fragile, any change might brake its functionality
+        """
+        orig_batches = self.env.context.get("orig_batches")
```
armandcela, line 31: Can find the batches first and later call the methods , will remove in this way duplicated code

`batches = orig_batches or self.batch_id`
batches._remove_unready_picks() if we can call with multi

addons/udes_stock_picking_batch/__manifest__.py
```diff
@@ -10,8 +10,11 @@
     "depends": [
         "base",
         "stock",
+        "stock_picking_batch"
```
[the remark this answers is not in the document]
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,202 @@
...
+        "batch_id",
+        states={
+            "draft": [("readonly", False)],
+            "waiting": [("readonly", False)],
+            "ready": [("readonly", False)],
+            "in_progress": [("readonly", False)],
+        },
+    )
+
+    @api.constrains("user_id")
```
armandcela, line 45: Did we agree on this api.depends or constraint

LoKKeR96: As it is now, constraint.

adamrpatrick (approved):

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,202 @@
...
+        if self.state != "draft":
+            raise ValidationError(
+                _("Batch (%s) is not in state draft can not perform " "action_confirm")
+                % ",".join(self.name)
+            )
+        if not self.picking_ids:
+            raise UserError(_("You have to set some pickings to batch."))
+        self._check_company()
+
+        pickings_todo = self.mapped("picking_ids")
```
armandcela, line 136:
```suggestion
        pickings_todo = self.picking_ids
```

armandcela (approved):

addons/udes_stock_picking_batch/models/stock_picking.py
LoKKeR96: @armandcela This should be addressed now

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
```diff
@@ -1,7 +1,231 @@
...
+
+        # complete pick1 and check state 'in_progress', forcibly compute_state
+        self.complete_pick(self.picking01, call_done=False)
+        self.picking01._action_done()
+        self.assertEqual(self.batch01.state, 'in_progress')
+
+        # complete pick2 and check state 'done', forcibly compute_state
+        self.complete_pick(self.picking02, call_done=False)
+        self.picking02._action_done()
+        self.assertEqual(self.batch01.state, 'done')
```
PeterC10, line 231: Missing newline and also using single quotes, can you black format all these files please?

addons/udes_stock_picking_batch/models/stock_picking_type.py
```diff
@@ -1,5 +1,17 @@
-from odoo import models
+from odoo import models, fields
 
 
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
+
+    u_auto_assign_batch_pick = fields.Boolean(
+        string="Auto assign picks when added to a running batch",
```
PeterC10, line 8: Let's cut this down, it won't display nicely on the form view.

Maybe something like: `Auto Assign Running Batch Picks`?

We have the help text for a more detailed explanation.

addons/udes_stock_picking_batch/models/stock_picking_type.py
```diff
@@ -1,5 +1,17 @@
...
+        help="Auto assign picks when added to a running batch",
+        default=False,
+    )
+
+    u_remove_unready_batch = fields.Boolean(
+        string="Auto remove unready picks from a running batch",
```
PeterC10, line 14: Similar to above, maybe `Auto Remove Unready Batch Picks`?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,202 @@
...
+from odoo.exceptions import UserError, ValidationError
+import logging
+
+_logger = logging.getLogger(__name__)
 
 
 class StockPickingBatch(models.Model):
     _inherit = "stock.picking.batch"
+
+    user_id = fields.Many2one(
```
PeterC10, line 11: Q: How does Odoo display this field on the UI? `User`?

I'm sure this has come up before but I can't remember if Odoo is smart enough to remove `id` from the string. :)

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,202 @@
...
+            ("waiting", "Waiting"),
+            ("ready", "Ready"),
+            ("in_progress", "Running"),
+            ("done", "Done"),
+            ("cancel", "Cancelled"),
+        ],
+        compute="_compute_state",
+        store=True,
+    )
+    picking_ids = fields.One2many(
```
PeterC10, line 34: Might need a string here too.

addons/udes_stock_picking_batch/models/stock_picking_type.py
LoKKeR96: Done

addons/udes_stock_picking_batch/models/stock_picking_type.py
LoKKeR96: Done

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
LoKKeR96: Done

PR #599, opened 2022-04-12
Story/x1174

armandcela:


addons/udes_stock/models/stock_move_line.py
```diff
@@ -230,3 +230,14 @@ def _split(self, uom_qty=None):
             )
             res = new_ml
         return res
+
+    def get_lines_todo(self):
```
ws-mps, line 234: Is this the same as `get_lines_incomplete`?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,581 @@
...
+            skipped_mls = all_available_mls.filtered(
+                lambda ml: ml.product_id.id in skipped_product_ids
+            )
+        elif skipped_move_line_ids:
+            skipped_mls = all_available_mls.filtered(lambda ml: ml.id in skipped_move_line_ids)
+        available_mls = all_available_mls - skipped_mls
+
+        num_tasks_picked = len(available_mls.filtered(lambda ml: ml.qty_done == ml.product_qty))
+
+        todo_mls = available_mls.get_lines_todo().sort_by_location_product()
```
ws-mps, line 304: Can we call a method that calls sort_by_location_product() instead of directly calling it? this will make the code easier to extend. Same for line 319.

addons/udes_stock/models/stock_move_line.py
armandcela: Yes replacing :)

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -1,5 +1,581 @@
...
+
+        return lambda ml: tuple(chain(*[part(ml) for part in parts]))
+
+    def get_available_move_lines(self):
+        """ Get all the move lines from available pickings
+        """
+        self.ensure_one()
+        available_pickings = self.picking_ids.filtered(lambda p: p.state == "assigned")
+
+        return available_pickings.move_line_ids
```
ws-mps, line 259: This would probably automatically sort the move lines by the ordering of the pickings (creating a JOIN behind the scenes) which in some cases it is not needed, usually we have some code to sort the move lines after calling this function; which can be inefficient with big batches.
Can we add a parameter `sort = False` and use `.search(XXX, order="id")` when it is false, please?

jrbirch99:
Don't forget to add ported methods to the README :)

jrbirch99:
Please run the black formatter on the files too

addons/udes_stock/models/stock_move_line.py
```diff
@@ -230,3 +233,15 @@ def _split(self, uom_qty=None):
             )
             res = new_ml
         return res
+
+    def get_fast_move_lines(self, domain, aux_domain=None):
```
PeterAlabaster, line 237: Not sure about the name of this function, is it just a faster (less cpu cycles) way to get stock.move.line recordsets with a search domain?

Could be worth putting on basemodel as `def search_orderby_id()` as it seems like a pretty generic function

addons/udes_stock/models/stock_picking.py
```diff
@@ -228,13 +228,25 @@ def get_move_lines(self, done=None):
...
+        mls = self.get_fast_move_lines()
         if done:
-            return self.move_line_ids.get_lines_done()
+            return mls.get_lines_done()
         elif done == False:
-            return self.move_line_ids.get_lines_incomplete()
+            return mls.get_lines_incomplete()
         return mls
 
+    def get_fast_move_lines(self, aux_domain=None):
```
PeterAlabaster, line 238: Why aren't we using aux_domain on the `stock.move.line` function `get_fast_move_lines()` to pass in `[("picking_id", "in", self.ids)]` and

Can't we just do
```python
StockMoveLine = self.env["stock.move.line"]
mls = StockMoveLine.get_fast_move_lines(aux_domain=[("picking_id", "in", self.ids)])
```

addons/udes_stock/models/stock_move_line.py
armandcela: It is Miquel's suggestion name, putting into a method etc :)

addons/udes_stock/models/stock_location.py
```diff
@@ -6,6 +6,7 @@ class StockLocation(models.Model):
     _name = "stock.location"
     # Add messages and abstract model to locations
     _inherit = ["stock.location", "mail.thread", "mixin.stock.model"]
+    MSM_STR_DOMAIN = ("name", "barcode")
```
caleb-shelton, line 9: What is MSM?

addons/udes_stock/models/stock_picking.py
armandcela: Needed the method in case that we need to override it , was Miquel's suggestion

addons/udes_stock/models/stock_location.py
armandcela: Dont know what sands for tbh, but is used to find the record from identifier.

kdwyer: It stands for Mixin Stock Model, common code defined [here](https://github.com/unipartdigital/udes-open/blob/c585241ca2f751986e41941e174249a4a631c8d8/addons/udes_stock/models/mixin_stock_model.py#L10).

addons/udes_stock/models/stock_move_line.py
PeterC10: Could definitely be useful on basemodel, but I can raise an issue for that.

This name is a bit misleading though, as we have product speeds etc.

Tbh just `get_move_lines` would do... with a comment in the doc-string about performance increase as Peter A suggested.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -200,3 +222,701 @@ def _remove_unready_picks(self):
...
+            limit=limit,
+        )
+
+        # Get tasks for movelines that have been skipped (if allowed)
+        incomplete_mls = (
+            skipped_mls.filtered(lambda ml: ml.picking_id.picking_type_id.u_return_to_skipped)
+            .get_lines_incomplete()
+            .next_task_sort()
+        )
+        # Determine the remaining limit (Need to do a distninct check as False != 0)
```
kdwyer, line 515: s/distninct/distinct

addons/udes_stock/models/stock_picking.py
PeterAlabaster: can we keep the method and replace with the above then?

addons/udes_stock/models/stock_location.py
PeterAlabaster: mainstream media :laughing:

addons/udes_stock/models/stock_move_line.py
adamrpatrick: As Pete suggests, let's call this function `get_move_lines` with an option `order` parameter. Everywhere we currently call `get_fast_move_lines` passing in "id" as the order. This should make it very clear as to what the function's doing.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -200,3 +222,701 @@ def _remove_unready_picks(self):
...
+        which will be incremented or set to '01'.
+        The batch will not be marked as ephemeral.
+        In case no pickings exist, return None.
+        """
+        self.ensure_one()
+
+        if not pickings:
+            return None
+
+        new_name = get_next_name(self, "picking.batch")
```
kdwyer, line 913: Have we ported `get_next_name`?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
PeterC10: Resolving, as this is done and we have another thread on how to name the method. :)

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
```diff
@@ -235,3 +236,418 @@ def test_computing_partial_assignment(self):
         self.complete_pick(self.picking02, call_done=False)
         self.picking02._action_done()
         self.assertEqual(self.batch01.state, "done")
+
+
+class TestBatchMultiDropOff(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestBatchMultiDropOff, cls).setUpClass()
```
kdwyer, line 244: We can just do `super().setUpClass()` I think.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
armandcela: Copying from Lorenzo's MR, as these were part of his story. I think have missed it at first

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
armandcela: Yes we can do, just copied from udes11 though

addons/udes_stock/models/stock_move_line.py
armandcela: get_move_lines exists like a method with optional done , which filters done moves or incomplete moves

addons/udes_stock_picking_batch/models/stock_move_line.py
```diff
@@ -0,0 +1,128 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _determine_priority_skipped_moveline(
+        self, skipped_product_ids=None, skipped_move_line_ids=None
+    ):
+        """Returns a priority move line based on the first move line found
```
LoKKeR96, line 19: Nit: can we rephrase a bit and give some context about how skipped move lines are ordered? The docstring is phrased a bit weirdly.

addons/udes_stock_picking_batch/models/stock_move_line.py
```diff
@@ -0,0 +1,128 @@
...
+        else:
+            return False
+
+    def next_task_sort(self):
+        """Creating a method so we can override it by sorting another combination.
+        Would be easier and meaningful than to extend sort_by_key as we may need to change the
+        behaviour of next_task_sort result rather than changing the behaviour of sort_by_key"""
+        return self.sort_by_key()
+
+    def _prepare_task_info(self):
```
LoKKeR96, line 49: Do we have a definition somewhere of what a task is? I know this might be out of scope but if we don't have any explanations of what a task is, can we add one? Maybe in the readme? I can see lots of functions that talk about a task but no explanations anywhere of what a task is and how it is relevant to a batch.

addons/udes_stock_picking_batch/models/stock_move_line.py
LoKKeR96: @armandcela Will resolve and suggest we look into this later as part of another story as you are time bounded right now.

addons/udes_stock_picking_batch/models/stock_move_line.py
LoKKeR96: @armandcela Will resolve and suggest we look into this later as part of another story as you are time bounded right now.

adamrpatrick (approved):

LoKKeR96 (approved):
Considering everything, as long as unit tests pass and QA and sign off go well, I am happy tagging as reviewed.

PR #603, opened 2022-04-19
Story #1173 Mobile UI: Batch user managing

jrbirch99:
The code associated with users and batches has been ported from udes_11. Units tests have also been ported and created, where needs be. 

This story relies on methods that are in the process of being ported in other stories: story/1172, story/1174. To use some of these methods this branch has been rebased with story/1174 and is compared to that branch so only the changes made in story/1173 are present in the PR. Once story/1172 has been merged to master this branch will need to be rebased with master. Not all the unit tests will pass until story/1174 has the methods from story/1172.

### DIFF
--- a/addons/udes_stock_picking_batch/models/stock_picking.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking.py
@@ -3,3 +3,34 @@ from odoo import models
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
+
+    def get_move_lines_done(self):
+        """ Return the recordset of move lines done. """
+        return self.move_line_ids.filtered(lambda o: o.qty_done > 0)
+
+    def is_valid_location_dest_id(self, location=None, location_ref=None):
+        """Whether the specified location or location reference is a valid
+        putaway location for the picking. Expects a singleton instance.
+
+        Parameters
+        ----------
+        location : Location obj
+            Location record
+        location_ref: char
+            Location identifier can be the ID, name or the barcode
+
+        Returns a boolean indicating the validity check outcome.
+        """
+        Location = self.env["stock.location"]
+        self.ensure_one()
+
+        if not location and not location_ref:
+            raise ValidationError("Must specify a location or ref")
+
+        dest_location = location or Location.get_location(location_ref)
+        if not dest_location:
+            raise ValidationError(_("The specified location is unknown."))
+
+        valid_locations = self._get_child_dest_locations([("id", "=", dest_location.id)])
+
+        return valid_locations.exists()
--- a/addons/udes_stock_picking_batch/models/stock_picking.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking.py
@@ -147,3 +147,56 @@ class StockPicking(models.Model):
             ("state", "=", "assigned"),
         ]
         return self.search_count(domain) >= 1
+
+    def _backorder_movelines(self, mls=None, dest_picking=None):
+        """Creates a backorder picking from self (expects a singleton)
+        and a subset of stock.move.lines are then moved into it.
+        The moves from which the move lines have been transferred
+        will be split if needed.
+        Completed move lines will be selected unless mls is set.
+        Optionally receive a picking into where the mls will be moved.
+        """
+        Move = self.env["stock.move"]
+
+        # Based on back order creation in stock_move._action_done
+        self.ensure_one()
+        # Select completed move lines when not passed
+        if mls is None:
+            mls = self.move_line_ids.filtered(lambda x: x.qty_done > 0)
+        # Check the intersection of mls and move lines in picking to ensure
+        # there are relevant move lines
+        if not (mls & self.move_line_ids):
+            raise ValidationError(
+                _("There are no move lines within picking %s to backorder" % self.name)
+            )
+        new_moves = Move.browse()
+        for current_move in mls.move_id:
+            current_mls = mls.filtered(lambda x: x.move_id == current_move)
+            new_moves |= current_move.split_out_move_lines(current_mls)
+
+        if dest_picking is None:
+            # Create picking for selected move lines
+            dest_picking = self._create_new_picking(name="/", backorder_id=self.id)
+
+        new_moves.write({"picking_id": dest_picking.id})
+        new_moves.mapped("move_line_ids").write({"picking_id": dest_picking.id})
+
+        return dest_picking
+
+    @api.model
+    def _prepare_new_picking_info(self, **kwargs):
+        """Copy the picking information from picking onto self"""
+        update_args = kwargs
+        if not kwargs.get("move_lines"):
+            update_args["move_lines"] = []
+        if not kwargs.get("move_line_ids"):
+            update_args["move_line_ids"] = []
+        return update_args
+
+    def _create_new_picking(self, **kwargs):
+        """Copy the picking information from picking onto self
+        Separating into a method in order to be easier to inherit with super"""
+        return self.copy(self._prepare_new_picking_info(**kwargs))
+
+
+
--- a/addons/udes_stock_picking_batch/models/stock_picking.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking.py
@@ -29,12 +29,8 @@ class StockPicking(models.Model):
         """Compute location category from move lines"""
         for picking in self:
             if picking.move_line_ids:
-                categories = picking.move_line_ids.mapped(
-                    "location_id.u_location_category_id"
-                )
-                picking.u_location_category_id = (
-                    categories if len(categories) == 1 else False
-                )
+                categories = picking.move_line_ids.mapped("location_id.u_location_category_id")
+                picking.u_location_category_id = categories if len(categories) == 1 else False
 
     def get_move_lines_done(self):
         """Return the recordset of move lines done."""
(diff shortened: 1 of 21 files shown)


addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -148,3 +158,35 @@ def _priorities_has_ready_pickings(self, priorities, picking_type_id):
             ("state", "=", "assigned"),
         ]
         return self.search_count(domain) >= 1
+
+    def batch_to_user(self, user):
+        """Throws error if picking is batched to another user
+        or if user already has a batch
+        creates batch and adds picking to it otherwise
+        """
```
LoKKeR96, line 166: Nit: can we rephrase it a bit? It's a little hard to follow. It should state clearly what the intentions are and not just what it does, then follow with the exceptions.

Also, I think the file is missing  a new line at the end? Might be wrong on this

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+                raise ValidationError(
+                    _("Found %d batches for the user, please contact " "administrator.")
+                    % len(batches)
+                )
+
+            batch = batches
+
+        return batch
+
+    def _check_user_id(self, user_id):
```
adamrpatrick, line 645: Nit: This method might be better named as `_get_user_id` as it's not really checking much

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -1,4 +1,4 @@
-from odoo import models, fields, api
+from odoo import models, fields, api, _
 from .common import PRIORITY_GROUPS
 from odoo.exceptions import ValidationError
```
LoKKeR96, line 4: I think the story asks to add a function called get_batch_for_pick? is that still required?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -391,15 +424,21 @@ def _populate_next_tasks(
...
             tasks.append(task)
             if limit and len(tasks) >= limit:
                 break
-            move_lines = move_lines.filtered(lambda ml: ml.id not in task["move_line_ids"])
+            move_lines = move_lines.filtered(
+                lambda ml: ml.id not in task["move_line_ids"]
+            )
         return tasks
 
-    def _populate_next_task(self, move_lines, task_grouping_criteria, priority_ml=False):
```
LoKKeR96, line 402: Nit: Lots of formatting but this line is 90 characters in length, the rule as far as I am aware is 100 and then it is formatted. Can you check your black formatting script? It should use 100 as the line maximum length.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+        if batches:
+            if len(batches) > 1:
+                raise ValidationError(
+                    _("Found %d batches for the user, please contact " "administrator.")
+                    % len(batches)
+                )
+
+            batch = batches
+
+        return batch
```
adamrpatrick, line 643: Nit: Some of this logic can be tidied up e.g. if the search returns no results it will be an empty recordset. Perhaps something like this:
```
batches = PickingBatch.search(BLAH)

if len(batches) > 1:
    raise ValidationError(BLAH)

return batches or None
```

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+
+        if not user_id:
+            raise ValidationError(_("Cannot determine the user."))
+
+        return user_id
+
+    def get_user_batches(self, user_id=None):
+        """Get all batches for user"""
+        if user_id is None:
+            user_id = self.env.user.id
```
adamrpatrick, line 657: This snippet is the same as `_check_user_id`.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+
+        Raise a ValidationError in case it cannot perform a search
+        or if multiple batches are found for the specified user.
+        """
+        PickingBatch = self.env["stock.picking.batch"]
+
+        user_id = self._check_user_id(user_id)
+        batches = PickingBatch.search(
+            [("user_id", "=", user_id), ("state", "=", "in_progress")]
+        )
```
adamrpatrick, line 631: This snippet is the same as `get_user_batches`.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+        if user_id is None:
+            user_id = self.env.user.id
+
+        if not user_id:
+            raise ValidationError(_("Cannot determine the user."))
+
+        return user_id
+
+    def get_user_batches(self, user_id=None):
+        """Get all batches for user"""
```
LoKKeR96, line 655: That's misleading, it is returning only running batches. Can we improve docstring please?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96, line 645: Small nit: not necessary but would be nice to add docstring explaining the intentions.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -575,3 +614,293 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+        pickings_to_remove.with_context(lock_batch_state=True).write(
+            {"batch_id": False, "u_reserved_pallet": False}
+        )
+        pickings_to_add.with_context(lock_batch_state=True).write({"batch_id": self.id})
+        self._compute_state()
+
+        return self
+
+
+def get_next_name(obj, code):
```
LoKKeR96, line 868: Could it be worth having a chat with a TL to see if this could live somewhere else? It feels wrong to have it only here but not a blocker for the review.

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -26,14 +26,18 @@ class StockPicking(models.Model):
...
         for picking in self:
             if picking.move_line_ids:
-                categories = picking.move_line_ids.mapped("location_id.u_location_category_id")
-                picking.u_location_category_id = categories if len(categories) == 1 else False
+                categories = picking.move_line_ids.mapped(
+                    "location_id.u_location_category_id"
+                )
+                picking.u_location_category_id = (
+                    categories if len(categories) == 1 else False
+                )
```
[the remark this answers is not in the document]
jrbirch99: yep will do.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
jrbirch99: Not sure I understand, as it seems to me that it checks that there is a user_id input and will return the environment user if no input is given or raise an exception if it set to false.

addons/udes_stock_picking_batch/models/stock_picking.py
jrbirch99: get_batch_for_pick lives in the workflows controller class in udes_11 in udes_workflows/controllers/common, and calls batch_to_user. I think when we port workflows this method will be ported doesn't seem right to port it as part of this story.

LoKKeR96: No problem there as long as TL and PO are happy.

addons/udes_stock_picking_batch/models/stock_picking.py
jrbirch99: I don't see any complaint in github for a newline at the end of the file. I have also check in my editor, I think it is fine :)

addons/udes_stock_picking_batch/models/stock_picking_batch.py
jrbirch99: I think as it stands in udes_11 it is only called in the picking_batch class, so I think it makes sense to keep it in that file for now. Maybe in the future it can be moved.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
jrbirch99: The black formatting script was checked and it is now set at 100 a line maximum length

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: Yeah I totally agree that it's only used here and it makes sense to live here, sometimes Miquel suggested me to move things in other places and this could be one of those times. I was only suggesting to ask a TL. I won't block the review either way so as soon as you're done and no one comments on this I will resolve.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: Can you run it again? I tried running it on the file in this commit and the output is different than what's in the last commit so it needs to be reformatted. If it doesn't change then the setting for line length isn't working.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
adamrpatrick: The way I read it is that it's returning a user_id therefore it's getting a user_id rather than checking. In reality you could argue that it's doing a bit of both.

adamrpatrick (approved):

PR #604, opened 2022-04-20
Story/52

kdwyer:


addons/udes_get_info/models/__init__.py
```diff
@@ -1,2 +1,8 @@
 from . import models
 from . import stock_location
+from . import stock_picking_type  # noqa: F401
```
LoKKeR96, line 3: what does noqa mean?

caleb-shelton: I asked the same question, it is used for: flake8 (linter) to ignore that we don't use the imports

kdwyer: https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors

addons/udes_get_info/tests/common.py
```diff
@@ -0,0 +1,136 @@
+"""Common code for testing get_info methods."""
```
LoKKeR96, line 1: Some of these functions are defined in the common.py in udes_stock. I know this module does not depend on that. I have seen other modules also copying these functions again. Could we not start doing something better? Sorry to drag it on your story but it seems like a good place to start doing that.

For example, move the common functions in the common.py in udes_stock into udes_common which is a dependency of this module and then mix the classes where needed for the other modules that need those functions. It would simplify code all around the codebase and we would be using the same functions everywhere which is a lot cleaner and reliable long term. Even in terms of porting code.

caleb-shelton (approved):

addons/udes_get_info/models/__init__.py
LoKKeR96: Thanks. I did not know we use it.

addons/udes_get_info/tests/common.py
kdwyer: In general I agree.  I'm not sure it's going to happen in this PR though :).  Perhaps @PeterC10 or @adamrpatrick will chime in on this.

PeterC10: Yep not enough scope in this story for this change.

Something we can look to handle once we get past our first milestone.

PeterC10:
Can we add `udes_get_info` to the dockerfile? I don't think the tests are running on CI atm.

PeterC10 (approved):

addons/udes_get_info/models/stock_move.py
```diff
@@ -0,0 +1,25 @@
...
+
+    _inherit = "stock.move"
+
+    # These fields will be extracted by get_info().
+    _get_info_field_names = {
+        "location_dest_id",
+        "location_id",
+        "move_line_ids",
+        "product_id",
+        "product_qty",
```
adamrpatrick, line 22: Can we add `product_uom_qty` and `product_uom` please

addons/udes_get_info/models/stock_picking_type.py
```diff
@@ -0,0 +1,24 @@
...
+    # These fields will be extracted by get_info().
+    _get_info_field_names = {
+        "code",
+        "count_picking_ready",
+        "default_location_dest_id",
+        "default_location_src_id",
+        "display_name",
+        "id",
+        "name",
+        "sequence",
```
adamrpatrick, line 23: Perhaps `company_id` and `warehouse_id` too

addons/udes_get_info/models/models.py
```diff
@@ -55,7 +55,7 @@ def _get_info(self, level=1, max_level=None, info_fields=None, extra_fields=None
             if level == max_level:
                 next_info_fields = {"info_fields": BASIC_GET_INFO_VALUES}
             else:
-                next_info_fields = {"extra_fields": extra_fields}
+                next_info_fields = {"info_fields": info_fields}
```
armandcela, line 58: If you pass info_fields on relational field i think that on next iteration it wont pick the info_fields of the relational field model

kdwyer: Possibly :)  I'm going to revert this change for now and rework the tests so that they pass instead; I'll log an issue to get this fixed properly.

kdwyer: Workaround added in [d660cc5](https://github.com/unipartdigital/udes-open/pull/604/commits/d660cc52d5c9122ee2aaec240e3675054e216a10)

addons/udes_get_info/models/stock_move.py
kdwyer: Added. please review.

addons/udes_get_info/models/stock_picking_type.py
kdwyer: Added. please review.

armandcela (approved):

PR #609, opened 2022-04-25
[IMP] udes_stock_refactoring, udes_stock_picking_type: Batch with sin…

LoKKeR96:
…gle picking type

During porting of UDES11 batch functionalities to UDES14 the batch picking_type_ids
has been ported which is not compatible with Odoo14 stock_picking_batch module as the
stock.picking.batch model already implements a single picking type per batch.

These changes ensure that the fact that only one picking type per batch is allowed
is reflected in the code and the tests.

User-story: 1175


addons/udes_stock_picking_batch/models/res_users.py
```diff
@@ -1,11 +1,12 @@
...
+from odoo import fields, models, _
+from odoo.exceptions import ValidationError
 
 
 class ResUser(models.Model):
     _inherit = "res.users"
 
     u_location_category_ids = fields.Many2many(
-        comodel_name="stock.location.category", index=True, string="Location Category"
+        comodel_name="stock.location.category", index=True, string=_("Location Category")
```
armandcela, line 9: Do we need _ on fields definition i thought we agreed that will put translate=True where needed?

PR #611, opened 2022-04-26
Story/x1788

LoKKeR96:
During porting of UDES11 batch functionalities to UDES14 the batch picking_type_ids
has been ported which is not compatible with Odoo14 stock_picking_batch module as the
stock.picking.batch model already implements a single picking type per batch.

These changes ensure that the fact that only one picking type per batch is allowed
is reflected in the code and the tests.


addons/udes_stock_picking_batch/models/res_users.py
```diff
@@ -1,4 +1,5 @@
-from odoo import fields, models
+from odoo import fields, models, _
```
armandcela, line 1: Do we need to import _ if not used

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -57,20 +55,36 @@ def action_assign(self):
...
+        If we create a batch and assign it to a picking while we are creating a picking then the picking type of
+        the batch will not be set by Odoo14 stock_picking_batch module during picking creation because in the UI
+        you cannot create a batch while you create a picking, but you can only assign a picking later to a batch.
+        Solve the issue by setting the batch picking type here.
+        """
+        res = super().create(vals)
+        if vals.get("batch_id"):
+            if not res.batch_id.picking_type_id:
+                res.batch_id.picking_type_id = res.picking_type_id
+                # Call sanity check again because we changed the allowed picking types
```
armandcela, line 73: nit:
```suggestion
                # Call sanity check again because we changed the allowed picking type
```

addons/udes_stock_picking_batch/models/stock_move_line.py
```diff
@@ -1,10 +1,6 @@
-from collections import defaultdict
-from datetime import datetime
 import logging
 
-from odoo import models, fields, api, _
-from odoo.tools.float_utils import float_compare, float_round
-from odoo.osv import expression
+from odoo import models, _
```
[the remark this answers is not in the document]
LoKKeR96: Scroll the file.

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
```diff
@@ -1046,69 +1044,62 @@ def test_get_batches_assigned_to_a_user(self):
...
-        Create two pickings, assign the pickings to the batch, put the batch in a ready state and then request that the
-        one of the pickings
+        Check that the batch picking type is assigned when a picking is created and assigned a batch
+        at the same time.
         """
-        self.create_quant(self.apple.id, self.test_received_location_01.id, 4)
         self.create_quant(self.banana.id, self.test_stock_location_01.id, 4)
 
-        batch = self.create_batch(user=self.outbound_user)
+        batch_pick = self.create_batch(user=False)
```
kdwyer, line 1054: Any reason for changing this name?  It seems sensible to name the thing that `create_batch` returns "batch".

LoKKeR96: We don't want to assign a user to the batch. create_batch() will assign cls.env.user if user parameter is None.

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -57,20 +55,36 @@ def action_assign(self):
...
         If writing picking, check if previous batch is now complete.
         This will be used to trigger recompute of the batch state
         we can't relate on the state after as batch_id might be
         removed during write.
         """
-
         if self.batch_id:
             self = self.with_context(orig_batches=self.batch_id)
-        return super(StockPicking, self).write(vals)
+        return super().write(vals)
```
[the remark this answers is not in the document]
LoKKeR96: Mmmm, I would prefer having more people weight on this. I don't feel confident going either way. I looked at Odoo's code and it uses super() so I thought it would be good to follow Odoo.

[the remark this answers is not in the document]
kdwyer: Personally I would use `super()` - the old way should only be used if there is something unusual going on.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -351,7 +325,6 @@ def add_extra_pickings(self, picking_type_id, limit=1):
             raise ValidationError(_("No more work to do."))
 
         picking_type = pickings.picking_type_id
-        picking_type.ensure_one()
```
armandcela, line 354: What not needed ensure_one(), i think still by code can be more than one picking_type

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
kdwyer: I mean changing the name from `batch` to `batch_pick`...

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -885,23 +842,21 @@ def assign_batch(self, picking_type_id, selection_criteria=None):
...
 
         Note that the transition from state 'ready' to 'in_progress'
         is handled by computation of state function.
         """
-        batch = self.env["stock.picking.batch"]
-        
-        batches = self.search([("state", "=", "ready")]).filtered(
-            lambda b: all([pt.id == picking_type_id for pt in b.picking_type_ids])
-        )
+        if not type(picking_type_id) == int:
```
armandcela, line 851:
```suggestion
        if not isinstance(picking_type_id, int):
```

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -885,23 +842,21 @@ def assign_batch(self, picking_type_id, selection_criteria=None):
...
+            raise ValidationError("picking_type_id should be an integer.")
+        batches = self.search([("state", "=", "ready"), ("picking_type_id", "=", picking_type_id)])
 
         if batches:
             batch = self._select_batch_to_assign(batches)
             batch.user_id = self.env.user
 
-        return batch
+        return batches
```
armandcela, line 859: Question: DO we need to return batches (can be more than one), or the fix should be to initialise batch = False before if batches: and return batch?

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
LoKKeR96: Oh sorry I completely misread it. No, I was copying code around and I wrote it with a more specific name because I used it somewhere else. Not any particular reason.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: This is actually a mistake of mine, I somehow thought we where reassigning batches but clearly we are not. I will fix it to return batch. Good spot! 👍🏻

addons/udes_stock_picking_batch/models/stock_picking_batch.py
armandcela: I think you are right, it was like that on udes11, ported in same way. But it doesnt seem a possibility that can be more than one picking, maybe just extra precautional was the reason of adding it first in udes11.

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -57,20 +55,36 @@ def action_assign(self):
...
+        super().action_assign()
         self.batch_id._compute_state()
 
+    @api.model
+    def create(self, vals):
+        """
+        If we create a batch and assign it to a picking while we are creating a picking then the picking type of
+        the batch will not be set by Odoo14 stock_picking_batch module during picking creation because in the UI
+        you cannot create a batch while you create a picking, but you can only assign a picking later to a batch.
+        Solve the issue by setting the batch picking type here.
```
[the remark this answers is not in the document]
LoKKeR96: Addressed.

armandcela (approved):

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
kdwyer: I think it would a little better in reverse - `pick_batch` and `putaway_batch` - I think that makes it clearer the variable refer to batches.

LoKKeR96: Addressed, sorry for force pushing (bad habit), at least it's a minor change.

kdwyer (approved):

caleb-shelton (approved):

PR #612, opened 2022-04-26
Story/x1175

LoKKeR96:
[IMP] udes_stock,udes_stock_batch_picking: Batch Exception Handling

User-story: 1175

As per title, port batch exception handling which is:

- Unpickable items functionality
- Stock Investigation Picking Type

Unpickable items function will create Stock Investigation Pickings when
an item is not pickable from a picking. This is tied with backordering.

### DIFF
--- a/addons/udes_stock/README.md
+++ b/addons/udes_stock/README.md
@@ -3,13 +3,13 @@
 ## Default settings
 
 - product.product
-    - Track visiblity of active field
+    - Track visibility of active field
 - product.template
-    - Track visiblity of active field
+    - Track visibility of active field
     - Disable name translation
     - Default type stockable
 - stock.location
-    - Track visiblity of active field
+    - Track visibility of active field
     - Disable name translation
 - data/stock_config.xml
     - Batch pickings turned on
@@ -23,9 +23,10 @@
     - Input → Received
     - Stock → Normal Locations
     - Output → Goods Out Lane / Trailers
+    - Stock Investigation -> Stock Investigation
 - data/routes.xml
     - Goods in → Putaway
-    - Pick → Goods Out → Trailer Despatch
+    - Pick → Goods Out → Trailer Dispatch
 
 ## Requires
 - udes_common:
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
 | u_assigned_user_ids     | one2many (res.users)     | Assigned users to the picking, can be more than one if picking type handles multi users    |
 
 
-| Helpers                   | Description                                                                                                                                                                                                                                                                                                                                       |
-|---------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| get_empty_locations       | Returns the recordset of locations that are child of the instance dest location and are empty.Expects a singleton instance.                                                                                                                                                                                                                       |
-| _get_child_dest_locations | Return the child locations of the instance dest location. Extra domains are added to the child locations search query, when specified. Expects a singleton instance.                                                                                                                                                                              |
-| get_move_lines            | Get move lines associated to picking, uses functions in stock_move_line.                                                                                                                                                                                                                                                                          |
-| backorder_move_lines     | Checks if a backorder is required, creates one according to _backorder_move_lines if so, else returns an empty record set; neither picking is validated. |
-| _backorder_move_lines     | Creates a backorder pick from self for all incomplete work, the original move retains with all completed work, new moves are created for the incomplete wrok.  The original pick is not validated. It should be called from within backorder_move_lines. |
-| _requires_backorder       | Checks if a backorder is required by checking if all move lines within a picking are present in mls. Cannot be consolidated with _check_backorder in Odoo core, because it does not take into account any move lines parameter.  |
-| _prepare_picking_info     | Prepare the picking_info and products_info, returning the picking_info as a list of pickings, and products_info None or as a list of lists of a dictionary of the product info. This is to allow us to create multiple pickings at once.                                                                                                          |
-| _prepare_move             | Return a list of the move details to be used later in creation of the move(s). Note: pickings and products_info should be formatted via _prepare_picking_info. Allows for multiple moves to be created at once.                                                                                                                                   | 
-| _create_batch             | Creates a batch                                                                                                                                                                                                                                                                                                                                   |
-| create_picking            | Create and return a picking for the given picking_type. For multiple pickings a list of lists of dicts of product_info should be passed, and pickings with the same picking_type and other kwargs are the same. The kwargs are applied to pickings, not moves. If needed, the moves can be created outside of create_pickings with _create_moves. |
-| _create_move              | Create and return move(s) for the given move_values. Should be used in conjunction with _prepare_move to obtain move_values                                                                                                                                                                                                                       |
-| get_empty_pickings        | Generate batches marked with u_is_empty in a recordset or picking type (if u_auto_unlink_pickings is on)                                                                                                                                                                                                                                          |
-| unlink_empty              | Deletes empty pickings generated by get_empty_pickings                                                                                                                                                                                                                                                                                            |
-| unassign_users            | Unassigning the user(s) from the pickings from self with optional skip_users                                                                                                                                                                                                                                                                      |
+| Helpers                         | Description                                                                                                                                                                                                                                                                                                                                       |
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| get_empty_locations             | Returns the recordset of locations that are child of the instance dest location and are empty.Expects a singleton instance.                                                                                                                                                                                                                       |
+| _get_child_dest_locations       | Returns the child locations of the instance dest location. Extra domains are added to the child locations search query, when specified. Expects a singleton instance.                                                                                                                                                                              |
+| get_move_lines                  | Gets move lines associated to picking, uses functions in stock_move_line.                                                                                                                                                                                                                                                                          |
+| backorder_move_lines            | Checks if a backorder is required, creates one according to _backorder_move_lines if so, else returns an empty record set; neither picking is validated.                                                                                                                                                                                          |
+| _backorder_move_lines           | Creates a backorder pick from self for all incomplete work, the original move retains with all completed work, new moves are created for the incomplete wrok.  The original pick is not validated. It should be called from within backorder_move_lines.                                                                                          |
+| _requires_backorder             | Checks if a backorder is required by checking if all move lines within a picking are present in mls. Cannot be consolidated with _check_backorder in Odoo core, because it does not take into account any move lines parameter.                                                                                                                   |
+| _prepare_picking_info           | Prepares the picking_info and products_info, returning the picking_info as a list of pickings, and products_info None or as a list of lists of a dictionary of the product info. This is to allow us to create multiple pickings at once.                                                                                                          |
+| _prepare_move                   | Returns a list of the move details to be used later in creation of the move(s). Note: pickings and products_info should be formatted via _prepare_picking_info. Allows for multiple moves to be created at once.                                                                                                                                   | 
+| _create_batch                   | Creates a batch                                                                                                                                                                                                                                                                                                                                   |
+| create_picking                  | Creates and returns a picking for the given picking_type. For multiple pickings a list of lists of dicts of product_info should be passed, and pickings with the same picking_type and other kwargs are the same. The kwargs are applied to pickings, not moves. If needed, the moves can be created outside of create_pickings with _create_moves. |
+| _create_move                    | Creates and returns move(s) for the given move_values. Should be used in conjunction with _prepare_move to obtain move_values                                                                                                                                                                                                                       |
+| get_empty_pickings              | Generate batches marked with u_is_empty in a recordset or picking type (if u_auto_unlink_pickings is on)                                                                                                                                                                                                                                          |
+| unlink_empty                    | Deletes empty pickings generated by get_empty_pickings                                                                                                                                                                                                                                                                                            |
+| unassign_users                  | Unassigning the user(s) from the pickings from self with optional skip_users                                                                                                                                                                                                                                                                      |
+| get_stock_investigation_message | Dummy method which is overridden in customer modules to provide further info in messages for stock investigations                                                                                                                                                                                                                                 |
+| raise_stock_investigation       | Raises a stock investigation picking with the provided reason, quants and                                                                                                                                                                                                                                                                         |
 
 ### Picking Type (model: stock.picking.type)
 
@@ -180,6 +183,7 @@ The type of stock.picking can is defined by this type. It can represent a goods 
 | u_over_receive            | Boolean | If the system is able to receive more than what is expected.                                                                             |
 | u_scan_parent_package_end | Boolean | If the system should ask the user to scan a parent package on drop off                                                                   |
 | u_multi_users_enabled     | Boolean | Flag to enable multi users processing same picking simultaneously.                                                                       |
+| u_enable_unpickable_items  | Boolean | Flag to indicate if the current picking type should support handling of unpickable items                                                 |
 
 | Helpers                       | Description                                          |
 |-------------------------------|------------------------------------------------------|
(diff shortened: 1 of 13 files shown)


addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -983,3 +1004,196 @@ def remove_unfinished_work(self):
...
+                    )
+                )
+
+            original_id = None
+
+            if len(picking.move_line_ids - move_lines):
+                # Create a backorder for the affected move lines if there are other move
+                # lines not relevant to unpickable move lines
+                original_id = picking.id
+                picking = picking._backorder_move_lines(picking.move_line_ids - move_lines)
```
peterpratt13, line 1144: backorder = picking._...

addons/udes_stock/models/stock_picking.py
```diff
@@ -681,3 +682,57 @@ def _unassign_users(self, skip_users=None):
         users_to_unassign.sudo().write(
             {"u_picking_id": False, "u_picking_assigned_time": False}
         )
+
+    @staticmethod
+    def get_stock_investigation_message(quants):
+        """
+        Dummy method to be overridden on a per-customer basis.
+
```
peterpratt13, line 690: nit: Can we add the args in the doc string to be consistent.

addons/udes_stock_picking_batch/tests/test_stock_picking_batch.py
```diff
@@ -1374,3 +1388,1127 @@ def _create_valid_batch(self):
...
+        """
+        picking, batch = self._create_valid_batch()
+        move_line = picking.move_line_ids[0]
+        reason = 'missing item'
+        batch.unpickable_item(package_name=move_line.package_id.name,
+                              reason=reason)
+        unpickable_picking = self.package_one.find_move_lines().picking_id
+
+        self.assertEqual(picking.state, 'confirmed',
+                         'picking was not confirmed')
```
peterpratt13, line 1406: Nit: We can probably drop the assert messages, makes it clearer. 
'stock investigation picking type not set by unpickable_item' isn't as clear as unpickable picking type does not equal internal picking type

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -877,6 +889,16 @@ def _create_batch(self, user_id, picking_type_id, picking_priorities=None, picki
         The batch will be marked as ephemeral.
         In case no pickings exist, return None.
         """
+        # TODO - This function actually searches for one picking but it talks about pickings... why?
+        #        What is this function actually supposed to do?
```
peterpratt13, line 893: Yeah it looks like single picking. I guess the only way for multiple if you pass multiple picking_id's into picking_id? Looks like it might be realted to matrix picking for some stuff when glancing at 11

addons/udes_stock/models/stock_picking.py
```diff
@@ -681,3 +682,57 @@ def _unassign_users(self, skip_users=None):
...
+        :returns: str
+        """
+        return ""
```
ws-mps, line 693: Can we add a default message like: product x quantity?

addons/udes_stock/models/stock_picking.py
```diff
@@ -681,3 +682,57 @@ def _unassign_users(self, skip_users=None):
...
+            group_id=group.id,
+            confirm=True,
+            assign=True,
+        )
+
+        picking_details = self.get_stock_investigation_message(quants.exists())
+
+        if picking_details:
+            picking_header = "Stock Investigation created with: <br>"
+            investigation_picking.message_post(body=_(picking_header + picking_details))
```
ws-mps, line 734: I would probably leave the translation of `pick_details` to `get_stock_investigation_message` and do something like:
```
picking_header = "Stock Investigation created with: <br>"
msg_body = _(picking_header) + pick_details
investigation_picking.message_post(body=msg_body)
```

addons/udes_stock_picking_batch/models/stock_picking_batch.py
ws-mps: I guess nothing prevents to have an overridden version of `search_for_pickings` where default limit value is >1. And I guess the tests were written taking into account that the default is 1 in this module. When `u_reserve_pallet_per_picking` = True, more pickings can be added using `add_extra_pickings`.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -877,6 +889,16 @@ def _create_batch(self, user_id, picking_type_id, picking_priorities=None, picki
...
+        #        It seems that it should search and add multiple pickings since we search by picking type
+        #        But actually it only assigns the first picking found. This is prone to issues if we create 2
+        #        pickings and then call this function as potentially either could be selected first. This was
```
ws-mps, line 896: Which one is selected first relies on the `_order` defined at stock.picking model, there is a note about it in `search_for_pickings`.

addons/udes_stock/models/stock_picking.py
ws-mps: Just seen that we have `if picking_details:`, so no need for this, will resolve.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -983,3 +1004,196 @@ def remove_unfinished_work(self):
...
+        original_picking_ids = {}
+        if raise_stock_investigation:
+            to_investigate = Picking.browse()
+        for picking in pickings:
+            picking.message_post(body=msg)
+
+            if picking.batch_id != self:
+                raise ValidationError(_("Move line is not part of the batch."))
+
+            if picking.state in ["cancel", "done"]:
```
ws-mps, line 1130: Better change to tuple instead of list: `("cancel", "done")`

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -983,3 +1004,196 @@ def remove_unfinished_work(self):
...
+            raise ValidationError(
+                _("Cannot find move lines todo for unpickable item in this batch.")
+            )
+
+        pickings = move_lines.picking_id
+        original_picking_ids = {}
+        if raise_stock_investigation:
+            to_investigate = Picking.browse()
+        for picking in pickings:
+            picking.message_post(body=msg)
```
ws-mps, line 1125: Do we want to do this after the validations?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -983,3 +1004,196 @@ def remove_unfinished_work(self):
...
+                raise ValidationError(
+                    _(
+                        "Cannot mark a move line as unpickable "
+                        "when it is part of a completed Picking."
+                    )
+                )
+
+            original_id = None
+
+            if len(picking.move_line_ids - move_lines):
```
ws-mps, line 1140: We could do:
```
mls_to_backorder = picking.move_line_ids - move_lines
if mls_to_backorder:
    ...
    picking = picking._backorder_move_lines(mls_to_backorder)
```

addons/udes_stock_picking_batch/__manifest__.py
```diff
@@ -16,6 +16,7 @@
         "udes_get_info",
         "udes_stock_refactoring",
         "udes_stock",
+        "package_hierarchy",
```
ws-mps, line 19: Is this a missing dependency?

LoKKeR96: I have no idea how this got here. I will double-check and remove if not needed.

addons/udes_stock/models/stock_picking.py
LoKKeR96: I'm not sure I understand? What are you suggesting here?

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: I'm not sure I follow what you are suggesting. Can you expand please?

addons/udes_stock_picking_batch/__manifest__.py
LoKKeR96: Not needed, not sure how it got there.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: It might seem logical to do that but the line here is specifically reassigning the picking variable such that the following code acts on the original picking or backorder picking depending on the case. I will try to see if I can make this a little clearer but I cannot simply rename to backorder.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
LoKKeR96: Thanks, that field has changed between Odoo 11 and Odoo14 which explains why the difference in behaviour in the tests.

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| get_empty_locations             | Returns the recordset of locations that are child of the instance dest location and are empty.Expects a singleton instance.                                                                                                                                                                                                                       |
+| _get_child_dest_locations       | Return the child locations of the instance dest location. Extra domains are added to the child locations search query, when specified. Expects a singleton instance.                                                                                                                                                                              |
```
armandcela, line 157:
```suggestion
| _get_child_dest_locations       | Returns the child locations of the instance dest location. Extra domains are added to the child locations search query, when specified. Expects a singleton instance.                                                                                                                                                                              |
```

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+| get_move_lines                  | Get move lines associated to picking, uses functions in stock_move_line.                                                                                                                                                                                                                                                                          |
```
armandcela, line 158:
```suggestion
| get_move_lines                  | Gets move lines associated to picking, uses functions in stock_move_line.                                                                                                                                                                                                                                                                          |
```

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+| _backorder_move_lines           | Creates a backorder pick from self for all incomplete work, the original move retains with all completed work, new moves are created for the incomplete wrok.  The original pick is not validated. It should be called from within backorder_move_lines.                                                                                          |
+| _requires_backorder             | Checks if a backorder is required by checking if all move lines within a picking are present in mls. Cannot be consolidated with _check_backorder in Odoo core, because it does not take into account any move lines parameter.                                                                                                                   |
+| _prepare_picking_info           | Prepare the picking_info and products_info, returning the picking_info as a list of pickings, and products_info None or as a list of lists of a dictionary of the product info. This is to allow us to create multiple pickings at once.                                                                                                          |
```
armandcela, line 162:
```suggestion
| _prepare_picking_info           | Prepares the picking_info and products_info, returning the picking_info as a list of pickings, and products_info None or as a list of lists of a dictionary of the product info. This is to allow us to create multiple pickings at once.                                                                                                          |
```

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+| _prepare_move                   | Return a list of the move details to be used later in creation of the move(s). Note: pickings and products_info should be formatted via _prepare_picking_info. Allows for multiple moves to be created at once.                                                                                                                                   |
```
armandcela, line 163:
```suggestion
| _prepare_move                   | Returns a list of the move details to be used later in creation of the move(s). Note: pickings and products_info should be formatted via _prepare_picking_info. Allows for multiple moves to be created at once.                                                                                                                                   | 
```

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+| _prepare_move                   | Return a list of the move details to be used later in creation of the move(s). Note: pickings and products_info should be formatted via _prepare_picking_info. Allows for multiple moves to be created at once.                                                                                                                                   | 
+| _create_batch                   | Creates a batch                                                                                                                                                                                                                                                                                                                                   |
+| create_picking                  | Create and return a picking for the given picking_type. For multiple pickings a list of lists of dicts of product_info should be passed, and pickings with the same picking_type and other kwargs are the same. The kwargs are applied to pickings, not moves. If needed, the moves can be created outside of create_pickings with _create_moves. |
```
armandcela, line 165:
```suggestion
| create_picking                  | Creates and returns a picking for the given picking_type. For multiple pickings a list of lists of dicts of product_info should be passed, and pickings with the same picking_type and other kwargs are the same. The kwargs are applied to pickings, not moves. If needed, the moves can be created outside of create_pickings with _create_moves. |
```

addons/udes_stock/README.md
```diff
@@ -150,22 +151,24 @@ This is essentially a collection of products that are to be moved from one locat
...
+| _create_move                    | Create and return move(s) for the given move_values. Should be used in conjunction with _prepare_move to obtain move_values                                                                                                                                                                                                                       |
```
armandcela, line 166:
```suggestion
| _create_move                    | Creates and returns move(s) for the given move_values. Should be used in conjunction with _prepare_move to obtain move_values                                                                                                                                                                                                                       |
```

addons/udes_stock/data/locations.xml
```diff
@@ -33,5 +33,12 @@
       <field name="active" eval="True"/>
     </record>
 
+    <!-- Construct "UPL/Stock investigation" location -->
+    <record id="location_stock_investigation" model="stock.location">
+      <field name="name">Stock Investigation</field>
+      <field name="comment">Stock Investigation</field>
```
armandcela, line 39: Nit: It feels like the comment can be more detailed than name. Example:
Location used when , i am not good with English

ws-mps (approved):

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -45,6 +45,16 @@ def _domain_u_damaged_location_id(self):
         string="Default Goods Location",
         help="Goods receive location used by mobile client",
     )
+    u_missing_stock_location_id = fields.Many2one(
+        comodel_name="stock.location",
+        string="Default Missing Stock Location",
```
armandcela, line 50: Question: Why "Default Missing Stock" Location on the field, shouldn't it be simply "Missing Stock Location"?

LoKKeR96: This and other fields use the same naming, not entirely sure. I think maybe what it is supposed to mean is, the warehouse has some default locations for a specific type of storage and these are used as default locations in the picking types but they are not necessarily always the locations used for such storage? That's my interpretation. I guess they want to avoid making the user think that's always the case for those types of storage cases.

A TL/PO should have a better opinion on this.

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -983,3 +1009,202 @@ def remove_unfinished_work(self):
...
+        if not self.picking_type_id.u_enable_unpickable_items:
+            raise ValidationError(
+                _(
+                    "This type of operation cannot handle unpickable items. "
+                    "Please, contact your team leader to resolve the issue. "
+                    "Press back when resolved."
+                )
+            )
+
+        product = Product.get_or_create(product_id) if product_id else None
```
armandcela, line 1071:
```suggestion
        product = Product.get_or_create(product_id, return_empty=True)
```

addons/udes_stock/data/locations.xml
LoKKeR96: This is done. There's not really much to say about it.

addons/udes_stock_picking_batch/README.md
```diff
@@ -35,96 +35,97 @@ The main class for handling batch transfers in warehouse management
...
-| _compute_picking_type                     | Computes the picking_type_ids in the batch                                                                |
-
-| Global Methods                       | Description                                                                                             |
-|--------------------------------------|---------------------------------------------------------------------------------------------------------|
-| get_next_name                        | Get the next name for an object.          
+| Helpers                                  | Description                                                                                               |
+|------------------------------------------|-----------------------------------------------------------------------------------------------------------|
+| is_valid_location_dest_id                | Whether the specified location is a valid putaway location for the relevant pickings of the batch         |
+| add_extra_pickings                       | Get the next possible available pickings and add them to the batch that current user is operating         |
```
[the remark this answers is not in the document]
armandcela: Some verbs on description have s and some no , is nit happy to resolve

PR #614, opened 2022-04-27
Story/1176 Mobile UI: User Picking Traceability

armandcela:
Porting methods assign_picking_to_users and unassign_pickings_from_users to UDES14,


addons/udes_common/models/fields.py
```diff
@@ -0,0 +1,59 @@
...
+            if validate:
+                # force parsing for validation
+                self.from_string(value)
+            value = value[:DATETIME_LENGTH]
+            if len(value) == DATE_LENGTH:
+                value += " 00:00:00.000000"
+            return value
+        return self.to_string(value)
+
+    def convert_to_display_name(self, value, record):
```
peterpratt13, line 55: If this method were to strip the microseconds from the time, would this mean it could be rendered in the views? (Not sure we would ever want to show the microseconds in the UI.)

addons/udes_stock/README.md
```diff
@@ -160,7 +163,7 @@ This is essentially a collection of products that are to be moved from one locat
...
 | get_empty_pickings        | Generate batches marked with u_is_empty in a recordset or picking type (if u_auto_unlink_pickings is on)                                                                                                                                                                                                                                          |
 | unlink_empty              | Deletes empty pickings generated by get_empty_pickings                                                                                                                                                                                                                                                                                            |
-
+| un_assign_users           | Un assigning the user(s) from the pickings with optional skip_users
```
peterpratt13, line 166: super nit:
 assigning the user(s) from the pickings ... ->  assigning the user(s) from the pickings from self ...

addons/udes_common/models/fields.py
armandcela: Maybe i can try to use that method where i compute the datetime field which is shown in the UI.
The PreciseDatetime field was giving error on UI, thats why created another one datetime to show in UI

addons/udes_stock/README.md
armandcela: Should that be the same as the docstring in the definition of method, didnt understand the nit sorry :(

addons/udes_stock/models/res_users.py
```diff
@@ -1,10 +1,41 @@
...
+    )
+    u_picking_assigned_precise_time = PreciseDatetime(
+        string="User Date Started (microseconds)",
+        help="Precise date with microseconds when the user got assigned to the picking.",
+        readonly=True,
+        copy=False,
+    )
+    u_picking_assigned_time = fields.Datetime(
+        string="User Date Started",
+        help="Datetime when the user got assigned to the picking.",
```
kdwyer, line 28: It makes more sense to use "time" or "datetime" here rather than "date".

addons/udes_stock/models/res_users.py
```diff
@@ -1,10 +1,41 @@
...
 
+    u_picking_id = fields.Many2one(
+        "stock.picking",
+        string="Assigned Picking",
+        help="The picking a user is currently assigned to",
+        readonly=True,
+    )
+    u_picking_assigned_precise_time = PreciseDatetime(
+        string="User Date Started (microseconds)",
+        help="Precise date with microseconds when the user got assigned to the picking.",
```
kdwyer, line 22: It makes more sense to use "time" or "datetime" here rather than "date".

addons/udes_stock/models/res_users.py
```diff
@@ -15,9 +46,8 @@ def get_user_warehouse(self, aux_domain=None):
         :returns: Warehouse(s), or a singular warehouse if aux_domain not None
         """
         Warehouse = self.env["stock.warehouse"]
-
+
```
kdwyer, line 49: Unnecessary whitespace

addons/udes_stock/models/stock_picking.py
```diff
@@ -533,3 +551,36 @@ def unlink_empty(self):
             to_unlink.unlink()
 
         return self - records
+
+    def un_assign_users(self, skip_users=None):
```
peterpratt13, line 555: I'm unsure on this. But can we use unassign - not sure if it is a word or i have made it up

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -43,6 +43,20 @@ class StockPickingType(models.Model):
         the system.
         """,
     )
+    # Enable user process partial
+    u_user_process_partial = fields.Boolean(
```
peterpratt13, line 47: Personal nit: This name confuses me. Is it possible to have something else like
u_multi_users_enabled or something, that way I know that many can work on it at the same time. Currently in the UI i always have to look in the code, because i always think this means under receive something, or do part of it. 
But the teams leads may reject this, just a personal preference.

addons/udes_stock/models/res_users.py
```diff
@@ -37,3 +67,59 @@ def get_user_warehouse(self, aux_domain=None):
...
+        """
+        picking.ensure_one()
+        picking_type = picking.picking_type_id
+
+        # Find the users to assign
+        users_to_assign = self.filtered(lambda u: u.u_picking_id != picking)
+        if users_to_assign:
+            if picking.state in NON_ASSIGNABLE_STATES:
+                _logger.warning(
+                    "Cannot assign users %s to picking %s as it is in state %s"
```
kdwyer, line 83: We should probably wrap the message with `_(...)` here and in the else branch.
Using the `%` operator here prevents lazy evaluation of the final message; 
```python
_logger.warning(_('..%s ... %s'), 'spam', 'ham')
```
would be better.

armandcela (commented):
Dont know how i started the review, maybe github was not responding :)

addons/udes_stock/models/res_users.py
```diff
@@ -37,3 +67,59 @@ def get_user_warehouse(self, aux_domain=None):
...
+    def _assign_picking_to_users(self, picking):
+        """Placing into a method in order to be able to inherit and change behaviour if needed.
+        If the picking has never been started, update the start time."""
+        now = PreciseDatetime.now()
+        self.sudo().write(
+            {"u_picking_id": picking.id, "u_picking_assigned_precise_time": now})
+        if not picking.u_date_started:
+            picking.write({"u_date_started": now})
+
+    def un_assign_pickings_from_users(self, new_picking=False):
```
kdwyer, line 113: "unassign" is better here I think; in written English the "un" and "assign" are not separated.

armandcela: Not very confident with English, trustet Pycharm (it is showing underlined which shows when it's not in English) and googled after and it shows that should be separated, thats why changed the name

addons/udes_stock/models/res_users.py
```diff
@@ -37,3 +67,59 @@ def get_user_warehouse(self, aux_domain=None):
...
+                # is needed when closing batches or if we extend the behaviour of multiple users
+                # working on a picking.
+                picking.un_assign_users(skip_users=users_to_assign)
+                users_to_assign._assign_picking_to_users(picking)
+
+    def _assign_picking_to_users(self, picking):
+        """Placing into a method in order to be able to inherit and change behaviour if needed.
+        If the picking has never been started, update the start time."""
+        now = PreciseDatetime.now()
+        self.sudo().write(
```
peterpratt13, line 108: Q: Are the sudo calls required for 14 son the user model?

addons/udes_stock/models/stock_picking_type.py
armandcela: Will wait for a TL suggestion, but i agree with you :)

addons/udes_stock/views/res_users_view.xml
```diff
@@ -0,0 +1,57 @@
...
+    <field name="type">ir.actions.act_window</field>
+    <field name="view_mode">tree</field>
+    <field name="view_id" ref="picking_assigned_to_users_tree"/>
+    <field name="search_view_id" ref="picking_assigned_to_users_search"/>
+    <field name="help" type="html">
+      <p class="o_view_nocontent_empty_folder">
+        There is no picking assigned yet
+      </p>
+      <p>
+        This menu gives you the list of users and the picking they are assigned too.
```
kdwyer, line 47: Typo s/too/to

addons/udes_stock/models/stock_picking.py
armandcela: Not very confident with English, trustet Pycharm (it is showing underlined which shows when it's not in English) and googled after and it shows that should be separated, thats why changed the name

addons/udes_stock/tests/common.py
```diff
@@ -488,6 +488,18 @@ def get_picking_names(cls, pickings):
         names = pickings.mapped("name")
         return ", ".join(map(str, names))
 
+    @classmethod
+    def update_move_lines(cls, move_lines, qty=None, user=None):
+        if user is None:
+            user = cls.env.user
+        # We pass the user in the env as well, just for consistency
+        cls.env.real_uid = user.id
```
peterpratt13, line 496: Think we can drop this now - as traking doesn't revert to real_uid. The sudo calls don't (i beileve) mask the real user anymore.

addons/udes_stock/tests/test_stock_picking_user_assignment.py
```diff
@@ -0,0 +1,242 @@
...
+        self.inbound_user.assign_picking_to_users(self.goods_in_picking)
+        # Assert not assigned
+        self.assertFalse(self.inbound_user.u_picking_assigned_time)
+        self.assertFalse(self.inbound_user.u_picking_id)
+
+    def test_simple_assign_a_user(self):
+        """Assign a user through assign_picking_to_users method"""
+        self.assertFalse(self.inbound_user.u_picking_id)
+        self.inbound_user.assign_picking_to_users(self.goods_in_picking)
+        self.assertTrue(self.inbound_user.u_picking_id)
```
kdwyer, line 100:
```python
self.assertEqual(self.inbound_user.u_picking_id, self.goods_in_picking)
```
expresses our intent better perhaps?

addons/udes_stock/tests/test_stock_picking_user_assignment.py
```diff
@@ -0,0 +1,242 @@
...
+        Test if u_user_process_partial is not enabled, multiple users cannot be assigned
+        to the same picking.
+        """
+        self.picking_type_goods_in.u_user_process_partial = False
+        users = self.inbound_user | self.inbound_user_2 | self.inbound_user_3
+        self.assertEqual(self.goods_in_picking.state, "assigned")
+
+        # Assert nothing currently assigned
+        for user in users:
+            with self.subTest(user=user.name):
```
kdwyer, line 113: yay subtests :+1:

addons/udes_common/models/fields.py
peterpratt13: So i was wondering if this method is the one that formats it for the UI, and if it is just a precision issue, in this return the striped down datetime then it can be viewed without an additional computed field.

addons/udes_stock/models/res_users.py
armandcela: What does it mean "son", i think sudos are needed when we want to be sure that user have the right to run the code even if he hasnt the right to access in this case to write. And that is even for udes14 i think, at least saw some scenarios in udes14 odoo core modules where was used

peterpratt13:
Not in this sotry, but it would be good to use the u_done_datetime fields in udes_stock to use the precise datetime model as well.

addons/udes_stock/tests/test_stock_picking_user_assignment.py
armandcela: Copy paste from udes11, not mine :)

armandcela:
> Not in this sotry, but it would be good to use the u_done_datetime fields in udes_stock to use the precise datetime model as well.

Typo: story not sotry :)

addons/udes_stock/tests/test_stock_picking_user_assignment.py
```diff
@@ -0,0 +1,242 @@
...
+        self.assertEqual(self.inbound_user.u_picking_id, self.goods_in_picking)
+        self.assertEqual(start_time, self.inbound_user.u_picking_assigned_precise_time)
+
+    def test_un_assign_user_with_u_user_process_partial_enabled(self):
+        """Test un assignment of a picking when multiple users can exist on it"""
+        self.picking_type_goods_in.u_user_process_partial = True
+        user = self.inbound_user
+        # Manually assign a user to the picking
+        user.assign_picking_to_users(self.goods_in_picking)
+        self.assertTrue(user.u_picking_id)
```
kdwyer, line 191: Again, we should assert that the assigned picking is the expected picking, so an `assertEqual` would be better here.

addons/udes_stock/views/res_users_view.xml
```diff
@@ -0,0 +1,57 @@
...
+<odoo>
+  <record id="picking_assigned_to_users_tree" model="ir.ui.view">
+    <field name="name">res.users.assigned.picking.tree</field>
+    <field name="model">res.users</field>
+    <field name="arch" type="xml">
+      <tree string="Assigned Pickings" create="0" default_order="u_picking_assigned_time">
+        <field name="name"/>
+        <field name="login"/>
+        <field name="u_picking_id"/>
+        <field name="u_picking_assigned_time"/>
```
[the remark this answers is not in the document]
armandcela: In that case a related picking state field need to be added on res users , and i think user will not be assigned when picking is done.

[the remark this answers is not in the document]
peterpratt13: ah yeah - no lets not do that

kdwyer (approved):

addons/udes_common/models/fields.py
armandcela: Tbh tried to show the precise datetime in the UI , and could not. Would appreciate if there is a suggestion how, and i can give a try. 
In that case yes the computed field should be removed, as the reason added was only to show in UI

addons/udes_stock/models/res_users.py
kdwyer: Probably "unassign" is not an "official" word, so it isn't in the dictionary, but in general "un" and the wod it negates aren't separated.  And "unassign" is already used else in the Odoo codebase, we should be consistent with that.

peterpratt13: sorry typo son -> on.

Ah no we need it because it can assign multiple users info. I was just wondering if you can write on your own record without permission issues. But perhaps thinking about it that is probably not the case, as they could give themselvees to much power.

armandcela: Making the changes of the methods again :)

addons/udes_common/models/fields.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @staticmethod
+    def now(*args):
+        """ Return the current day and time in the format expected by the ORM.
+            This function may be used to compute default values.
+        """
+        return dt.datetime.now().strftime(DATETIME_FORMAT)
+
+    @staticmethod
+    def from_string(value):
```
peterpratt13, line 25: Can we change to `to_datetime`  instead to be consistent with Odoo versioning. In v12 they say they are dropping it, but still there atm :)

addons/udes_stock/models/res_users.py
kdwyer: I think the `string` value needs to the same change?

addons/udes_stock/models/res_users.py
kdwyer: Some of the tests still have `un_assign` in their names

addons/udes_common/models/fields.py
peterpratt13: So just for visibility: We had a chat offline and really I thought we want the precise datetime in most places (like u_done_datetime etc) so the events have it but in the frontend everything gets trimmed to ignore microseconds. 

1. Handle datetime as a precise datetime everywhere- probably to risky for this story.
2. Set the same type as the datetime field, i.e type="datetime" so odoo front end renders it like datetime. But that does mean we have two feilds with the same type.
3. Add the new field properly, this involves files like basic_fields.js, field_registry etc. updating them for the new field. 

Only 2 is probably acceptable for this story. 

The main issue was raised is that if we want the precise datetime for events, and want to show any of those fields in the UI will need an additional computed field. This seems like over kill, 1 might be ok, but don't want to let it grow really.

I would prefer maybe an issue to either do 1 or 3, not sure any side affects of 2.

addons/udes_stock/tests/test_stock_picking_user_assignment.py
```diff
@@ -0,0 +1,242 @@
...
+        """Test un assignment of a picking when multiple users can exist on it"""
+        self.picking_type_goods_in.u_user_process_partial = True
+        user = self.inbound_user
+        # Manually assign a user to the picking
+        user.assign_picking_to_users(self.goods_in_picking)
+        self.assertEqual(user.u_picking_id, self.goods_in_picking)
+        user.unassign_pickings_from_users()
+        self.assertFalse(user.u_picking_id)
+
+    def test_un_assign_multiple_users(self):
```
peterpratt13, line 195: nit: un_assign -> unassign  for the test names as well please. Sorry, did a  search

addons/udes_common/models/fields.py
armandcela: Changed the type to datetime in PreciseDatetime class, assigned a picking to a user and when psql on database the datetime field shows with microseconds while in UI without and doesnt raise any error which i think is waht we need.

addons/udes_stock/models/res_users.py
armandcela: Changed

addons/udes_stock/models/res_users.py
armandcela: Checked all strings where was un_assign, i think there isnt any left

addons/udes_common/models/fields.py
armandcela: When i change from_string method to to_datetime it raises error.

addons/udes_common/models/fields.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @staticmethod
+    def to_string(value):
+        """ Convert a :class:`datetime` value into the format expected by the ORM. """
+        return value.strftime(DATETIME_FORMAT) if value else False
+
+    def convert_to_column(self, value, record, values=None):
+        return super(PreciseDatetime, self).convert_to_column(value or None, record, values)
+
+    def convert_to_cache(self, value, record, validate=True):
```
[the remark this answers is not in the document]
armandcela: As from_string replacement to_datetime gives error i think we should let at the moment convert_to_cache

addons/udes_stock/tests/test_stock_picking_user_assignment.py
armandcela: Changed

addons/udes_stock/models/res_users.py
kdwyer: We're still using the `%` operator?
```python
                _logger.warning(
                    _("Cannot assign users %s to picking %s as picking type %s only allows one user"),
                   (self.mapped("name"), picking.name, picking_type.name)
                )
```
is better; if the log message is not emitted the complete string will never be constructed.

armandcela: I thought was translatable thing as the % operator it is used widely in odoo core UserErrors etc

kdwyer: UserErrors are a different thing.

If I do `s = "spam %s' % 'eggs'" then the formatting is executed immediately - s is always 'spam eggs'.

The logging package does not behave in quite the same way as normal code.  With a logging statement like `_logger.info('spam %s', 'eggs')`, the logging package computes whether or not the message will be emitted, and only executes the formatting if the message will be emitted.  However if we do `_logger.info('spam %s' % 'eggs')` then the formatting is evaluated _before_ the logging package determines whether the message will ever be logged.  So not using `%` in this way is a (very small) efficiency gain.

addons/udes_common/models/fields.py
peterpratt13: What error does it raise? Anything useful?

addons/udes_common/models/fields.py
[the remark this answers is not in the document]
peterpratt13: I think the type version returned by convert_to_cache has changed. In 14 it returns a datetime, in 11 i think it is a string

addons/udes_common/models/fields.py
adamrpatrick: @peterpratt13 As you say I think option 2 is best for this story unless we have a breakthrough for the other ones. I'll create an issue to handle precise date times nicely in the front end.

addons/udes_stock/models/stock_picking_type.py
adamrpatrick: I agree personally, it's not abundantly clear what this flag is doing. Happy to rename the field to something clear and update the help text. Also update the README so we know what this field used to be called. `u_multi_users_enabled` is a good suggestion.

adamrpatrick (commented):
LGTM once the remaining comments have been resolved. The currently failing pipeline will be resolved when new images of odoo- and dues-tester have been updated, so keep an eye out for any _new_ failures until then.

peterpratt13 (approved):

PR #615, opened 2022-04-27
Story/x1175 rewind

LoKKeR96:


LoKKeR96:
This was used to work on story while backordering logic was still being re-implemented.

PR #616, opened 2022-04-28
wip

ws-mps:
wip


addons/udes_stock/models/stock_move.py
```diff
@@ -185,9 +193,8 @@ def split_out_incomplete_move(self, **kwargs):
         total_done_qty = sum(mls.mapped("qty_done"))
```
ws-mps, line 193: Maybe this should be total quantity of the mls instead of only qty done when mls_to_keep is not None.

PR #617, opened 2022-04-28
[FIX] udes_stock: Stops backorder being created for cancelled movelines

caleb-shelton:
User-story: 1773

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -963,19 +963,17 @@ class StockPicking(models.Model):
             )
 
     def _requires_backorder(self, mls):
-        """Checks if a backorder is required
-        by checking if all move.lines
-        within a picking is present in mls
+        """
+        Checks if a backorder is required by checking if all moves within a picking are present
+        in the movelines. Cancelled moves are not considered.
         """
         mls_moves = mls.mapped("move_id")
-        # return (mls_moves | self.move_lines) != mls_moves or \
-        # (mls | self.move_line_ids) != mls or \
-        # self.mapped('move_lines.move_orig_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
-        for move in self.move_lines:
+        
+        for move in self.move_lines.filtered(lambda mv: mv.state != "cancel"):
             if (
                 move not in mls_moves
-                or not move.move_line_ids == mls.filtered(lambda x: x.move_id == move)
-                or move.move_orig_ids.filtered(lambda x: x.state not in ("done", "cancel"))
+                or not move.move_line_ids == mls.filtered(lambda ml: ml.move_id == move)
+                or move.move_orig_ids.filtered(lambda mv: mv.state not in ("done", "cancel"))
             ):
                 return True
         return False
--- a/addons/udes_stock/tests/test_backorder.py
+++ b/addons/udes_stock/tests/test_backorder.py
@@ -1,0 +1,51 @@
+from . import common
+
+
+class TestBackOrder(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestBackOrder, cls).setUpClass()
+
+        # Create two quants in location_01
+        cls.create_quant(cls.apple.id, cls.test_location_01.id, 5)
+        cls.create_quant(cls.banana.id, cls.test_location_01.id, 5)
+        # Create a picking for two different products
+        cls.pick1 = cls.create_picking(
+            picking_type=cls.picking_type_pick,
+            products_info=[
+                {"product": cls.apple, "qty": 5},
+                {"product": cls.banana, "qty": 5},
+            ],
+            assign=True,
+        )
+
+    def test_requires_backorder_for_incomplete_moves(self):
+        """
+        Test a picking is flagged as needing a backorder if it has incomplete
+        moves.
+        """
+
+        # Set qty done to 5 for apple and leave as 0 banana
+        apple_move_lines = self.pick1.move_line_ids.filtered(lambda ml: ml.product_id == self.apple)
+        apple_move_lines.qty_done = 5
+
+        # Assert that this picking requires a backorder
+        self.assertTrue(self.pick1._requires_backorder(apple_move_lines))
+
+    def test_backorder_not_created_for_cancelled_moves(self):
+        """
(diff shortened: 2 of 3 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -963,19 +963,17 @@ def update_picking(
...
+        Checks if a backorder is required by checking if all moves within a picking are present
+        in the movelines. Cancelled moves are not considered.
         """
         mls_moves = mls.mapped("move_id")
-        # return (mls_moves | self.move_lines) != mls_moves or \
-        # (mls | self.move_line_ids) != mls or \
-        # self.mapped('move_lines.move_orig_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
-        for move in self.move_lines:
+        
+        for move in self.move_lines.filtered(lambda ml: ml.state != "cancel"):
```
PeterC10, line 972: Nit: Can we change the lamba variable to `mv`? As it is a move record, not a move line (thank you Odoo for the field names...)

addons/udes_stock/tests/test_backorder.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+    def test_requires_backorder_for_incomplete_moves(self):
+        """
+        Test a picking is flagged as needing a backorder if it has incomplete
+        moves.
+        """
+
+        # Set qty done to 5 for apple and leave as 0 banana
+        apple_move_lines = self.pick1.move_lines.filtered(lambda ml: ml.product_id == self.apple)
```
PeterC10, line 29: Q: Does this give a warning in the terminal? `qty_done` is on the `stock.move.line` record, not `stock.move`.

We can just do:

`apple_move_line = self.pick1.move_line_ids.filtered(lambda ml: ml.product_id == self.apple)` (Suggest just `apple_move_line` as it is only one record)

This always shows that `_requires_backorder` doesn't check if the move lines that it is given are done... but that is not for this story. :)

addons/udes_stock/tests/test_backorder.py
```diff
@@ -0,0 +1,51 @@
...
+        moves are cancelled.
+        """
+        # Set qty done to 5 on apple
+        apple_move_lines = self.pick1.move_lines.filtered(lambda ml: ml.product_id == self.apple)
+        apple_move_lines.qty_done = 5
+
+        # Cancel the banana move
+        banana_move = self.pick1.move_lines.filtered(
+            lambda ml: ml.product_id == self.banana
+        ).ensure_one()
```
PeterC10, line 47: Q: Why do we do `ensure_one()` here? As we setup the picking I think we can safely assume there is only one move for the banana product, but happy for it to stay.

caleb-shelton: I was just eering on the side of caution as I know this _can_ contain more than one, and maybe if someone adjusts the setup in the future as this file is for generic backorder testing.

addons/udes_stock/models/stock_picking.py
```diff
@@ -963,19 +963,17 @@ def update_picking(
...
+        Checks if a backorder is required by checking if all moves within a picking are present
+        in the movelines. Cancelled moves are not considered.
         """
         mls_moves = mls.mapped("move_id")
-        # return (mls_moves | self.move_lines) != mls_moves or \
-        # (mls | self.move_line_ids) != mls or \
-        # self.mapped('move_lines.move_orig_ids').filtered(lambda x: x.state not in ('done', 'cancel'))
-        for move in self.move_lines:
+        
+        for move in self.move_lines.filtered(lambda mv: mv.state != "cancel"):
```
peterpratt13, line 972: Q: Do we also want to move the mls into a backorder if the remaining moves are done? Not sure just a question.

peterpratt13, line 972: Q: Do we also want to move the mls into a backorder if the remaining moves are done? Not sure just a question.

addons/udes_stock/tests/test_backorder.py
```diff
@@ -0,0 +1,51 @@
...
+        # Assert that this picking requires a backorder
+        self.assertTrue(self.pick1._requires_backorder(apple_move_lines))
+
+    def test_backorder_not_created_for_cancelled_moves(self):
+        """
+        Test a picking is not flagged as needing a backorder if the remaining incomplete
+        moves are cancelled.
+        """
+        # Set qty done to 5 on apple
+        apple_move_lines = self.pick1.move_line_ids.filtered(lambda ml: ml.product_id == self.apple)
```
peterpratt13, line 41: nit: Do we need to set the quantity? I think not setting ut should be fine provided we just pass the apple mls to the method

peterpratt13, line 41: nit: Do we need to set the quantity? I think not setting ut should be fine provided we just pass the apple mls to the method

addons/udes_stock/models/stock_picking.py
peterpratt13: We might not get to that actually, maybe the moves are done on drop off always then backorder. Rather than drop and keep the picking open

peterpratt13: We might not get to that actually, maybe the moves are done on drop off always then backorder. Rather than drop and keep the picking open

addons/udes_stock/tests/test_backorder.py
peterpratt13: don't know why it created this comment twice...

peterpratt13: don't know why it created this comment twice...

PeterC10: Nah we don't, the `_requires_backorder` doesn't check the quantities or anything like that, but no harm in having a semi-realistic setup.

addons/udes_stock/models/stock_picking.py
PeterC10: I think this would be covered in the check below, I think for what we need in this story just filtering out cancelled moves should be enough anyway.

Obviously there is a lot we can do better here in general.

PeterC10 (approved):
LGTM

peterpratt13 (approved):

PeterAlabaster (approved):

PR #618, opened 2022-04-28
[ADD] udes_sale_stock: Merge Confirm Orders and  Reserve Stock Cron Job

armandcela:
Story: 1685

### DIFF
--- a/addons/udes_sale_stock/data/ir_cron.xml
+++ b/addons/udes_sale_stock/data/ir_cron.xml
@@ -16,5 +16,21 @@
             <field name="code">model.confirm_orders()</field>
         </record>
 
+        <!-- Confirm Sale Orders and Reserve Stock in a single cron job -->
+        <record id="merge_confirm_orders_reserve_stock" model="ir.cron">
+          <field name="name">Merged Confirm Orders and Reserve Stock</field>
+            <field name="active" eval="False" />
+            <field name="user_id" ref="base.user_root" />
+            <field name="interval_number">5</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field name="doall">0</field>
+            <field name="model_id" ref="udes_sale_stock.model_sale_order"/>
+            <field name="state">code</field>
+            <field name="code">
+                model.merge_confirm_orders_reserve_stock(confirm_orders_batch_size=1000, confirm_orders_commit_size=1000, finish=False)
+            </field>
+        </record>
+
 	</data>
 </odoo>
--- a/addons/udes_sale_stock/data/stock_config.xml
+++ b/addons/udes_sale_stock/data/stock_config.xml
@@ -12,5 +12,11 @@
       <value eval="[ref('settings_stock_config')]"/>
     </function>
 
+    <!-- Setting batch size to split orders to confirm when is needed especially on specific requirements -->
+    <record id="confirm_orders_batch_size" model="ir.config_parameter">
+      <field name="key">udes_sale_stock.confirm_orders_batch_size</field>
+      <field name="value" eval="10"/>
+    </record>
+
   </data>
 </odoo>
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -315,7 +315,7 @@ class SaleOrder(models.Model):
         """Returns the domain for sale orders to attempt confirmation."""
         return [("state", "=", "draft")]
 
-    def confirm_orders(self):
+    def confirm_orders(self, size=1000):
         """Attempt to confirm sale orders.
 
         If self is empty, find orders to confirm via _get_confirmation_domain.
@@ -341,7 +341,7 @@ class SaleOrder(models.Model):
         else:
             to_confirm = self.search(self._get_confirmation_domain())
 
-        for _, batch in to_confirm.batched(size=1000):
+        for _, batch in to_confirm.batched(size=size):
             tries = 0
             while True:
                 try:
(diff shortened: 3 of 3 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -393,6 +393,53 @@ def extract_exception_data(err):
...
+            remaining orders to confirm on next cron run.
+        """
+        SaleOrder = self.env["sale.order"]
+        StockPicking = self.env["stock.picking"]
+
+        orders_to_confirm = SaleOrder.get_orders_to_confirm()
+        # If there aren't any orders to confirm just run the reserve stock cron
+        if not orders_to_confirm:
+            StockPicking.reserve_stock()
+        for _, batched_orders in orders_to_confirm.batched(size=confirm_orders_batch_size):
```
jrbirch99, line 429: nit: I know _ is not used in the for loop but maybe you could give it a different name like batch_range, just so it doesn't get confused with the translator function _()

armandcela: When it is not used if have a variable name, it would complain that unused variable name. In fact we dont need it to use. Seen in other cases the same approach thats why did in the same way

armandcela: When it is not used if have a variable name, it would complain that unused variable name. In fact we dont need it to use. Seen in other cases the same approach thats why did in the same way

addons/udes_sale_stock/data/stock_config.xml
```diff
@@ -12,5 +12,11 @@
       <value eval="[ref('settings_stock_config')]"/>
     </function>
 
+    <!-- Setting batch size to split orders to confirm when is needed especially on specific requirements -->
+    <record id="orders_confirm_batch_size" model="ir.config_parameter">
+      <field name="key">udes_sale_stock.orders_confirm_batch_size</field>
```
jrbirch99, line 17: Should this be `confirm_orders_batch_size`?

jrbirch99:
Also does there need to be unit tests for this new method?

addons/udes_sale_stock/data/ir_cron.xml
```diff
@@ -16,5 +16,21 @@
...
+            <field name="active" eval="False" />
+            <field name="user_id" ref="base.user_root" />
+            <field name="interval_number">5</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field name="doall">0</field>
+            <field name="model_id" ref="udes_sale_stock.model_sale_order"/>
+            <field name="state">code</field>
+            <field name="code">
+                model.merge_confirm_orders_reserve_stock(confirm_orders_batch_size=1000, confirm_orders_commit_size=500, finish=False)
```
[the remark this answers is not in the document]
armandcela: Thought to make reduce the commit size by default because after that is reserve_stock cron job too. There is not any very clear idea why, but can be changed through UI. I put in the cron method to be easier to change it from UI so we know what are the optional variable names

armandcela:
> Should there need to be unit tests for this new method?

It is not in the story :). Normally the confirm and reserve stock are existing methods

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -393,6 +393,53 @@ def extract_exception_data(err):
...
+        6. Not confirmed orders and not reserved stock will be picked in the
+        next cron job if finish is set to true.
+        :param int confirm_orders_batch_size:
+            Number of orders will be confirmed before reserving stock.
+        :param int confirm_orders_commit_size:
+            Number of orders will be confirmed before committing.
+        :param boolean finish:
+            Flag if enabled will finish the cron job after the first iteration and pick the
+            remaining orders to confirm on next cron run.
+        """
```
LoKKeR96, line 421:
```suggestion
        """Confirm orders and Reserve Stock in a single cron job so the two processes can be
        run simultaneously.
        The process is described below:
        1. Reserve stock if there are no orders to confirm.
        2. Split all orders to confirm in sizes with confirm_orders_batch_size.
        3. For every orders_to_confirm batch ,confirm and commit
            orders within confirm_orders_commit_size.
        4. After confirming orders_to_confirm batch, reserve stock.
        5. If needed to confirm only 1 confirm_orders_batch_size finish variable
        will get out and finish the cron job.
        6. Not confirmed orders and not reserved stock will be picked in the
        next cron job if finish is set to true.
        :param int confirm_orders_batch_size:
            Number of orders to be confirmed before reserving stock.
        :param int confirm_orders_commit_size:
            Number of orders to be confirmed before committing.
        :param boolean finish:
            Flag if True will finish the cron job after the first iteration and pick the
            remaining orders to confirm on the next cron run.
        """
```

LoKKeR96 (commented):
LGTM, waiting for comments.

LoKKeR96 (commented):
Suggestion: I noticed there isn't any README. It would be nice to have a summary of what functionality does the module provide.  Have you got time in the story? If not I will close the comment.

armandcela:
> Suggestion: I noticed there isn't any README. It would be nice to have a summary of what functionality does the module provide. Have you got time in the story? If not I will close the comment.

I don't think is a matter of time, the doc string is a lot explanatory even better with your suggestions.

LoKKeR96:
> > Suggestion: I noticed there isn't any README. It would be nice to have a summary of what functionality does the module provide. Have you got time in the story? If not I will close the comment.
> 
> I don't think is a matter of time, the doc string is a lot explanatory even better with your suggestions.

I meant having a README for the whole module. Either way no problem.

LoKKeR96 (approved):

armandcela:
> > > Suggestion: I noticed there isn't any README. It would be nice to have a summary of what functionality does the module provide. Have you got time in the story? If not I will close the comment.
> > 
> > 
> > I don't think is a matter of time, the doc string is a lot explanatory even better with your suggestions.
> 
> I meant having a README for the whole module. Either way no problem.

I dont know what is udes_sale_stock doing, or you mean copy-pastes? And dont think is a part of this story anyway

jrbirch99 (approved):

PeterAlabaster (approved):

PR #619, opened 2022-04-29
#1178 Mobile UI: Printers and users

jrbirch99:
[ADD] udes_stock: A tools folder was added with needed print method.

Story/1178

### DIFF
--- a/addons/udes_stock/tools/__init__.py
+++ b/addons/udes_stock/tools/__init__.py
@@ -1,0 +1,17 @@
+"""Tools and utilties for the udes_stock addon."""
+
+from odoo.exceptions import ValidationError
+from odoo.tools.translate import _
+
+
+def get_printer_by_barcode(env, barcode):
+    """Return the printer with the provided barcode."""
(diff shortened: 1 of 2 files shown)


addons/udes_stock/tools/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""Tools and utilties for the udes_stock addon."""
+
+from odoo.exceptions import ValidationError
+from odoo.tools.translate import _
+
+
+def get_printer_by_barcode(env, barcode):
```
PeterC10, line 7: Don't think this should be here.

We don't have a dependency on the print module and if it is just a helper method for returning a printer we might as well just put it in the print module.

jrbirch99: This will be moved to the print module. It will go in print/tools/__init__.py.

PeterC10: Cool, as this is the only commit here I'll go close the PR.

PR #621, opened 2022-05-04
[IMP] udes_stock: Ensure last reserved pallet is cleared on drop-off

LoKKeR96:
Fixes the bug where the pallet is asked to be scanned again after
being dropped off in workflows.

User-story: 1827

### DIFF
--- a/addons/udes_stock/models/stock_picking_batch.py
+++ b/addons/udes_stock/models/stock_picking_batch.py
@@ -762,10 +762,16 @@ class StockPickingBatch(models.Model):
 
     def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, result_package_name):
         """
-        Validate the move lines of the batch (expects a singleton)
-        by moving them to the specified location.
+        Validate the move lines of the batch (expects a singleton) by moving them
+        to the specified location and if continue_batch is not flagged then close the batch.
+        Also clears the u_last_reserved_pallet_name flag in the batch as the pallet has been dropped off.
 
-        In case continue_batch is not flagged, close the batch.
+        :args:
+            - continue_batch: Flag, if True the batch is not closed
+            - move_line_ids: model stock.move.line, the move lines to validate
+            - location_barcode: String, the destination location barcode for the move lines
+            - result_package_name: String, the barcode for the package to be set as result package in move lines
+        :returns: the batch in self
         """
         self.ensure_one()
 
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -829,6 +829,10 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
                 picks_todo |= pick_todo
                 pick.write({"u_reserved_pallet": False})
 
+                # If we dropped a package then reset the last package reserved as
+                # it will not be possible to keep using it
+                self.u_last_reserved_pallet_name = False
```
armandcela, line 834: Is the update inside the loop?

armandcela:
I think the docstring should be updated to mention that at the end u_last_reserved_pallet_name will be False

addons/udes_stock/models/stock_picking_batch.py
LoKKeR96: Yep, should be outside to be correct, fixing up soon.

jrbirch99:
Q: In the story it mentions two MRs one branching off of master, and the other off of tag 11.14.2. Should there be another PR for the tagged version?

LoKKeR96:
> Q: In the story it mentions two MRs one branching off of master, and the other off of tag 11.14.2. Should there be another PR for the tagged version?

Yes, I wanted to have a word with Adam first, also making sure code is reviewed before pushing another MR and maintaining two which is painful.

PeterAlabaster (approved):
LGTM once loop comment addressed

armandcela (approved):

jrbirch99:
> > Q: In the story it mentions two MRs one branching off of master, and the other off of tag 11.14.2. Should there be another PR for the tagged version?
> 
> Yes, I wanted to have a word with Adam first, also making sure code is reviewed before pushing another MR and maintaining two which is painful.

Okay, but its the same fix, right?

LoKKeR96:
> > > Q: In the story it mentions two MRs one branching off of master, and the other off of tag 11.14.2. Should there be another PR for the tagged version?
> > 
> > 
> > Yes, I wanted to have a word with Adam first, also making sure code is reviewed before pushing another MR and maintaining two which is painful.
> 
> Okay, but its the same fix, right?

Yes, the story is written a bit ambiguously and it's not clear what they mean with tagging.
A TL will tag master once this story is finished and merged.

jrbirch99 (approved):

PR #623, opened 2022-05-04
[ADD] udes_open: Reordering rules constraint

caleb-shelton:
Port reordering rules constraint.
Add UI field for limiting orderpoints per location.

User-story: 1180

### DIFF
--- a/addons/udes_stock/data/routes.xml
+++ b/addons/udes_stock/data/routes.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+    <!-- Add a new route for Replen -->
+    <record id="route_warehouse0_replen" model='stock.location.route'>
+      <field name="name">Replen</field>
+      <field name="company_id"></field>
+      <field name="sequence">10</field>
+    </record>    
+
+    <!-- Create new pull rule with stock as procurement location -->
+    <record id="procurement_replen" model="stock.rule">
+      <field name="name">Replen</field>
+      <field name="route_id" ref="route_warehouse0_replen"/>
+      <field name="sequence" eval="10"/>
+      <field name="location_id" ref="stock.stock_location_output"/>
+      <field name="location_src_id" ref="stock.stock_location_stock"/>
+      <field name="action">pull</field>
+      <field name="picking_type_id" ref="udes_stock.picking_type_replen"/>
+      <field name="warehouse_id" ref="stock.warehouse0"/>
+    </record>
+  </data>
+</odoo>
--- a/addons/udes_stock/data/scheduler.xml
+++ b/addons/udes_stock/data/scheduler.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+
+        <record id="stock.ir_cron_scheduler_action" model="ir.cron">
+            <field eval="False" name="active"/>
+        </record>
+
+        <record id="ir_cron_scheduler_action_orderpoints" model="ir.cron">
+            <field name="name">Check order points</field>
+            <field name="model_id" ref="stock.model_stock_warehouse_orderpoint"/>
+            <field name="state">code</field>
+            <field name="code">
+model.check_order_points(True)
+</field>
+            <field eval="True" name="active"/>
+            <field name="user_id" ref="base.user_root"/>
+            <field name="interval_number">1</field>
+            <field name="interval_type">days</field>
+            <field name="numbercall">-1</field>
+            <field eval="False" name="doall"/>
+        </record>
+
+    </data>
+</odoo>
--- a/addons/udes_stock/models/procurement.py
+++ b/addons/udes_stock/models/procurement.py
@@ -1,4 +1,4 @@
-from odoo import api, models, _
+from odoo import api, models, _, registry
 
 
 class ProcurementGroup(models.Model):
@@ -8,3 +8,45 @@ class ProcurementGroup(models.Model):
 
     # Allow groups to be created
     MSM_CREATE = True
+
+    def _get_orderpoint_domain(self, company_id=False):
+        """ Override order point domain to filter by active_ids
+        """
+        domain = super()._get_orderpoint_domain(company_id=company_id)
+        orderpoints = self.env.context.get('active_ids', None)
+        if orderpoints:
+            domain += [('id', 'in', orderpoints)]
+        return domain
+
+        
+class StockWarehouseOrderPoint(models.Model):
+    _inherit = 'stock.warehouse.orderpoint'
+
+    @api.model
+    def check_order_points(self, use_new_cursor=False, company_id=False):
+        """
+        Copy of run_scheduler from Odoo's stock module ProcurementGroup class.
+        This allows us to only check order points.
+        """
+        OrderPoint = self.env["stock.warehouse.orderpoint"]
+        ProcurementGroup = self.env["procurement.group"]
+        try:
+            if use_new_cursor:
+                domain = ProcurementGroup._get_orderpoint_domain()
+                self = OrderPoint.with_context(prefetch_fields=False).search(domain)
+                cr = registry(self._cr.dbname).cursor()
+                self = self.with_env(self.env(cr=cr))
+
+            self.sudo()._procure_orderpoint_confirm(
+                use_new_cursor=use_new_cursor,
+                company_id=company_id)
+            if use_new_cursor:
+                self._cr.commit()
+        finally:
+            if use_new_cursor:
+                try:
+                    self._cr.close()
+                except Exception:
+                    pass
+        return {}
+
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -48,6 +48,12 @@ class StockLocation(models.Model):
         help="Used to know which pickers have the right equipment to pick from it.",
     )
 
+    u_limit_orderpoints = fields.Boolean(
+        index=True,
+        string="Limit Orderpoints",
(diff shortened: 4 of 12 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -48,6 +48,12 @@ def _domain_height_category(self):
         help="Used to know which pickers have the right equipment to pick from it.",
     )
 
+    u_limit_orderpoints = fields.Boolean(
+        index=True,
+        string="Limit Orderpoints",
+        help="If set, allow only one orderpoint on this location " "and its descendants.",
```
PeterAlabaster, line 54: There's a break in the quotes here, not sure if it's intentional?

PeterAlabaster (approved):
LGTM one comment

LoKKeR96 (approved):

armandcela (changes requested):

addons/udes_stock/models/stock_location.py
```diff
@@ -70,3 +76,13 @@ def get_common_ancestor(self):
...
+
+    def limits_orderpoints(self):
+        """Determines whether this location, or an ancestor, permits only a
+        single orderpoint on itself.
+
+        Returns: a boolean: True if limited, False otherwise.
+        """
+        self.ensure_one()
+        limited = self.search([("u_limit_orderpoints", "=", True)])
+        return bool(self.search_count([("id", "child_of", limited.ids), ("id", "=", self.id)]))
```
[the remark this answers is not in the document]
armandcela: I think should be an and as checked on udes11 too and maybe got confused with the doc string

peterpratt13:
Late to the party but LGTM, Can ignore, but can you just run black on it lots of single quotes. :)

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,42 @@ class ProcurementGroup(models.Model):
...
+            domain += [('id', 'in', orderpoints)]
+        return domain
+
+        
+class StockWarehouseOrderPoint(models.Model):
+    _inherit = 'stock.warehouse.orderpoint'
+
+    @api.model
+    def check_order_points(self, use_new_cursor=False, company_id=False):
+        """  Copy of run_scheduler to only check order points """
```
kdwyer, line 27: A little more explanation would be helpful here, for example:
- why would we use a new cursor?
- where is the run_scheduler that we are copying?

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,42 @@ class ProcurementGroup(models.Model):
...
+        OrderPoint = self.env["stock.warehouse.orderpoint"]
+        ProcurementGroup = self.env["procurement.group"]
+        try:
```
kdwyer, line 30: Everything is guarded by `use_new_cursor`, so if it's `False` nothing will happen - is that our intention?  If so this would also merit a comment.

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,42 @@ class ProcurementGroup(models.Model):
...
+            if use_new_cursor:
+                cr = registry(self._cr.dbname).cursor()
+                domain = ProcurementGroup._get_orderpoint_domain()
+                self = OrderPoint.with_context(prefetch_fields=False).search(domain)
+                self = self.with_env(self.env(cr=cr))
+
+                self.sudo()._procure_orderpoint_confirm(
+                    use_new_cursor=use_new_cursor,
+                    company_id=company_id)
+            if use_new_cursor:
```
kdwyer, line 40: Redundant `if`?

PeterAlabaster: me and caleb noticed this too, it exists in 11 too and can probably be removed, i can't see any reason for it existing

addons/udes_stock/models/procurement.py
kdwyer: It's copied  from odoo's stock procurement.py

addons/udes_stock/models/procurement.py
kdwyer: Looking at the original, I think the `sudo` call above should be dedented, so it is always executed.  But it's the same in 11 so it could be intentional.  Maybe @ws-mps can remember?

addons/udes_stock/data/scheduler.xml
```diff
@@ -0,0 +1,26 @@
...
+            <field eval="False" name="active"/>
+        </record>
+
+        <record id="ir_cron_scheduler_action_orderpoints" model="ir.cron">
+            <field name="name">Check order points</field>
+            <field name="model_id" ref="stock.model_stock_warehouse_orderpoint"/>
+            <field name="state">code</field>
+            <field name="code">
+model.check_order_points(True)
+model._cr.commit()
```
caleb-shelton, line 15: Also not sure if this is needed, this was in 11

ws-mps: Probably not needed, in 11 we had more code.

addons/udes_stock/models/procurement.py
ws-mps: I agree, seems that is to be dedented, probably a typo we never noticed.

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,44 @@ class ProcurementGroup(models.Model):
...
+                cr = registry(self._cr.dbname).cursor()
+                domain = ProcurementGroup._get_orderpoint_domain()
+                self = OrderPoint.with_context(prefetch_fields=False).search(domain)
+                self = self.with_env(self.env(cr=cr))
+
+            self.sudo()._procure_orderpoint_confirm(
+                use_new_cursor=use_new_cursor,
+                company_id=company_id)
+            self._cr.commit()
```
kdwyer, line 43: I think we should be committing `cr` here, not `self._cr`? (once the guard on `use_new_cursor` is reinstated)

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,44 @@ class ProcurementGroup(models.Model):
...
+        finally:
+            if use_new_cursor:
+                try:
+                    self._cr.close()
```
kdwyer, line 47: Likewise we should be closing `cr`, not `self._cr`.

addons/udes_stock/models/procurement.py
kdwyer: Nope, I didn't noticed we'd redefined `self` :facepalm:

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+    <!-- Replace pull_ids of route_warehouse0_replen by the new procurement rule -->
```
PeterAlabaster, line 4: Are we actually replacing anything here? It looks more like we're just creating a new route with no rules.

I don't think route_warehouse0_replen exists at the point this is run - so should the comment not be:
```xml
<!-- Add a new route for Replen -->
```

kdwyer (approved):

addons/udes_stock/data/routes.xml
```diff
@@ -0,0 +1,23 @@
...
+    <record id="route_warehouse0_replen" model='stock.location.route'>
+      <field name="name">Replen</field>
+      <field name="company_id"></field>
+      <field name="sequence">10</field>
+    </record>    
+
+    <!-- Create new buy procurement rule with stock as procurement location -->
```
PeterAlabaster, line 11:
```suggestion
    <!-- Create new pull rule with stock as procurement location -->
```

addons/udes_stock/models/procurement.py
```diff
@@ -8,3 +8,45 @@ class ProcurementGroup(models.Model):
...
+        This allows us to only check order points.
+        """
+        OrderPoint = self.env["stock.warehouse.orderpoint"]
+        ProcurementGroup = self.env["procurement.group"]
+        try:
+            if use_new_cursor:
+                cr = registry(self._cr.dbname).cursor()
+                domain = ProcurementGroup._get_orderpoint_domain()
+                self = OrderPoint.with_context(prefetch_fields=False).search(domain)
+                self = self.with_env(self.env(cr=cr))
```
kdwyer, line 38: NIT: I'd consider reordering  these lines to make it clearer that we query before we switch cursors.
```suggestion
                domain = ProcurementGroup._get_orderpoint_domain()
                self = OrderPoint.with_context(prefetch_fields=False).search(domain)
                 cr = registry(self._cr.dbname).cursor()
                self = self.with_env(self.env(cr=cr))
```

addons/udes_stock/data/scheduler.xml
```diff
@@ -0,0 +1,25 @@
...
+        <record id="stock.ir_cron_scheduler_action" model="ir.cron">
+            <field eval="False" name="active"/>
+        </record>
+
+        <record id="ir_cron_scheduler_action_orderpoints" model="ir.cron">
+            <field name="name">Check order points</field>
+            <field name="model_id" ref="stock.model_stock_warehouse_orderpoint"/>
+            <field name="state">code</field>
+            <field name="code">
+model.check_order_points(True)
```
PeterAlabaster, line 14: are we not porting the `model._cr.commit()` line?

PeterAlabaster (changes requested):
LGTM left a few comments though

addons/udes_stock/data/scheduler.xml
caleb-shelton: See resolved comment above. No longer needed

addons/udes_stock/data/routes.xml
caleb-shelton: fixed

PeterAlabaster (approved):

PR #627, opened 2022-05-05
Story/x1715

armandcela:


addons/udes_security/models/models.py
```diff
@@ -22,6 +22,16 @@ def fields_view_get(self, view_id=None, view_type="form", toolbar=False, submenu
...
         and take into account archiving restrictions for the given user"""
 
+        def _add_confirmation_warning_on_toggle_active(doc):
+            """Add confirm attribute on toggle_active button"""
+            nodes = doc.xpath("//button[@name='toggle_active']")
+            confirm_message = "Are you sure you want to modify this record? " \
+                              "If archived will become unavailable and hidden from searches."
+            for node in nodes:
+                node.attrib["confirm"] = confirm_message
+                setup_modifiers(node, {})
```
PeterAlabaster, line 32: more of a question, maybe a comment is needed to clarify - but what does this line do?

PeterAlabaster (approved):

addons/udes_security/models/models.py
armandcela: TBH found on google it is used to apply modifiers.
[](https://www.odoo.com/forum/help-1/i-unable-to-change-the-properties-of-some-fields-from-within-fields-view-get-method-89215)

PeterAlabaster: ah ok, didn't spot the import as it was already there. fine to leave without comment as the one in odoo/odoo/osv/orm.py has a docstring

LoKKeR96 (approved):
LGTM, leaving thumb up for TL review.

adamrpatrick (approved):

PR #630, opened 2022-05-13
[IMP] udes_stock_picking_batch: Recompute state after action done

LoKKeR96:
This fixes the test for the backordering logic where the batch state
is not being updated after the picking is done.

User-story: x1800

### DIFF
--- a/addons/udes_stock_picking_batch/models/stock_picking.py
+++ b/addons/udes_stock_picking_batch/models/stock_picking.py
@@ -52,11 +52,21 @@ class StockPicking(models.Model):
 
     def action_assign(self):
         """
-        Recompute batch state. In theory this is not necessary
+        Ensure the batch state is recomputed. In theory this is not necessary
         but the constraint on state has not proven to work correctly.
         """
-        super().action_assign()
+        res = super().action_assign()
         self.batch_id._compute_state()
+        return res
+
+    def _action_done(self):
+        """
+        Ensure the batch state is recomputed. In theory this is not necessary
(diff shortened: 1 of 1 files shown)


PeterAlabaster (approved):

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -52,12 +52,20 @@ def _trigger_batch_state_recompute(self):
...
         """
         super().action_assign()
         self.batch_id._compute_state()
 
+    def _action_done(self):
+        """
+        Ensure the batch state is recomputed. In theory this is not necessary
+        but the constraint on state has not proven to work correctly.
+        """
+        super()._action_done()
```
armandcela, line 66:
```suggestion
        res = super()._action_done()
```

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -52,12 +52,20 @@ def _trigger_batch_state_recompute(self):
...
+        self.batch_id._compute_state()
+
```
armandcela, line 68: Add a line: `return res` at the end. The reason is because the core function is returning True at the end and not sure if somewhere is needed the value that returns. I think its a good practice to return it even when the code method doesnt return anything

LoKKeR96: Thanks, I overlooked this.

armandcela (approved):

PR #632, opened 2022-05-16
#1805 Stock - Picking Types handled inconsistently

jrbirch99:
[IMP] udes_stock: Ensure that the prefixes of picking types are consistent with the UDES style and not Odoo's style.

The sequence_code field is not needed in UDES as it is used strictly in the Odoo style sequence (which we do not want). When updating picking_types, if there was sequence_code field in the values to update, Odoo was hardcoded to update the sequences on the picking types to the Odoo prefix style. This resulted in changes to prefixes when changing data on the picking type, such as updating the picking_type with data that set a sequence_code.

Additionally when the warehouse is created or updated with a new name or code, Odoo would update the picking type sequence prefixes.  _get_sequence_values() was were these new picking_type sequences prefixes were generated. This method has been overwritten to prevent this from happening.

Some unit tests were added to ensure that writing to a picking type or updating the warehouse name didn't result in a change to the picking type sequence prefix.

Story/1805

### DIFF
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -1,5 +1,8 @@
-from odoo import fields, models
+import logging
 
+from odoo import fields, models, _, api
+
+_logger = logging.getLogger(__name__)
 
 TARGET_STORAGE_FORMAT_OPTIONS = [
     ("pallet_products", "Pallet of products"),
@@ -11,6 +14,9 @@ TARGET_STORAGE_FORMAT_OPTIONS = [
 
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
+
+    # Overwrite sequence_code as it is only needed for old Odoo prefixes.
+    sequence_code = fields.Char("Sequence_code", required=False)
 
     u_user_scans = fields.Selection(
         [("pallet", "Pallets"), ("package", "Packages"), ("product", "Products")],
@@ -46,9 +52,8 @@ class StockPickingType(models.Model):
     # Enable multi users processing same picking simultaneously
     u_multi_users_enabled = fields.Boolean(
         string="Multi Users Enabled",
-        help="Flag to enable multi users processing same picking simultaneously. "
-             "On validate of picking will be confirmed only what is done by the specific user "
-             "unless all moves have been picked from other users on the validation time.",
+        help="Flag to enable multiple users to work on a picking simultaneously."
+        "On validation of a picking only the moves done by the specific user will be confirmed unless all moves have already been picked by other users",
         default=False,
     )
     u_enable_unpickable_items = fields.Boolean(
@@ -67,3 +72,33 @@ class StockPickingType(models.Model):
         """
         self.ensure_one()
         return self.u_multi_users_enabled
+
+    def write(self, vals):
+        """
+        Purpose: This will remove the sequence code from the list of values to be updated. This is to prevent the sequences
+        for each picking type from being updated with hardcoded values. sequence_code is only relevant to the old odoo prefixes
+        and not the UDES ones, hence a warning message is also displayed.
+        """
+        if "sequence_code" in vals:
+            _logger.warning(
+                _(
+                    "The 'sequence code' field, which has value: (%s), is no longer being used in UDES, this field has been removed from the values to update and will not be updated.",
+                    vals["sequence_code"],
+                )
+            )
+            del vals["sequence_code"]
+        return super(StockPickingType, self).write(vals)
+
+        
+    @api.model
+    def create(self, vals):
+        """
+        Purpose: If picking_type is to be full copy, then set the sequence_id on the record else set the sequence_code to the name 
+        """
+        Stockpickingtype = self.env["stock.picking.type"]
+        stock_picking_type_record = Stockpickingtype.search([("name","=",vals["name"])])
+        if len(stock_picking_type_record) != 0: 
+            vals["sequence_id"] = stock_picking_type_record[0].sequence_id.id
+        else:
+            vals["sequence_code"] = vals["name"]
+        return super(StockPickingType, self).create(vals)
--- a/addons/udes_stock/models/stock_picking_type.py
+++ b/addons/udes_stock/models/stock_picking_type.py
@@ -16,7 +16,7 @@ class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
     # Overwrite sequence_code as it is only needed for old Odoo prefixes.
-    sequence_code = fields.Char("Sequence_code", required=False)
+    sequence_code = fields.Char(required=False)
 
     u_user_scans = fields.Selection(
         [("pallet", "Pallets"), ("package", "Packages"), ("product", "Products")],
@@ -93,7 +93,11 @@ class StockPickingType(models.Model):
     @api.model
     def create(self, vals):
         """
-        Purpose: If picking_type is to be full copy, then set the sequence_id on the record else set the sequence_code to the name 
+        Purpose: When copying the picking type, the sequence id is not copied. This can lead to issues when trying to create the picking type.
+        If the picking type is to be a direct copy then a search will be done for the original record and the sequence will be attached to the
+        new record. If it isn't going to be a direct copy i.e. have a different name, then the sequence code should be set. This can happen when
+        calling .copy(vals) on a record, where vals contains an updated name field. In this case the copied record will have the old Odoo style
+        prefix.  
         """
         Stockpickingtype = self.env["stock.picking.type"]
         stock_picking_type_record = Stockpickingtype.search([("name","=",vals["name"])])
--- a/addons/udes_stock/models/stock_warehouse.py
+++ b/addons/udes_stock/models/stock_warehouse.py
@@ -67,22 +67,17 @@ class StockWarehouse(models.Model):
             raise ValidationError(_("Cannot find picking types for warehouse %s.") % self.name)
 
         return picking_types
-    
+
     def _get_sequence_values(self):
         """
         Purpose: Overwrite the hardcoded sequence values. This will mean that when the warehouse is created the sequences for each picking type
         will not be created with the old Odoo prefix (The sequences should not be created at all). When the warehouse name is updated the
-        sequences for the picking types will not be updated with the old Odoo prefix. 
+        sequences for the picking types will not be updated with the old Odoo prefix.
         """
         return {
-            'in_type_id': {
-            },
-            'out_type_id': {
-            },
-            'pack_type_id': {
-            },
-            'pick_type_id': {
-            },
-            'int_type_id': {
-            },
+            "in_type_id": {},
+            "out_type_id": {},
+            "pack_type_id": {},
+            "pick_type_id": {},
+            "int_type_id": {},
         }
--- a/addons/udes_stock/models/stock_warehouse.py
+++ b/addons/udes_stock/models/stock_warehouse.py
@@ -67,3 +67,22 @@ class StockWarehouse(models.Model):
             raise ValidationError(_("Cannot find picking types for warehouse %s.") % self.name)
 
         return picking_types
+    
+    def _get_sequence_values(self):
+        """
+        Purpose: Overwrite the hardcoded sequence values. This will mean that when the warehouse is created the sequences for each picking type
+        will not be created with the old Odoo prefix (The sequences should not be created at all). When the warehouse name is updated the
+        sequences for the picking types will not be updated with the old Odoo prefix. 
+        """
+        return {
+            'in_type_id': {
+            },
+            'out_type_id': {
+            },
+            'pack_type_id': {
+            },
+            'pick_type_id': {
+            },
+            'int_type_id': {
+            },
+        }
--- a/addons/udes_stock/tests/test_stock_picking_type.py
+++ b/addons/udes_stock/tests/test_stock_picking_type.py
@@ -4,24 +4,27 @@ from . import common
 
 
 class TestStockPickingType(common.BaseUDES):
-    
     def test_writing_to_picking_type_does_reset_prefix_to_Odoo_style(self):
         Pickingtype = self.env["stock.picking.type"]
 
-        goods_out = Pickingtype.search([("name", "=", "Goods Out")])
+        goods_out = self.env.ref("stock.picking_type_out")
         goods_out_prefix = goods_out.sequence_id.prefix
         self.assertEqual(goods_out_prefix, "OUT")
 
         goods_out.write({"name": "Changed goods out", "sequence_code": "out"})
+
+        goods_out = self.env.ref("stock.picking_type_out")
         goods_out_prefix = goods_out.sequence_id.prefix
-        
+
         self.assertEqual(goods_out_prefix, "OUT")
 
         self.create_quant(self.apple.id, self.test_stock_location_01.id, 10)
         products_info = [{"product": self.apple, "uom_qty": 10}]
-        
         picking = self.create_picking(
-            picking_type=goods_out, products_info=products_info, confirm=True, location_dest_id=self.test_received_location_01.id,
+            picking_type=goods_out,
+            products_info=products_info,
+            confirm=True,
+            location_dest_id=self.test_received_location_01.id,
         )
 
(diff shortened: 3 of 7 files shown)


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -12,6 +15,9 @@
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
+    # Overwrite sequence_code as it is only needed for old odoo prefixes.
```
kdwyer, line 18: NIT: we should capitalise "Odoo" when it's being used as the name of the framework.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -47,8 +53,8 @@ class StockPickingType(models.Model):
     u_multi_users_enabled = fields.Boolean(
         string="Multi Users Enabled",
         help="Flag to enable multi users processing same picking simultaneously. "
-             "On validate of picking will be confirmed only what is done by the specific user "
-             "unless all moves have been picked from other users on the validation time.",
+        "On validate of picking will be confirmed only what is done by the specific user "
```
kdwyer, line 56: This isn't very clear.  Perhaps it should read:

> Flag to enable multiple users to work on a picking simultaneously.  On validation of a picking only the moves done by the specific user will be confirmed unless all moves have already been picked by other users.

(I think that's what it's supposed to mean...)

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -67,3 +73,17 @@ def can_handle_multiple_users(self):
...
         return self.u_multi_users_enabled
+
+    def write(self, vals):
+        """
+        Purpose: This will remove the sequence code from the list of values to be updated. This is to prevent the sequences
+        for each picking type from being updated with hardcoded values. sequence_code is only relevant to the old odoo prefixes
+        and not the UDES ones, hence a warning message is also displayed.
+        """
+        if "sequence_code" in vals:
+            _logger.warning(
```
kdwyer, line 84: Does this message need to be internationalised?

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -67,3 +67,22 @@ def get_picking_types(self):
             raise ValidationError(_("Cannot find picking types for warehouse %s.") % self.name)
 
         return picking_types
+    
+    def _get_sequence_values(self):
+        """
+        Purpose: Overwrite the hardcoded sequence values. This will mean that when the warehouse is created the sequences for each picking type
+        will not be created with the old Odoo prefix (The sequences should not be created at all). When the warehouse name is updated it means
+        the sequences for the picking types will not be updated with the old Odoo prefix.
```
kdwyer, line 75: > When the warehouse name is updated it means  the sequences 

This will read just as well without the "it means"

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -67,3 +67,22 @@ def get_picking_types(self):
...
 
         return picking_types
+    
+    def _get_sequence_values(self):
+        """
+        Purpose: Overwrite the hardcoded sequence values. This will mean that when the warehouse is created the sequences for each picking type
+        will not be created with the old Odoo prefix (The sequences should not be created at all). When the warehouse name is updated it means
+        the sequences for the picking types will not be updated with the old Odoo prefix. 
+        """
+        return {
```
kdwyer, line 77: Is black really formatting the empty dict lines like this? :vomiting_face:

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,37 @@ def setUpClass(cls):
...
-    def test01_get_test_picking_types(self):
+    def test_get_test_picking_types(self):
         """Test checks if the test picking types are present"""
         test_picking_types = self.warehouse.get_picking_types().filtered(
             lambda pt: "TEST" in pt.name
         )
         self.assertEqual(len(test_picking_types), 6)
 
-    def test02_get_picking_types(self):
+    def test_get_picking_types(self):
```
kdwyer, line 17: Duplicated method name.  The first method could be called `test_can_find_expected_picking_type`, this one `test_finds_expected_number_of_picking_types`

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,37 @@ def setUpClass(cls):
...
         """Create a new warehouse and look at the default picking types"""
         # Create a new company creates a new warehouse
         test_company = self.create_company("test_company")
         new_warehouse = self.Warehouse.search([("company_id", "=", test_company.id)])
         test_picking_types = new_warehouse.get_picking_types()
         self.assertEqual(len(test_picking_types), 3)
+    
+    def test_check_picking_types_prefix_after_updating_warehouse_name(self):
+        """
+        Test that making changes to the warehouse does not cause the pickingtype prefixes to reset.
```
kdwyer, line 32: s/pickingtype/picking type/

kdwyer (changes requested):
I like the effort that's gone into documenting this PR, both in the commit message and the docstrings :+1:

addons/udes_stock/tests/test_stock_warehouse.py
peterpratt13: and: to reset -> to be reset

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,37 @@ def setUpClass(cls):
...
-    def test03_get_default_picking_types_new_warehouse(self):
+    def test_get_default_picking_types_new_warehouse(self):
         """Create a new warehouse and look at the default picking types"""
         # Create a new company creates a new warehouse
         test_company = self.create_company("test_company")
         new_warehouse = self.Warehouse.search([("company_id", "=", test_company.id)])
         test_picking_types = new_warehouse.get_picking_types()
         self.assertEqual(len(test_picking_types), 3)
+    
+    def test_check_picking_types_prefix_after_updating_warehouse_name(self):
```
peterpratt13, line 30: nit: This test doesn't update warehouse name, it changes  the code.

addons/udes_stock/tests/test_stock_picking_type.py
```diff
@@ -0,0 +1,14 @@
+from . import common
+
+
+class TestStockPickingCommon(common.BaseUDES):
+    def test_writing_to_picking_type(self):
```
peterpratt13, line 5: Can we make this clearer what it is testing please. 
Test writing on picing type does not reset the picking type prefix?

addons/udes_stock/tests/test_stock_picking_type.py
```diff
@@ -0,0 +1,14 @@
...
+        Pickingtype = self.env["stock.picking.type"]
+
+        goods_out = Pickingtype.search([("name", "=", "Goods Out")])
+        goods_out_prefix = goods_out.sequence_id.prefix
+        self.assertEqual(goods_out_prefix, "OUT")
+
+        goods_out.write({"name": "Changed goods out", "sequence_code": "out"})
+        goods_out_prefix = goods_out.sequence_id.prefix
+        self.assertEqual(goods_out_prefix, "OUT")
```
peterpratt13, line 14: `Can` we check the full name of the picking? WH/OUT/ 
because we only test the prefiexs.

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,37 @@ def setUpClass(cls):
...
+        """ 
+        Pickingtype = self.env["stock.picking.type"]
+        
+        goods_in_prefix = Pickingtype.search([("name", "=", "Goods In")]).sequence_id.prefix
+        self.assertEqual(goods_in_prefix, "IN")
+        
+        self.warehouse.write({"code":"NEW_WAREHOUSE_CODE"})
+
+        goods_in_prefix = Pickingtype.search([("name", "=", "Goods In")]).sequence_id.prefix
+        self.assertEqual(goods_in_prefix, "IN")
```
peterpratt13, line 42: Can we check the full name again, so that we can see the NEW_WAREHOUSE_CODE/IN ?

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,37 @@ def setUpClass(cls):
...
+
```
peterpratt13, line 43: Might be nice to have a test with  a new warehouse in as well

peterpratt13:
LGTM - super minor comments

peterpratt13:
Do the tests fail due to lack of Odoo commit bump, or is there something else?

addons/udes_stock/models/stock_picking_type.py
jrbirch99: Some instance of the logger.warning message are internationalised and some are not. I am not sure if there is a rule for internationalising strings it that I am not aware of, but I will internationalise it here.

addons/udes_stock/tests/test_stock_warehouse.py
kdwyer: Yes - the notation is from tools like [sed](https://en.wikipedia.org/wiki/Sed), it means

> substitute pickingtype with "picking type"

addons/udes_stock/tests/test_stock_picking_type.py
jrbirch99: Yep, I can do this for a picking so we can see the full name. However the warehouse code won't be included in the name. WH/OUT/XXXX1 is the Odoo style of name. The UDES style name is OUTXXXX1. 

Creating multi-warehouse sequence names will be handled in this story https://taiga.unipart.digital/project/udes-x/us/1898.

addons/udes_stock/tests/test_stock_warehouse.py
jrbirch99: Upon creating a new warehouse, Odoo will automatically create picking types with the Odoo style prefix WH/OUT and sequences WH/OUT/XXXX1. We set our own sequences and prefixes in data, in UDES14. For example we set picking type goods in to have prefix IN which results in a sequence INXXXX1 for example.

As it will never be a warehouse user creating a picking type and always a support user or dev, it should not matter if these picking types are created with the wrong sequence and prefix as we will always set our own.

Hence the test of creating a new warehouse would mean that I would have to reset all the sequences and do a similar test to what I have already done, which feels a bit redundant to me.

addons/udes_stock/tests/test_stock_warehouse.py
jrbirch99: okay cool :)

addons/udes_stock/tests/test_stock_warehouse.py
jrbirch99: Same as comment above.

jrbirch99:
> Do the tests fail due to lack of Odoo commit bump, or is there something else?

Yep it is because of the Odoo changes, I ran them locally and they run okay :) I will attach a file with the unit tests passing to the story. When I run the changes from Odoo on the UDES_14 branch, I get the same failures.

addons/udes_stock/tests/test_stock_picking_type.py
peterpratt13: Can we use a regex then? The only thing is because, before the issue was the prefix often existed in the picking name, but not as expected.

addons/udes_stock/models/stock_picking_type.py
```diff
@@ -12,6 +15,9 @@
 class StockPickingType(models.Model):
     _inherit = "stock.picking.type"
 
+    # Overwrite sequence_code as it is only needed for old Odoo prefixes.
+    sequence_code = fields.Char("Sequence_code", required=False)
```
PeterC10, line 19: Q: Do we need to set `Sequence_code`? Should just be able to do `required=False`.

It might be worth considering adding `groups="base.no_one"` - to prevent this field from being accessible by normal users, but not really a big deal.

addons/udes_stock/models/stock_warehouse.py
PeterC10: How would you format it out of interest?

(Not for the story as black is king, but just curious!)

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +7,36 @@ def setUpClass(cls):
...
         test_picking_types = new_warehouse.get_picking_types()
         self.assertEqual(len(test_picking_types), 3)
+
+    def test_check_picking_types_prefix_after_updating_warehouse_code(self):
+        """
+        Test that making changes to the warehouse does not cause the pickingtype prefixes to be reset.
+        """
+        Pickingtype = self.env["stock.picking.type"]
+
+        goods_in_prefix = Pickingtype.search([("name", "=", "Goods In")]).sequence_id.prefix
```
PeterC10, line 36: Suggestion: Reference the picking type directly with `self.env.ref` - on the tiny chance that it gets renamed... :)

addons/udes_stock/models/stock_picking_type.py
jrbirch99: This should be changed now

addons/udes_stock/tests/test_stock_picking_type.py
jrbirch99: yep should be done now

jrbirch99:
When installing udes_rockhopper and running the udes_stock tests, some of the units tests failed. This was due to .copy() when copying a picking_type in udes_stock/tests/common.py. Essentially it would not copy the sequence_id field, meaning that when it got to the create method of the picking_type, it would try to create a new ir_sequence, but as the sequence_code is not set on the picking_type it was copying from, it would fail. To get around this I extended the create method such that, if you are creating an exact copy it will have the same sequence as the original, but if you are calling copy(vals) and updating the name in vals. Then it will create an odoo style prefix.

addons/udes_stock/models/stock_warehouse.py
kdwyer: It turns out I'd format it the same as black would ;)

```diff
black --diff test.py 
--- test.py	2022-05-17 11:21:23.511293 +0000
+++ test.py	2022-05-17 12:48:25.233572 +0000
@@ -1,11 +1,7 @@
 def foo():
     return {
-        'aaaa_type_id': {
-        },
-        'bbbb_type_id': {
-        },
-        'cccc_type_id': {
-        },
-        'dddd_type_id': {
-        },
+        "aaaa_type_id": {},
+        "bbbb_type_id": {},
+        "cccc_type_id": {},
+        "dddd_type_id": {},
     }
would reformat test.py
```

addons/udes_stock/tests/test_stock_picking_type.py
```diff
@@ -0,0 +1,28 @@
...
+        self.create_quant(self.apple.id, self.test_stock_location_01.id, 10)
+        products_info = [{"product": self.apple, "uom_qty": 10}]
+        picking = self.create_picking(
+            picking_type=goods_out,
+            products_info=products_info,
+            confirm=True,
+            location_dest_id=self.test_received_location_01.id,
+        )
+
+        self.assertTrue(bool(re.search(r"OUT\d{5}", picking.name)))
```
kdwyer, line 28: [re.fullmatch](https://docs.python.org/3/library/re.html#re.fullmatch) might be better than `search`, assuming that we don't want to match "SHOUT00001ABC" or suchlike.

addons/udes_stock/tests/test_stock_warehouse.py
```diff
@@ -7,22 +9,46 @@ def setUpClass(cls):
...
+        """
+        Test that making changes to the warehouse does not cause the picking type prefixes to be reset.
+        """
+        goods_in = self.env.ref("stock.picking_type_in")
+        goods_in_prefix = goods_in.sequence_id.prefix
+        self.assertEqual(goods_in_prefix, "IN")
+
+        self.warehouse.write({"code": "NEW_WAREHOUSE_CODE"})
+
+        goods_in_prefix = self.env.ref("stock.picking_type_in").sequence_id.prefix
```
PeterC10, line 42: This should be `goods_in_prefix = goods_in.sequence_id.prefix` like on line 37.

PeterC10 (approved):

PR #633, opened 2022-05-17
[IMP] udes_stock,udes_stock_picking_batch: Refactor Buttons

LoKKeR96:
Refactor buttons that are not necessary and rename cancel pickings
button to make it easy to distinguish from discard changes.

User-story: x1804

### DIFF
--- a/addons/udes_stock/views/stock_picking.xml
+++ b/addons/udes_stock/views/stock_picking.xml
@@ -101,6 +101,17 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
 
+            <!-- Hide Return button for all -->
+            <xpath expr="//button[@name='%(stock.act_stock_return_picking)d']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+
+            <!-- Rename to avoid confusion between discard and cancel -->
+            <xpath expr="//button[@name='action_cancel']" position="attributes">
+                <attribute name="string">Cancel Pick</attribute>
+                <attribute name="attrs">{'invisible': [('state', 'in', ('draft', 'done', 'cancel'))]}</attribute>
+            </xpath>
+
         </field>
     </record>
 
--- a/addons/udes_stock_picking_batch/views/stock_picking_batch.xml
+++ b/addons/udes_stock_picking_batch/views/stock_picking_batch.xml
@@ -25,6 +25,16 @@
           <field name="priority"/>
           <field name="u_last_reserved_pallet_name" groups="udes_security.group_trusted_user"/>
         </xpath>
+
+        <!-- Hide Put in Pack button -->
+        <xpath expr="//button[@name='action_put_in_pack']" position="attributes">
+          <attribute name="invisible">1</attribute>
+        </xpath>
+
+        <!-- Hide Validate button -->
+        <xpath expr="//button[@name='action_done']" position="attributes">
+          <attribute name="invisible">1</attribute>
+        </xpath>
       </field>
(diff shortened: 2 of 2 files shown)


addons/udes_stock/views/stock_picking.xml
```diff
@@ -101,6 +101,15 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
 
+            <!-- Remove Return button for all -->
+            <xpath expr="//button[@name='%(stock.act_stock_return_picking)d']" position="replace"/>
+
+            <!-- Rename to avoid confusion between discard and cancel -->
+            <xpath expr="//button[@name='action_cancel']" position="attributes">
+                <attribute name="string">Cancel Pick</attribute>
+                <attribute name="attrs">{'invisible': [('state', 'not in', ('assigned', 'confirmed', 'partially_available', 'waiting'))]}</attribute>
```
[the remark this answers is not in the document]
LoKKeR96: Everyone is tripping on this double negation. It's actually invisible when the expression on the right ("state","not in", ...) is True. So it's the opposite. It's invisible when the state is not in one of those.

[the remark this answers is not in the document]
jrbirch99: right, so surely we can change it to "in" instead and just put "draft" ? As it stands in your code I think you would also have to add "done" and "cancel" otherwise the buttons will appear in those states too.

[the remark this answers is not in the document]
LoKKeR96: Yes, we can switch the logic around and make it more readable if you are suggesting that.

addons/udes_stock/views/stock_picking.xml
```diff
@@ -101,6 +101,15 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
 
+            <!-- Remove Return button for all -->
+            <xpath expr="//button[@name='%(stock.act_stock_return_picking)d']" position="replace"/>
```
armandcela, line 105: I would suggest to make it invisible without condition than to replace.
```suggestion
            <xpath expr="//button[@name='%(stock.act_stock_return_picking)d']" position="attributes">
                <attribute name="invisbile">1</attribute>
            </xpath>
```

PeterC10, line 105: Can we make this invisible instead of replacing? (Same for others)

It's a bit safer, incase another view references it in future.

`<attribute name="invisible">1</attribute>` should do it.

PeterC10: Armand beat me to it. :)

LoKKeR96: Done

addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -25,6 +25,12 @@
           <field name="priority"/>
           <field name="u_last_reserved_pallet_name" groups="udes_security.group_trusted_user"/>
         </xpath>
+
+        <!-- Remove Put in Pack button -->
+        <xpath expr="//button[@name='action_put_in_pack']" position="replace"/>
```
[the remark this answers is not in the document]
LoKKeR96: Done

addons/udes_stock/views/stock_picking.xml
LoKKeR96: Done

armandcela (approved):

PeterC10 (approved):

PR #634, opened 2022-05-17
Story/1809

kdwyer:


kdwyer:
This PR:
1.  Removes mutation of Location._get_info_field_names in tests
2. Makes the default behaviour more like that in UDES 11: relational fields always return the result of their get_info method
3. Tries to simplify and clarify the code.

It's probably best reviewed as a single blob, but I've haven't yet squashed the commits as they document my thought process while I was working on this.

PeterAlabaster (approved):

armandcela (approved):

PR #635, opened 2022-05-18
Story/x1177

ws-mps:


addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +740,268 @@ def raise_stock_investigation(self, reason, quants, location):
...
+        result_parent_package_name,
+        target_storage_format=None,
+    ):
+        """
+        Compute result_package and result_parent_package based on the u_target_storage_format of the
+        picking type (if `target_storage_format` is not supplied) + the input parameters.
+        """
+        Package = self.env["stock.quant.package"]
+
+        result_parent_package = None
```
kdwyer, line 807: Should some of this really be on the package model?  Picking should tell Package what format it wants, but Package decides how to create an instance of itself?

addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +740,268 @@ def raise_stock_investigation(self, reason, quants, location):
...
+
+        elif target_storage_format == "product":
+            # Error when trying to mark_as_done a full package or setting result package
+            # when result storage format is products
+            if result_package_name:
+                raise ValidationError(_("Invalid parameters for products target storage format."))
+
+        return result_package, result_parent_package
+
+    def picked(
```
kdwyer, line 883: Should we give this a more expressive name, like `mark_picked_lines`?  `some_picking.picked()` doesn't tell the reader much about what might be happening.

addons/udes_stock/models/stock_picking.py
ws-mps: We have 2 packages, you mean one of them or just Package in generic?

addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +740,268 @@ def raise_stock_investigation(self, reason, quants, location):
...
+                raise ValidationError(
+                    _("Invalid parameters for target storage format, expecting result package.")
+                )
+
+        elif target_storage_format == "pallet_products":
+            if result_package_name:
+                # Moving stock into a pallet of products, result_package
+                # might be new pallet id
+                result_package = Package.get_or_create(result_package_name, create=True)
+                print("****** here we are")
```
kdwyer, line 862: :eyes:

addons/udes_stock/models/stock_picking.py
kdwyer: In generic - ideally logic about creating packages would be the responsibility of Package rather than Picking.  This method is about 80 lines long, but `self` is only referenced three times - that suggests to me some of the logic might belong elsewhere :)

addons/udes_get_info/models/__init__.py
```diff
@@ -6,3 +6,4 @@
 from . import stock_move_line  # noqa: F401
 from . import stock_move  # noqa: F401
 from . import stock_picking  # noqa: F401
+from . import stock_picking_batch  # noqa: F401
```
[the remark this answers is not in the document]
PeterAlabaster: surpresses pep8 warning `module imported but unused`

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,132 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+            funcs.append(lambda ml: ml.location_id == location)
+        if lot_name:
+            funcs.append(lambda ml: (lot_name == lot_name or ml.lot_id.name == lot_name))
+        func = lambda ml: all(f(ml) for f in funcs)
+
+        # Find move lines
+        mls = self.filtered(func)
+        mls_fulfill, new_ml, uom_qty = mls.move_lines_for_qty(uom_qty)
+        if uom_qty > 0:
+            self.picking_id.add_unexpected_parts()
```
armandcela, line 275: It looks like add_unexpected_parts should be outside this method, as the method name is to find move lines.
I think add_unexpected_parts should be called where is called _find_move_lines or changed the docstring to mention this

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,132 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+
+        vals = {}
+        if result_package:
+            vals["result_package_id"] = result_package.id
+        if location_dest:
+            vals["location_dest_id"] = location_dest.id
+        mls = self.get_lines_incomplete()
+        incoming = mls.picking_type_id.code == "incoming"
+        res = defaultdict(dict)
+        if product_ids:
```
armandcela, line 294: I would suggest to delete if product_ids and loop directly through them.
After the loop 
at the beginning of the line smth like:

```
if not product_ids:
    if package:
        pack_mls = package._get_current_move_lines() ....
   else:
         res[mls] = vals
```

The reason why i am suggesting this is because you gain some characters on the product_ids loop and i think looks better the code

addons/udes_stock/models/stock_move_line.py
ws-mps: We haven't implemented this feature yet, so I should have probably returned the new created move line and add it to `mls_fulfill`, but I will review it thanks.

addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +740,268 @@ def raise_stock_investigation(self, reason, quants, location):
...
+                        _(
+                            "Cannot validate %s because there are move lines todo and "
+                            "backorder not allowed"
+                        )
+                        % self.log_name()
+                    )
+                else:
+                    requires_backorder = True
+            if requires_backorder:
+                new_picking = self.backorder_move_lines(mls_to_keep=mls_to_keep)
```
peterpratt13, line 997: This is a drive by comment and not related specifically to this story really but, 
What happens if:
pick has 3 move lines apples, bananas and cherries
user 1 scans apples
user 2 scans bananas

User 1 then drops bananas - it moves the picking bananas and cherries into a backorder. 
Will the user 1 work at the move line level, so validate according to the move line, i.e when they try to drop the move line, the pass the move line id, the picking id is found and then they drop so all is well, or does it use the picking id, which now does not have the banana move lines in it?
Also in this scenario, we may need to consider the user assigned picking id behaviour - essentially the user has been re-assigned, but the start time should remain the same.

Like i say out of scope here - just curious.

addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +740,268 @@ def raise_stock_investigation(self, reason, quants, location):
...
+        result_parent_package = None
+
+        if target_storage_format is None:
+            target_storage_format = self.picking_type_id.u_target_storage_format
+        scan_parent_package_end = self.picking_type_id.u_scan_parent_package_end
+
+        if target_storage_format == "pallet_packages":
+
+            # CASE A: both package names given
+            if result_package_name and result_parent_package_name:
```
peterpratt13, line 816: We could with newer verions for storage format configs use cases, and then have this consistent throughout but not sure the python version used so feel free to ignore.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,132 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+        mls = self.filtered(func)
+        mls_fulfill, new_ml, uom_qty = mls.move_lines_for_qty(uom_qty)
+        if uom_qty > 0:
+            self.picking_id.add_unexpected_parts()
+        return mls_fulfill, new_ml
+
+    def prepare_move_lines(
+        self, product_ids=None, package=None, location=None, result_package=None, location_dest=None
+    ):
+        """ Prepare incomplete move lines in self into groups of mls to be updated with
```
peterpratt13, line 281: nit: Could we add what we expect from product_ids, if not none {"barcode": ..., "qty":....}

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,132 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+                        _("Too many lots provided.")
+                    )
+                qty_done = quantity
+                if product_tracking != "none":
+                    quantity_fulfilled = 0
+                    if is_serial:
+                        qty_done = 1
+                    for lot_name in lot_names:
+                        lot_name_val = lot_name
+                        # If it is incoming we don't need the match the lot
```
peterpratt13, line 318: nit: the -> to

peterpratt13:
LGTM - mine are mainly questions on implications with later stories or super minor so approved it anyway

peterpratt13 (approved):

addons/udes_stock/models/stock_picking.py
ws-mps: When multiple users working on the same picking is still not supported, but the idea is to have some code that when the picking is validated and multi users is enabled it should find the backorder in state assigned. We could add that code instead of `assert_valid_state()`.

addons/udes_stock/models/stock_picking.py
ws-mps: At the moment we have ported the same funciton we have in 11. But in future work we could have a class, like in suggested locations, that handles what should happen for a given target storage format.

addons/udes_stock/models/stock_move_line.py
ws-mps: Have added a note so we review if this is the best place to have it when we implement add_unexpected_parts

armandcela (approved):

addons/udes_get_info/models/stock_picking_batch.py
```diff
@@ -0,0 +1,20 @@
+"""
+Stock picking batch model.
+
```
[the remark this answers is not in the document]
ws-mps: All the other files have it, so I will leave it.

kdwyer (approved):

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,139 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+
+        # Find move lines
+        mls = self.filtered(func)
+        mls_fulfill, new_ml, uom_qty = mls.move_lines_for_qty(uom_qty)
+        if uom_qty > 0:
+            # Note: when implementing add_unexpected_parts() review if this is the best place
+            mls_fulfill |= self.picking_id.add_unexpected_parts()
+        return mls_fulfill, new_ml
+
+    def prepare_move_lines(
```
kdwyer, line 279: Bit of a nit, but if `self` is an instance of `StockMoveLine` we could just call this method "prepare" - likewise in other method names that duplicate the class name (assuming that they aren't inherited).

addons/udes_stock_packaging/README.md
```diff
@@ -13,16 +13,25 @@
...
+
+| Helpers                         | Description                                                                                                                                                                                                                                                                                                                                         |
+|---------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| _prepare_result_packages        | Prepare result packages according to a given target storage format and the parameters provided                                                                                                                                                                                                                                                      |
```
kdwyer, line 37: Nit: we should remove the leading underscore as it's part of the public API

addons/udes_stock/models/stock_move_line.py
```diff
@@ -252,3 +252,139 @@ def get_move_lines_ordered_by(self, domain, aux_domain=None, order="id"):
...
+                prod_mls, new_ml = mls._find_move_lines(
+                    qty_done, product, package, None, location
+                )
+                prod_dict = {"qty_done": qty_done}
+                prod_dict.update(vals)
+                res[prod_mls] = prod_dict
+                mls -= prod_mls
+                if new_ml:
+                    mls |= new_ml
+        return res
```
kdwyer, line 368: I've recently seen it suggested that once a `defaultdict` has been populated, it should be converted to a standard `dict` so that it isn't possible to accidentally add more keys.  So we could return `dict(res)` here, assuming the caller isn't depending on the `defaultdict` features.  If you think that kind of defensive programming is worth it.

PeterC10 (approved):

PR #636, opened 2022-05-18
Story/1664

armandcela:


addons/udes_common/models/models.py
```diff
@@ -134,3 +135,43 @@ def selection_display_name(self, selection_field_name):
...
+        doc = etree.XML(res["arch"])
+        if self._context.get('view_all_fields'):
+            # Hide the view more button
+            doc = etree.XML(res["arch"])
+            for node in doc.xpath("//button[@name='view_more_info']"):
+                node.getparent().remove(node)
+        else:
+            hidden_fields_from_form_view = tools.hidden_fields_from_form_view.get(self._name, False)
+            if not hidden_fields_from_form_view:
+                return res
```
LoKKeR96, line 155: Nit: I always wondered if a function should have one exit point or not, I cannot argue which is better except that having one exit point seems a little cleaner. Are you returning res here to avoid having multiple if statements?

I would suggest simplifying it as follows:

if hidden_fields_from_form_view and view_type == "form":
    for field_name in hidden_fields_from_form_view:
    ...

armandcela: Sometimes where we need to get out without running extra lines we may have more than one returns in the methods.
I think still is the case because at the end is this line which i didnt want to repeat more than once 
`res["arch"] = etree.tostring(doc)` i thought is better to return res in case there wont be any changes

LoKKeR96:
LGTM, the same comments apply on the UDES11 MR.

addons/udes_common/models/models.py
armandcela: I think that i responded in a logic way the reason why i am returning , just not to execute any further code which is after and unnecessary to run. 
You are saying that you think should be only one return which i agree at some point but we are using in many cases in our custom modules and in core modules more than one return in a method, soi hope you should be happy to resolve this thread without any change.

LoKKeR96: Sounds good, I wanted to spark a discussion more than forcing my nit 😁

addons/udes_stock/views/stock_picking.xml
```diff
@@ -101,6 +101,11 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
 
+            <!-- Add button to show all fields, by default hidden fields will not be shown -->
+            <xpath expr="//button[@name='action_confirm']" position="before">
+                <button name="view_more_info" string="View More" type="object" class="oe_highlight"/>
```
PeterC10, line 106: Can we call this `Detailed View`? Think it reads a bit better but might be something for product. :)

Also I suggest removing `class="oe_highlight"` - it is a difficult one to judge but it isn't a button that affects the flow of any model.

armandcela:
@PeterC10 As discussed i will investigate little further that when we show the fields clicking on them to open a full view not  a pop up like wizard.

addons/udes_common/tools/helpers.py
```diff
@@ -0,0 +1,15 @@
+"""Helpers"""
+
+# Fields that will be hidden from the form views if by default will not be needed to be shown.
+# In order to extend should be added the model like a key if doesnt exist yet, adding the fields
+# needed to be hidden and to show a button View More should be added to the form view.
+# Fields that are added in helpers but not in models will not raise any error just skip them, this
+# is done in order not to add dependencies with modules where fields are defined.
+hidden_fields_from_form_view = {
```
PeterC10, line 8: Could we look into setting the value of the attribute on the relevant model directly?

So by default it could be an empty list.

Then in `udes_stock` we can set this for `stock.picking`.

And for hiding the fields, `self.hidden_fields_from_form_view` should do the trick I think.

armandcela: Liked this suggestion :)

addons/udes_common/models/models.py
```diff
@@ -134,3 +135,43 @@ def selection_display_name(self, selection_field_name):
     # Odoo has a nice function that handles all the weird ways you can define selections!
     selection_values_dict = dict(selection_field._description_selection(self.env))
     return selection_values_dict.get(self[selection_field_name], False)
+
+
+class Base(models.AbstractModel):
+    _inherit = 'base'
```
PeterC10, line 141: Needs black formatting, happy for it to be done at the end though.

addons/udes_common/models/models.py
```diff
@@ -134,3 +135,43 @@ def selection_display_name(self, selection_field_name):
...
+
+    @api.model
+    def fields_view_get(self, view_id=None, view_type='form', toolbar=False, submenu=False):
```
PeterC10, line 144: Can we add doc-strings to all the methods?

addons/udes_common/tools/helpers.py
```diff
@@ -0,0 +1,15 @@
+"""Helpers"""
+
+# Fields that will be hidden from the form views if by default will not be needed to be shown.
```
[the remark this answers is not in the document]
armandcela: From Peter C suggestions helpers was not needed any more as i moved in specific classes the fields like a parameter the fields that need to be hidden.

Made and pushed changes in both MR-s

addons/udes_stock/views/stock_picking.xml
armandcela: Made the changes in both MR-s

addons/udes_common/tools/helpers.py
armandcela: Made the suggested changes in both MR-s

addons/udes_common/models/models.py
armandcela: Made the changes

addons/udes_common/models/models.py
armandcela: Added the docstring in this method, the others i think already had

addons/udes_stock/views/stock_picking.xml
```diff
@@ -111,6 +111,43 @@
...
+            
+            <!-- Add button to show all fields, by default hidden fields will not be shown -->
+            <xpath expr="//button[@name='action_confirm']" position="before">
+                <button name="action_detailed_view" string="Detailed View" type="object"/>
+            </xpath>
+
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button class="oe_stat_button"
+                        name="open_first_pickings"
+                        icon="fa-building-o"
```
PeterC10, line 123: This is always the toughest part even though it is minor. :)

The icon doesn't match well with pickings imo, could we try `fa-exchange` - this icon is used for Product Moves on the product view, and pickings essentially mean products are moved.

Could be something to discuss with product, I believe we should have access to most of these icons for v14 at least: https://fontawesome.com/v4/icons/

addons/udes_common/models/models.py
```diff
@@ -134,3 +135,46 @@ def selection_display_name(self, selection_field_name):
...
     selection_values_dict = dict(selection_field._description_selection(self.env))
     return selection_values_dict.get(self[selection_field_name], False)
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    # Adding a default empty list for base model, it will be override when needed for
+    # specific models.
+    HiddenFieldsFromFormView = []
```
PeterC10, line 145: Suggestion: Not related to your recent changes, but this variable name is a bit long and doesn't indicate that the fields are only hidden by default (although the comment does explain this of course).

Maybe `DetailedFormViewFields` ?

Not a huge deal at all, just something to consider. :)

addons/udes_stock/models/stock_picking.py
```diff
@@ -739,3 +775,61 @@ def raise_stock_investigation(self, reason, quants, location):
...
+                "views": [(form_view.id, "form")],
+                "res_model": "stock.picking",
+                "view_id": form_view.id,
+                "res_id": picking_ids[0],
+                "context": {"view_all_fields": False},
+            }
+        else:
+            action = {
+                "type": "ir.actions.act_window",
+                "name": _(f"{related_pickings_description}"),
```
PeterC10, line 804: f-string doesn't work with translation

addons/udes_stock/views/stock_picking.xml
armandcela: fa-exchange doesnt look bad, if you dont like it than will speak with PO, but i think is ok

PeterC10 (approved):
LGTM

PR #637, opened 2022-05-18
Story/x1664

armandcela:


LoKKeR96:
LGTM, I have the same comments as per other MR on 14.0 branch. Will thumb up once both MRs are updated or discussions are resolved.

LoKKeR96 (approved):
A much better approach, LGTM, need a TL review for a thumb up now.

PeterC10 (approved):
LGTM

addons/udes_stock/models/stock_picking.py
```diff
@@ -37,6 +37,14 @@ class StockPicking(models.Model):
 
     _order = "priority desc, scheduled_date asc, sequence asc, id asc"
 
+    # Adding fields to be hidden by default from form view.
+    DetailedFormViewFields = [
+        "u_first_picking_ids",
+        "u_prev_picking_ids",
+        "u_next_picking_ids",
+        "u_created_back_orders",
+    ]
```
PeterC10, line 46: Sorry, something I missed before... can we add the `_qty` fields here? Otherwise they will be computed even in simple view.

PR #638, opened 2022-05-19
[IMP] udes_stock: Ensure correct warehouse for picking types creation

LoKKeR96:
User-story: x1807


addons/udes_stock/models/stock_picking_type.py
```diff
@@ -101,7 +100,7 @@ def create(self, vals):
         """
         Stockpickingtype = self.env["stock.picking.type"]
         stock_picking_type_record = Stockpickingtype.search([("name","=",vals["name"]), ("company_id", "=", vals.get("company_id"))])
-        if len(stock_picking_type_record) != 0 and not vals.get("sequence_id"): 
+        if len(stock_picking_type_record) != 0 and not vals.get("sequence_id"):
```
kdwyer, line 103: The changes to this file seem to be only removing whitespace?  In that case a better message might be 
```
[FMT] udes_stock: remove unnecessary whitespace
```

LoKKeR96: Will drop the commit, I didn’t notice that I still had it lying around. It had different use before.

kdwyer (approved):

PeterC10 (approved):

PR #641, opened 2022-05-24
Story/x1663

LoKKeR96:


addons/udes_stock/models/stock_picking.py
```diff
@@ -930,3 +930,18 @@ def open_back_orders(self):
...
+        Generate the domain for searching pickings that use a package.
+        Useful as this can be overridden in other modules if functionality is expanded.
+
+        :args:
+            - package: a stock.quant.package object
+        """
+        return [
+            "|",
+            ("move_line_ids.package_id", "=", package.id),
+            ("move_line_ids.result_package_id", "=", package.id),
```
armandcela, line 946: Question: Does result_package_id exist in udes14?

LoKKeR96: On stock move line model, yes. Look at Odoo stock module,  stock.move model.

adamrpatrick (approved):

PR #642, opened 2022-05-26
Story/1712 Porting storage_format

armandcela:


addons/udes_stock/models/stock_picking.py
```diff
@@ -524,18 +524,23 @@ def create_picking(
         confirm=False,
         assign=False,
         create_batch=False,
+        product_quantity=None,
```
armandcela, line 527: Need to remove this line, i think have been forgotten when i was testing some issues that were picking all stock quants not the scanned quantities. 
But made a fix when generating products_info so products_info have the needed information.

addons/udes_stock/models/stock_quant.py
```diff
@@ -49,12 +49,15 @@ def get_quantity(self):
...
-    def get_quantities_by_key(self, get_key=lambda q: q.product_id, only_available=False):
+    def get_quantities_by_key(
+            self, get_key=lambda q: q.product_id, only_available=False, product_quantity=None
+    ):
         """Returns a dictionary with the total quantity per product, mapped by product_id.
         :kwargs:
             - only_available: Boolean
                 Whether to include those reserved or not in the grouping
             - get_key: a callable which takes a quant and returns the key
+            - product_quantity: Float Needed not to get all available products
```
LoKKeR96, line 60: Needed not? Not sure I follow

addons/udes_stock/models/stock_quant.py
```diff
@@ -64,11 +67,17 @@ def get_quantities_by_key(self, get_key=lambda q: q.product_id, only_available=F
             value = quant.quantity
             if only_available:
                 value -= quant.reserved_quantity
-            products[get_key(quant)] += value
+            if product_quantity is not None and product_quantity < value:
```
LoKKeR96, line 70: I think it's safe to do `if product_quantity and product_quantity < value:` since if product_quantity is 0 then it will be false and if it is None it will be false as well.

addons/udes_stock/models/stock_quant.py
armandcela: Not needed at least for now , because needed information is on products_info

addons/udes_stock/models/stock_quant.py
armandcela: Removing after the call had with Miquel, will not added the optional parameter and will let like it was. Will handle on controller the change needed

addons/udes_stock/models/stock_quant.py
LoKKeR96: Can you explain the changes here? I'm not sure I follow :/ Maybe we can have a call.

ws-mps (approved):

LoKKeR96 (approved):
Happy but I would want test.

PR #643, opened 2022-05-27
[FIX] udes_stock: Optimize u_next_picking_ids/u_prev_picking_ids

PeterAlabaster:
Use SQL in favour of .mapped() and going via the ORM, which results
in costly joins/orderbys.

User-story: 1665

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -281,18 +281,45 @@ class StockPicking(models.Model):
     )
     def _compute_related_picking_ids(self):
         Picking = self.env["stock.picking"]
-        for picking in self:
-            if picking.id:
-                back_orders = Picking.get_pickings(backorder_id=picking.id)
-                picking.u_created_back_orders = back_orders
-                picking.u_back_orders_qty = len(back_orders)
-
-            next_pickings = picking.mapped("move_lines.move_dest_ids.picking_id")
-            prev_pickings = picking.mapped("move_lines.move_orig_ids.picking_id")
-            picking.u_prev_picking_ids = prev_pickings
-            picking.u_prev_pickings_qty = len(prev_pickings)
-            picking.u_next_picking_ids = next_pickings
-            picking.u_next_pickings_qty = len(next_pickings)
+        if self.ids:
+            # To avoid messy ORM joins, get next/prev picking ids
+            # with a direct query for all picks in self
+            self.env.cr.execute(
+                """
+                SELECT
+                    sp.id,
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -281,18 +281,45 @@ def _compute_picking_packages(self):
...
+                    LEFT JOIN stock_move_move_rel orig_move_rel ON sm.id = orig_move_rel.move_dest_id
+                    LEFT JOIN stock_move orig_move ON orig_move.id = orig_move_rel.move_orig_id
+                    LEFT JOIN stock_move dest_move ON dest_move.id = dest_move_rel.move_dest_id
+                    LEFT JOIN stock_picking next_pick ON dest_move.picking_id = next_pick.id
+                    LEFT JOIN stock_picking prev_pick ON orig_move.picking_id = prev_pick.id
+                WHERE
+                    sp.id IN %s
+                GROUP BY sp.id
+            """,
+                (tuple(self.ids),),
```
armandcela, line 308: If length is 1 i think this will raise sql syntax error:

u think should be an extra check of the length of self.ids
'if len(self.ids) == 1:
   pickings = "(f{self.ids[0]})"
else:
   pickings = tuple(self.ids)'

And replace on query with pickings variable. Have a look though if i am right

kdwyer (approved):

addons/udes_stock/models/stock_picking.py
PeterC10: `self.ids` will always return a list, so it should be fine I would have thought.

But yes worth checking out to be safe... speeding up is good, but not so good when it comes with an exception. :grin:

PeterC10: Ah I see. Well if we load a picking in the form view then `self` would only have one record anyway right?

@caleb-shelton please check this, but hopefully fine. :)

caleb-shelton: It works fine with self having one record. Also the SQL is checking 'IN'

armandcela (approved):

PeterC10:
Can we do something similar for `_compute_first_picking_ids` ? There are a couple of `mapped` calls we can probably replace either with a search or SQL query that should improve timings.

PeterC10:
> Can we do something similar for `_compute_first_picking_ids` ? There are a couple of `mapped` calls we can probably replace either with a search or SQL query that should improve timings.

We've attempted a change with an SQL query, but the results were slower. It probably just needs more fine tuning but that will be out of scope for this story, we will just leave the changes for the other compute method.

PR #644, opened 2022-05-27
Story #1711 Mobile UI: port udes_workflows to UDES14 Outbound

jrbirch99:
[FIX] udes_open: Stop python from raising an error as it expects an argument to add_unexpected_parts().

Over receiving is not to be handled yet however when over-receiving python would raise an error as add_unexpected_parts() had no argument.

Story/1711


ws-mps:
There is another PR with this change:

https://github.com/unipartdigital/udes-open/pull/641/files

We can probably close this one.

jrbirch99:
> There is another PR with this change:
> 
> https://github.com/unipartdigital/udes-open/pull/641/files
> 
> We can probably close this one.

Yep will close this PR then. 

This PR is being closed as the changes have already been made in another PR.

PR #645, opened 2022-05-30
Story #1711 Mobile UI: port udes_workflows to UDES14 Outbound

jrbirch99:
Sudo()s are being moved to inner functions in udes_stock. 

Story/1711

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -855,7 +855,9 @@ class StockPicking(models.Model):
                 else:
                     requires_backorder = True
             if requires_backorder:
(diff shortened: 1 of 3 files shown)


LoKKeR96:
Can we add a note for all sudos added so that we know why they are there? Can you also make sure these comments are titled with the same name? We might review these in the near future so it will be useful to be able to find them with ease.

caleb-shelton (approved):

PeterC10 (approved):

jrbirch99:
> Instead of adding many sudo-s can we make at the beginning of methods where needed with the comment self = self.sudo() example, does it overpass access rights

Not sure what you mean, can you explain some more :)

armandcela (approved):

adamrpatrick (approved):

PR #647, opened 2022-06-06
Story/1821

armandcela:


kdwyer (approved):

addons/udes_sale_stock/exceptions.py
```diff
@@ -0,0 +1,19 @@
+"""Custom exceptions for the UDES Sale Stock addon."""
+
+
+from odoo.exceptions import ValidationError
+
+
+class CombinedException(ValidationError):
```
PeterAlabaster, line 7: This looks as if it could live in `udes_common`

PeterAlabaster (changes requested):
Looks good just 2 minor comments

PeterAlabaster (approved):

PR #653, opened 2022-06-09
[ADD] udes_warehouse_classification: Get warehouse classification

armandcela:
for all products

Story: 1635

### DIFF
--- a/addons/udes_warehouse_classification/models/stock_picking.py
+++ b/addons/udes_warehouse_classification/models/stock_picking.py
@@ -5,10 +5,19 @@ from odoo import fields, models
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
-    def _get_classification_messages_for_product_picking(self):
-        """Return the product with classification messages appropriate for a picking"""
+    def _get_classification_messages_for_product_picking(self, all_products=False):
+        """
+        Return the product with classification messages appropriate for a picking if
+        all_products is False.
+        Return all products with classification messages appropriate if all_products is True
+        """
+        Product = self.env["product.product"]
+        self.ensure_one()
         product_classifications = super()._get_classification_messages_for_product_picking()
-        products = self.mapped("move_lines.product_id")
+        if all_products:
+            products = Product.search([("u_has_warehouse_classifications", "=", True)])
+        else:
+            products = self.mapped("move_lines.product_id")
         for product in products:
             classifications = product.u_product_warehouse_classification_ids.filtered(
                 lambda c: self.picking_type_id in c.picking_type_ids
--- a/addons/udes_warehouse_classification/models/product_template.py
+++ b/addons/udes_warehouse_classification/models/product_template.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
-from odoo import fields, models
+from odoo import fields, models, api
 
 _logger = logging.getLogger(__name__)
 
@@ -13,6 +13,22 @@ class ProductTemplate(models.Model):
         string="product_warehouse_classification_ids",
         help="Classifications on products used for messaging purposes.",
     )
+    u_has_warehouse_classifications = fields.Boolean(
+        "Has Warehouse Classifications?",
+        compute="_compute_has_warehouse_classifications",
+        store=True,
+    )
+
+    @api.depends("u_product_warehouse_classification_ids")
+    def _compute_has_warehouse_classifications(self):
+        """
+        For each product in self, set True if the product contains warehouse classifications,
+        otherwise False
+        """
+        for product in self:
+            product.u_has_warehouse_classifications = bool(
+                product.u_product_warehouse_classification_ids
+            )
 
     def get_classification_messages_for_report(self, report_name):
         """Find product classification messages needed for a given report"""


addons/udes_warehouse_classification/models/stock_picking.py
```diff
@@ -6,16 +6,24 @@ class StockPicking(models.Model):
...
+        """
+        Return all products with classification messages appropriate to warn users. It is getting
+        all products, not only the products of the picking because it may be some products with
+        classifications that are not inside the picking but we still need to warn users.
+        """
+        Product = self.env["product.product"]
+        self.ensure_one()
         product_classifications = super()._get_classification_messages_for_product_picking()
-        products = self.mapped("move_lines.product_id")
+        products = Product.search([("u_has_warehouse_classifications", "=", True)])
```
jrbirch99, line 17: Q: why do we do this on every product in the system? I think if a warehouse user has cherries, and bananas in his, he should only get warning messages about those products not the ones not in his picking. Maybe a question for a PO.

jrbirch99, line 17: Also do we need the extra field `u_has_warehouse_classifications`? Could we perhaps not just have something like 
```suggestion
        products = Product.search([("u_product_warehouse_classification_ids", "!=", False)])
```

armandcela: I think that i tried , but didnt work. And thought is safer to check with a boolean field. Good spot though

addons/udes_warehouse_classification/models/stock_picking.py
```diff
@@ -6,16 +6,24 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     def _get_classification_messages_for_product_picking(self):
-        """Return the product with classification messages appropriate for a picking"""
+        """
+        Return all products with classification messages appropriate to warn users. It is getting
+        all products, not only the products of the picking because it may be some products with
+        classifications that are not inside the picking but we still need to warn users.
```
LoKKeR96, line 12:
```suggestion
        Return classification messages for all products that have one.
        Note: returns all products because we may still need to warn 
        users if scan a product that was not expected on a picking
```

If you like my suggestion, use it, if not, feel free to resolve it.

addons/udes_warehouse_classification/models/stock_picking.py
LoKKeR96: I was also thinking that, I think SQL can handle that so Odoo should be able too? Let's see.

jrbirch99: To me it seems like we only want to do the check on the expected and unexpected items that has been scanned in, not all the products in a specific picking_type. Not sure how that would be done though.

addons/udes_warehouse_classification/models/stock_picking.py
armandcela: I found it not confusing , not even a little. Maybe because i know what i wrote :)
Do you have any suggestion how to write, as i am afraid that you still will not like.

LoKKeR96: That makes sense :D I wrote a suggestion there with the docstring paraphrased, can you not see it? Let me know

addons/udes_warehouse_classification/models/stock_picking.py
jrbirch99: sorry I meant 
```suggestion
        products = Product.search([]).filtered(lamdba p: p.u_product_warehouse_cassification_ids)
```

armandcela: Yes but unexpected items are products which are not on moves that why are unexpected. 
Check the technical notes on the story if that helps

jrbirch99: I am not sure because the new field does also have a compute method which goes through every product in self anytime a new u_product_warehouse_classification_ids is added to a product, which is a lot of iterations.

LoKKeR96 (approved):

addons/udes_warehouse_classification/models/stock_picking.py
armandcela: Yes but u_product_warehouse_classification_ids are not added very frequently , and only added on form view which will be only one product.
While Goods in or other flows that use the same controller are used a lot, which trigger calling the method.
I think is better to ask a TL about this if you dont agree

jrbirch99: Yeah I understand now

jrbirch99: Yeah I see your point, it would be worse if the user on the HHT is slowed down during Goods in then the desktop user being slowed down in form view.

jrbirch99 (approved):

addons/udes_warehouse_classification/models/stock_picking.py
armandcela: Expect of types of user, u_product_warehouse_classification_ids is a config field which we think will be used less frequently than operations.
I mean its not that we give priority to users on HHT comparing to the desktop users.

PR #658, opened 2022-06-14
[ADD] udes_stock: Port location state

peterpratt13:
Port the location state from 11 to 14. The only possible location states
are:
  * empty
  * archived
  * has_stock
Blocked is currently not ported.

User-story: 2106

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -13,7 +13,7 @@ class StockLocation(models.Model):
         "u_location_storage_format",
         "location_id",
         "location_id.u_storage_format",
-        "location_id.u_location_storage_format"
+        "location_id.u_location_storage_format",
     )
     def _compute_storage_format(self):
         """Determine the storage format of the location.
@@ -70,9 +70,7 @@ class StockLocation(models.Model):
         help="Product category speed to match with location speed.",
     )
     u_size = fields.Float(
-        string="Location size",
-        help="""The size of the location in square feet.""",
-        digits=(16, 2),
+        string="Location size", help="The size of the location in square feet.", digits=(16, 2)
     )
     u_location_category_id = fields.Many2one(
         comodel_name="stock.location.category",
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -106,9 +104,51 @@ def _domain_height_category(self):
 
                 If not set then the location will use the format
                 of the nearest ancestor that has a format specified.
-                """
+                """,
+    )
+    u_state = fields.Selection(
```
armandcela, line 109: Question: I know that fields that we add have the underscore, maybe in this case we may need an exception.
Field state,active, parent_id example are special fields which i think is better to be state as can be easier on xml to add invisible attributes

addons/udes_stock/models/stock_location.py
```diff
@@ -106,9 +104,51 @@ def _domain_height_category(self):
...
+        to fields in the location form, which doesn't include quant_ids. This can lead to the
+        computed value temporarily displaying empty before being set to has_stock once the record
+        is saved.
+
+        TODO: Handle blocked locations
+        """
+        Quant = self.env["stock.quant"]
+
+        for location in self:
+            state = False
```
armandcela, line 137: Would suggest to be empty by default, so maybe its better to initiate with `state="empty"`

armandcela (commented):
LGTM, just the name of field state i would suggest to change

addons/udes_stock/models/stock_location.py
PeterC10: This is not meant as a standard state field, it doesn't track progress or anything like that.

addons/udes_stock/models/stock_location.py
```diff
@@ -106,9 +104,51 @@ def _domain_height_category(self):
...
+
+            if location.id and location.usage == "internal":
+                if not location.active:
+                    state = "archived"
+                # TODO: Not implemented yet
```
PeterC10, line 142: I thought we had a simple blocking mechanism, will take a look.

armandcela (approved):

addons/udes_stock/models/stock_location.py
```diff
@@ -70,9 +70,7 @@ def _domain_height_category(self):
         help="Product category speed to match with location speed.",
     )
     u_size = fields.Float(
-        string="Location size",
-        help="""The size of the location in square feet.""",
-        digits=(16, 2),
+        string="Location size", help="""The size of the location in square feet.""", digits=(16, 2)
```
PeterC10, line 73: Nit: Not strictly your change but can we remove the triple quotes in the help string?

addons/udes_stock/models/stock_location.py
PeterC10: Blocking is currently done via `udes_simple_location_blocking`, so we'd need to override this method in that module. Going to be a bit awkward though as we'd want to keep this order intact.

addons/udes_stock/models/stock_location.py
```diff
@@ -106,9 +104,51 @@ def _domain_height_category(self):
...
+            ("has_stock", "Has Stock"),
+            ("archived", "Archived"),
+        ],
+        compute="_compute_state",
+        store=True,
+        string="State",
+        help="""Computed field describing the state of the stock location.""",
     )
 
+    @api.depends("quant_ids", "usage", "active")
```
PeterC10, line 122: Sorry something that wasn't considered while making the story, we'll also need `u_is_countable` and `u_location_is_countable` - so that we cut down on locations that have this computed. Can you also port these two fields?

addons/udes_stock/models/stock_location.py
PeterC10: Could be worth renaming it something like `u_countable_state` instead, to be a bit more explicit that it doesn't apply to all locations.

addons/udes_stock/models/stock_location.py
peterpratt13: Sorry, I didn't realise this was wanted for some reason i read this and was like that is a later todo. Added the last two commits for this. (Have force pushed another - the if/elif logic is in a different method that overrides it)

addons/udes_stock/models/stock_location.py
```diff
@@ -106,8 +104,90 @@ def _domain_height_category(self):
...
+        compute="_compute_state",
+        store=True,
+        string="State",
+        help="""
+        Computed field describing the state of a stockable location that is countable, 
+        see u_countable field.""",
+    )
+
+    @api.depends("u_location_is_countable", "location_id", "location_id.u_is_countable")
+    def _compute_countable_locations(self):
```
PeterC10, line 145: Very minor: Could we rename the computed methods to `_compute_is_countable` and `_compute_countable_state` to fit in the more with the Odoo standard?

peterpratt13: Ah good point - the name no longer matches really. Will change

PeterC10 (approved):

PR #659, opened 2022-06-16
[ADD] udes_stock: Date of transfer applies to cancelled pickings

armandcela:
Story: 2013

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -820,6 +820,15 @@ class StockPicking(models.Model):
         """
         raise ValidationError(_("Cannot handle over receiving!"))
 
+    def action_cancel(self):
+        """Extend to set date_done field when cancelling transfers"""
+        res = super().action_cancel()
+
+        # To be reviewed later, as our priority approach is different. Having here just to remember
+        # that when we made changes we need to change _action_done method as well on stock module
+        self.write({"date_done": fields.Datetime.now(), "priority": "0"})
+        return res
+
     def validate_picking(self, create_backorder=False, force_validate=False):
         """ Validates a picking and returns its backorder if any has been created.
             Will raise an error if create_backorder is False and there are incomplete lines.
--- a/addons/udes_stock/views/stock_picking.xml
+++ b/addons/udes_stock/views/stock_picking.xml
@@ -158,6 +158,31 @@
                 <attribute name="attrs">{'readonly': [('state', '!=', 'draft')],  'required': [('picking_type_code', '=', 'incoming')]}</attribute>
             </xpath>
 
+            <xpath expr="//field[@name='date_done']" position="attributes">
+                <attribute name="attrs">{'invisible': [('state', 'not in', ('done', 'cancel'))]}</attribute>
+                <attribute name="string"/>
+            </xpath>
+
+        </field>
+    </record>
+
+
+    <!-- Adding Date of Transfer in tree view, and overriding scheduled_date to be shown always without remaining_days widget -->
+    <record id="vpicktree" model="ir.ui.view">
+        <field name="name">stock.picking.tree.udes</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.vpicktree"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//field[@name='scheduled_date']" position="attributes">
+                <attribute name="widget"></attribute>
+                <attribute name="attrs"></attribute>
+            </xpath>
+
+            <xpath expr="//field[@name='scheduled_date']" position="after">
+                <field name="date_done" optional="show"/>
+            </xpath>
+
         </field>
     </record>
 


peterpratt13:
nit: Commit message data -> Date

addons/udes_stock/models/stock_picking.py
```diff
@@ -820,6 +820,12 @@ def add_unexpected_parts(self, product_quantities):
         """
         raise ValidationError(_("Cannot handle over receiving!"))
 
+    def action_cancel(self):
+        """Extend to set date_done field When cancelling transfers"""
+        res = super().action_cancel()
+        self.write({'date_done': fields.Datetime.now(), 'priority': '0'})
```
peterpratt13, line 826: nit: Black format.

PeterC10: Does this in core Odoo too, we'll probably need to review this.

In Odoo terms it sets it to normal priority, so something cancelled/done can never be urgent... can see where they are coming from but conflicts against the way we want to use priorities.

addons/udes_stock/views/stock_picking.xml
```diff
@@ -158,6 +158,30 @@
...
+            <xpath expr="//field[@name='date_done']" position="attributes">
+                <attribute name="attrs">{'invisible': [('state', 'not in', ('done', 'cancel'))]}</attribute>
+                <attribute name="string">Date of Transfer</attribute>
+            </xpath>
+
+        </field>
+    </record>
+
+
+    <record id="vpicktree" model="ir.ui.view">
```
PeterC10, line 170: Suggestion: Add a comment above each override to explain why we are making the changes.

addons/udes_stock/views/stock_picking.xml
```diff
@@ -158,6 +158,30 @@
...
+        <field name="name">stock.picking.tree.udes</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.vpicktree"/>
+        <field name="arch" type="xml">
+
+            <field name="scheduled_date" position="attributes">
+                <attribute name="widget">datetime</attribute>
```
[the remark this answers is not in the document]
PeterC10: Ah looks like it is set to use `remaining_days`, would setting this to blank also have the same change? Just curious.

addons/udes_stock/views/stock_picking.xml
```diff
@@ -158,6 +158,30 @@
...
+                <attribute name="attrs"></attribute>
+            </field>
+
+            <field name="scheduled_date" position="after">
```
PeterC10, line 181: I don't have a problem personally with hooking into fields this way, but we've tended to use `xpath` as standard.

Will check tomorrow because I can't remember if we had any policy around this tbh, but this can probably be ignored. :)

addons/udes_stock/views/stock_picking.xml
```diff
@@ -158,6 +158,30 @@
                 <attribute name="attrs">{'readonly': [('state', '!=', 'draft')],  'required': [('picking_type_code', '=', 'incoming')]}</attribute>
             </xpath>
 
+            <xpath expr="//field[@name='date_done']" position="attributes">
+                <attribute name="attrs">{'invisible': [('state', 'not in', ('done', 'cancel'))]}</attribute>
+                <attribute name="string">Date of Transfer</attribute>
```
PeterC10, line 163: It is annoying to have to explicity set the string back to what it is set in the Python code.

Could you try the following and see if it works?

`<attribute name="string"/>`

addons/udes_stock/models/stock_picking.py
PeterC10: Will check this again in the morning, but probably something to handle in another story, makes sense to make it consistent with completed pickings atm I think.

addons/udes_stock/models/stock_picking.py
```diff
@@ -820,6 +820,12 @@ def add_unexpected_parts(self, product_quantities):
         """
         raise ValidationError(_("Cannot handle over receiving!"))
 
+    def action_cancel(self):
+        """Extend to set date_done field When cancelling transfers"""
```
PeterC10, line 824: Minor typo: `When` -> `when`

addons/udes_stock/views/stock_picking.xml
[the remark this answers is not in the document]
PeterC10: So maybe:

`<attribute name="widget"/>`

addons/udes_stock/models/stock_picking.py
peterpratt13: Ah yeah - https://github.com/unipartdigital/odoo/commit/e40553723b893bcd2244e9b9fcc5c1a7e1bf27d3

PeterC10:
```
- MO/picking/PO/stock.move have now the "same" `priority` fields
(Boolean logic: Normal or Urgent).
```

Yeah that's not going to play nicely at all with how we want to use priorities in UDES...! Will raise an issue in the morning. :)

addons/udes_stock/views/stock_picking.xml
armandcela: In fact on field definition on python it was Date of Transfer, but on xml in the form view was changed to Effective Date.
We discussed in the morning and i remember that you said better on python to be same on extended filters

addons/udes_stock/views/stock_picking.xml
armandcela: Ok it's fine , know that odoo itself uses this way when its can be found easily example not inside pages etc.
But happy to change with xpath if we have that standard

addons/udes_stock/views/stock_picking.xml
PeterC10: Yeah, so the name on the Python side is fine. What I mean is that here we are now manually setting it in the XML even though it is the same as in Python. So that is why I'm wondering if we could just remove the `string` attribute on the XML view.

addons/udes_stock/views/stock_picking.xml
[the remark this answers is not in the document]
armandcela: It was widget="remaining_days" in main form view and that widget shows Today
There are 2 other widgets for date fields as far i noticed "date" and "datetime". In this case it is needed datetime because we want to show hrs too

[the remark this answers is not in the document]
armandcela: Even that can be as field is datetime , most probably if no widget will show as datetime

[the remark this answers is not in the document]
PeterC10: Yeah so if the default is `datetime`, which it seems it still is, then maybe we can just remove the `widget` attribute to be a tiny bit cleaner. Not a big deal at all though and happy to resolve if you want to keep it explicit.

The main thing is I'm relieved that by default dates and datetimes in the UI will not turn into `Today`, `Tomorrow` etc. :)

addons/udes_stock/models/stock_picking.py
peterpratt13: Will resolve as not part of this story

peterpratt13 (approved):

addons/udes_stock/views/stock_picking.xml
PeterC10: Will resolve, couldn't find anything.

Also as it is a tree view we're only ever going to have a `tree` element with `field` and `button` elements inside. So we don't have the potential like form views where you can have fields from one2many, many2many etc.

PeterC10 (approved):

PR #665, opened 2022-06-24
[FIX] udes_stock: Adding company on replen route

armandcela:
Story: 2022


armandcela:
Found an issue while upgrading udes_stock module o gentoo instance.
On Gentoo story added the warehouse in replen stock rule, and i think because warehouse is linked with company id was crying that route needs to be in a company.
Had a discussion with Peter Clark and he approved the change.

jrbirch99 (approved):

jrbirch99:
Q: Why are the units test on udes_open failing?

PeterC10:
> Q: Why are the units test on udes_open failing?

Problem with the `get_info` setup, issue 2280 will track and resolve this.

PR #666, opened 2022-06-28
Story #2021 Sale - Orders Improvements and Fixes

jrbirch99:
A default filter was removed from the search bar on the Quotations page. 

All pickings associated with a Sale Order are now linked to the sale order, and all moves are now linked to a sale order line. 

The fix involved ensuring that sale_line_id was propagated with other move information down the procurement rule chain (as they are created recursively) i.e. move information for one picking type is given to a picking type down the chain to create a move and on and on. This way all moves in the chain were attached to the sale order.

### DIFF
--- a/addons/udes_sale_stock/tests/test_sale_order.py
+++ b/addons/udes_sale_stock/tests/test_sale_order.py
@@ -1,0 +1,107 @@
+from odoo.addons.udes_stock.tests import common
+from odoo import fields
+
+
+class TestSaleOrder(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestSaleOrder, cls).setUpClass()
+        # Unlink the TestGoodsOut pull_push rule as it was causing the route to misbehave. 
+        # The rule was not being triggered when a sale order was created in the customer location i.e.
+        # no pickings were being created for GoodsOut or Pick.
+        # Add in the final step to the route which links customers and the warehouse i.e. Trailer Dispatch rule
+        Rule = cls.env["stock.rule"]
+        cls.Picking = cls.env["stock.picking"]
+
+        Rule.search([("name", "=", "TestGoodsOut")]).unlink()
+        cls.rule_out = Rule.create(
+            {
+                "name": "TestTrailer",
+                "route_id": cls.route_out.id,
+                "picking_type_id": cls.picking_type_trailer_dispatch.id,
+                "location_id": cls.picking_type_trailer_dispatch.default_location_dest_id.id,
+                "location_src_id": cls.picking_type_trailer_dispatch.default_location_src_id.id,
+                "action": "pull",
+                "procure_method": "make_to_order",
+            }
+        )
+
+    def test_all_created_pickings_are_attached_to_sale_order(self):
+        """
+        Create the sale order for a single product. Check all the pickings created due to sale order are attached to the sale order.
+        """
+        sale_order = self._create_sale_order(self.apple.id, 10)
+        sale_order.action_confirm()
+
+        all_created_pickings = self.Picking.search([])
+        for created_picking in all_created_pickings:
+            with self.subTest(created_picking=created_picking):
+                self.assertIn(created_picking, sale_order.picking_ids)
+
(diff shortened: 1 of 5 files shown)


addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
...
+        sale_order.action_confirm()
+
+        all_created_pickings = Picking.search([])
+        sale_order.action_cancel()
+
+        for picking in all_created_pickings:
+            with self.subTest(picking=picking):
+                self.assertEqual(picking.state, "cancel")
+
+    def test_only_moves_created_by_sale_order_are_cancelled(self):
```
kdwyer, line 55: All these test names are very expressive :+1:

addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
...
+                "action": "pull",
+                "procure_method": "make_to_order",
+            }
+        )
+
+    def test_all_created_pickings_are_attached_to_sale_order(self):
+        """
+        Create the sale order for a single product. Check all the pickings created due to sale order are attached to the sale order.
+        """
+        Picking = self.env["stock.picking"]
```
PeterC10, line 29: Suggestion: Put this variable into `setUpClass` as it is being called in 3 of the tests.

So in `setUpClass` you can do: `cls.Picking = cls.env["stock.picking"]`

Then you can remove `Picking = self.env["stock.picking"]` from the three tests.

addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
+from odoo.addons.udes_stock.tests import common
+from odoo import fields
+
+
+class TestSaleOrder(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestSaleOrder, cls).setUpClass()
+        # Unlink the TestGoodsOut pull_push rule as it was causing the route to misbehave
```
kdwyer, line 9: It would be better to explain precisely how the route misbehaved.  That way we know what to look out for if we reinstate the rule, and can distinguish any additional unexpected behaviour.

addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
...
+    
+    def _create_sale_order(self, product_id, product_uom_qty):
+        """
+        Create a sale order for a the product_id
+        """
+        Partner = self.env["res.partner"]
+        SaleOrder = self.env["sale.order"]
+        SaleOrderLine = self.env["sale.order.line"]
+
+        partner = Partner.search([("name", "=", "Administrator")])
```
PeterC10, line 94: Suggestion: Use the XML reference rather than search, unlikely but this could break if a test added another partner called "Administrator".

`partner = self.env.ref("base.partner_admin")` should do it.

addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
...
+        sale_order = SaleOrder.create(
+            {
+                "partner_id": partner.id,
+                "client_order_ref": "sale order",
+                "commitment_date": fields.Datetime.now(),
```
PeterC10, line 99: Suggestion: Have a `now` variable, and use this for `commitment_date` and `requested_date`.

addons/udes_sale_stock/tests/test_sale_order.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertEqual(len(created_picks), 2)
+        second_picking.move_lines.write({"picking_id": first_picking.id})
+        self.assertEqual(len(first_picking.move_lines), 2)
+
+        second_sale_order.action_cancel()
+        self.assertEqual(second_move.state, "cancel")
+    
+    def _create_sale_order(self, product_id, product_uom_qty):
+        """
+        Create a sale order for a the product_id
```
PeterC10, line 88: Typo: `for a the`.

kdwyer (approved):

armandcela (approved):

PR #667, opened 2022-06-29
[ADD] udes_stock: u_heatmap_data_updated column to stock.location

PeterAlabaster:
NB: Needed to change slightly to use `._origin` as the date in the
database would not get set otherwise.

User-story: 2317


addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,41 @@ def _domain_height_category(self):
...
 
+    u_heatmap_data_updated = fields.Datetime(
+        string="Heatmap data updated",
+        help="""
+            The date when the location name, co-ordinates or size was last changed. 
+        """,
+        default=fields.Datetime.now,
+    )
+
+    def set_u_heatmap_data_updated(self, vals):
```
PeterAlabaster, line 145: not 100% on the name, initially had `update_u_heatmap_data_updated()` but that also felt a bit weird

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,48 @@ def _domain_height_category(self):
         see u_countable field.""",
     )
 
+    u_heatmap_data_updated = fields.Datetime(
+        string="Heatmap data updated",
+        help="""
+            The date when the location name, co-ordinates or size was last changed.
```
PeterC10, line 140: Can this be on one line?

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,48 @@ def _domain_height_category(self):
...
+        # only want to update the subsets `u_heatmap_data_updated` datetime.
+        locations_to_update.with_context(
+            bypass_heatmap_check=True
+        ).u_heatmap_data_updated = fields.Datetime.now()
+
+
+    def write(self, vals):
+        if not self.env.context.get("bypass_heatmap_check"):
+            self.set_u_heatmap_data_updated(vals)
+        return super(StockLocation, self).write(vals)
```
PeterC10, line 177: Suggestion: Can just do `super().write(vals)`

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,48 @@ def _domain_height_category(self):
...
+        Iterate over locations in `self`, checking if any heatmap fields
+        have been written to with updated values, and if so - dispatch to write()
+        with a bypass in context (to avoid recursion) to update `u_heatmap_data_updated`
+        with the current date & time.
+
+        :param: vals: dict() of vals from `write()`
+        """
+        StockLocation = self.env["stock.location"]
+        locations_to_update = StockLocation.browse()
+        for location in self:
```
PeterC10, line 156: Might be worth checking if any of the heatmap fields are in `vals.keys()` before looping through.

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,48 @@ def _domain_height_category(self):
...
+
+        # If .write() is called on multiple locations
+        # but only a subset of those locations values have changed, we
+        # only want to update the subsets `u_heatmap_data_updated` datetime.
+        locations_to_update.with_context(
+            bypass_heatmap_check=True
+        ).u_heatmap_data_updated = fields.Datetime.now()
+
+
+    def write(self, vals):
```
PeterC10, line 174: Can we have a doc-string here? Probably only a few words needed.

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +134,48 @@ def _domain_height_category(self):
         see u_countable field.""",
     )
 
+    u_heatmap_data_updated = fields.Datetime(
```
peterpratt13, line 137: Could make use of PreciseDatetime here perhaps?

peterpratt13:
LGTM just minor comments

addons/udes_stock/models/stock_location.py
PeterAlabaster: sorry, do you mean good as is, or you'd like it changed to `update_u_heatmap_data_updated()`?

addons/udes_stock/models/stock_location.py
PeterAlabaster: How would this affect the timestamp in the incoming location streams data? would it remove the need for the `/1000`?

addons/udes_stock/models/stock_location.py
PeterAlabaster: good shout

addons/udes_stock/models/stock_location.py
peterpratt13: I believe so

addons/udes_stock/models/stock_location.py
```diff
@@ -134,6 +135,50 @@ def _domain_height_category(self):
...
+        with the current date & time.
+
+        :param: vals: dict() of vals from `write()`
+        """
+        StockLocation = self.env["stock.location"]
+        locations_to_update = StockLocation.browse()
+        for location in self:
+            heatmap_fields = ["u_size", "name", "posx", "posy", "posz"]
+            # Don't need to check old/new values of these fields if they aren't in vals
+            if not any([x in vals.keys() for x in heatmap_fields]):
```
PeterC10, line 158: We can put this (along with `heatmap_fields`) outside of the loop completely. Even if there are multiple locations in `self`, `vals` will always be the same for each location.

So we can check the vals before we start looping through locations.

PeterC10 (approved):

caleb-shelton (approved):

kdwyer (approved):

kdwyer (approved):

PR #671, opened 2022-07-08
Story #2015 Improve tests common code of user permissions

jrbirch99:
Correct permissions were implemented for users in the udes_open unit tests.

### DIFF
--- a/addons/udes_stock/tests/test_stock_picking_user_assignment.py
+++ b/addons/udes_stock/tests/test_stock_picking_user_assignment.py
@@ -6,16 +6,15 @@ class TestUserAssignments(common.BaseUDES):
     def setUpClass(cls):
         super().setUpClass()
 
-        # TODO - Fix to be outbound_user and inbound_user when users are implemented
-        cls.inbound_user = cls.env["res.users"].search([("login", "=", "admin")])
-        cls.inbound_user_2 = cls.inbound_user.copy(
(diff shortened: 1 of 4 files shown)


jrbirch99:
The tests seem to be failing due to the infamous get_info unit test.

addons/udes_stock/tests/test_stock_picking_user_assignment.py
```diff
@@ -6,16 +6,19 @@ class TestUserAssignments(common.BaseUDES):
     def setUpClass(cls):
         super().setUpClass()
 
-        # TODO - Fix to be outbound_user and inbound_user when users are implemented
-        cls.inbound_user = cls.env["res.users"].search([("login", "=", "admin")])
-        cls.inbound_user_2 = cls.inbound_user.copy(
-            {"name": "Inbound User 2", "login": "inbound_user_2_login"}
+        cls.stock_user = cls.create_user(
```
armandcela, line 9: To me it doesnt seem right changing the test user from inbound to stock user.
 I would suggest to move the tests where inbound users are created, or if these tests are needed to change names and to create new ones for specific users in udes_stock_permission example. Dont know if it makes sense

LoKKeR96 (approved):
LGTM apart from the issue that Armand mentioned but as you said it is out of scope to fix that.

armandcela (approved):

PR #673, opened 2022-07-12
Story/2466 - Inventory Adjustment overhaul and various changes/improvements for UDES

PeterC10:
Descriptions in the individual commits.

### DIFF
--- a/addons/udes_stock/__manifest__.py
+++ b/addons/udes_stock/__manifest__.py
@@ -32,6 +32,7 @@
         "report/external_layout_standard.xml",
         "wizard/stock_backorder_confirmation_views.xml",
         "wizard/stock_immediate_transfer_views.xml",
+        "views/assets.xml",
         "views/product_views.xml",
         "views/stock_picking.xml",
         "views/stock_picking_type.xml",
--- a/addons/udes_stock/data/locations.xml
+++ b/addons/udes_stock/data/locations.xml
@@ -5,6 +5,7 @@
     <!-- Update "Warehouse/Stock" view location -->
     <record id="stock.stock_location_stock" model="stock.location">
       <field name="usage">view</field>
+      <field name="u_location_is_countable">yes</field>
     </record>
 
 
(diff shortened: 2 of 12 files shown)


LoKKeR96 (approved):
This must have been tedious work, did you get inspired by Odoo 11?

addons/udes_stock/models/stock_inventory.py
```diff
@@ -0,0 +1,97 @@
+from odoo import models, fields
+
+
+class Inventory(models.Model):
+    _inherit = "stock.inventory"
+
+    # Make locations field mandatory
+    location_ids = fields.Many2many(required=True)
```
armandcela, line 8: Question:
Why are these changes thought to be here and not in udes_stock_check module?

In that case should we consider that overriding location_ids in udes_stock and removing at all in udes_stock_check.
Because we created udes_stock_check for stock inventory changes it sound better to me that these changes are in udes_stock_check

addons/udes_stock/__manifest__.py
```diff
@@ -32,6 +32,7 @@
         "report/external_layout_standard.xml",
         "wizard/stock_backorder_confirmation_views.xml",
         "wizard/stock_immediate_transfer_views.xml",
+        "views/assets.xml",
```
armandcela, line 35: Nit: think is forgotten to add the views for stock inventory and location in manifest if i am not missing smth

PeterAlabaster (approved):
LGTM, Sent PM about potential need to amend an inherited view elsewhere.

PeterC10:
> This must have been tedious work, did you get inspired by Odoo 11?

One way of putting it. :)

Essentially a lot of the issues product had with inventory adjustments was the amount of clicks needed to get things done, and part of this was caused by having to go into another screen to do the actual adjustments.

addons/udes_stock/models/stock_inventory.py
PeterC10: So I want to be able to offer a better interface by default with UDES, so if we ever have a custom module later down the line, we could have "the-boy-who-lived" which doesn't want the "you-know-who" functionality.

But I take the point that `udes_stock` is becoming bloated, which we want to avoid. So will put these changes into `udes_stock_inventory`.

PeterC10: Moved the functionality to a new module.

PR #677, opened 2022-07-13
Story #2168 Move Cancellation message to show partially cancelled qtys correctly

jrbirch99:
[IMP] udes_stock: Overwrite the get_current_user_id method.

Story/2168

### DIFF
--- a/addons/udes_stock/models/res_users.py
+++ b/addons/udes_stock/models/res_users.py
@@ -81,3 +81,11 @@ class ResUser(models.Model):
     def unassign_user_on_logout(self):
         """Unassign a user from a picking on user logout"""
         return
+    
+    def get_current_user_id(self):
+        """
+        Override function to use `real_uid` instead of `self.env.user.id`,
+        as the original function is defined in a module which can not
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/res_users.py
```diff
@@ -81,3 +81,6 @@ def assign_picking_to_users(self, picking):
     def unassign_user_on_logout(self):
         """Unassign a user from a picking on user logout"""
         return
+    
+    def get_current_user_id(self):
+        return self.env.real_uid
```
PeterAlabaster, line 86: nit, can we add a docstring?
```python
    def get_current_user_id(self):
        """
        Override function to use `real_uid` instead of `self.env.user.id`,
        as the original function is defined in a module which can not
        necessarily depend on UDES odoo customisations
        """
        return self.env.real_uid
```
or something similar, i'm not 100% sure - is `real_uid` coming from our changes to `odoo` submodule?

PeterC10 (approved):
LGTM after doc-string added as Peter A suggested. :)

PR #678, opened 2022-07-13
Fix failing udes_get_info test

armandcela:
Returning false for related fields when they don't have value, in
this way we always will have the expected fields on get_info.

### DIFF
--- a/addons/udes_get_info/__manifest__.py
+++ b/addons/udes_get_info/__manifest__.py
@@ -6,7 +6,7 @@
     "website": "http://github/unipartdigital/udes-open",
     "category": "UDES",
     "version": "0.1",
-    "depends": ["base", "stock", "udes_common"],
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common"],
     "data": [],
     "demo": [],
 }
--- a/addons/udes_get_info/models/models.py
+++ b/addons/udes_get_info/models/models.py
@@ -65,8 +65,10 @@ def _get_info(self, level, info_fields, extra_fields):
         result = self[field_name]
         if isinstance(self[field_name], models.BaseModel):
             # Stop recursing once we reach level zero or if the field is empty.
-            if (level == 0) or (not result):
+            if level == 0:
                 continue
+            if not result:
+                _info = False
             # If field is Many2one return dict, otherwise list of dicts.
             # Passing extra fields means that it will look for extra fields on every level.
             # Example if needed default_barcode from product and the field is not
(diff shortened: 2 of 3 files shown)


addons/udes_get_info/models/models.py
```diff
@@ -65,16 +65,18 @@ def _get_info(self, level, info_fields, extra_fields):
         result = self[field_name]
         if isinstance(self[field_name], models.BaseModel):
             # Stop recursing once we reach level zero or if the field is empty.
-            if (level == 0) or (not result):
+            if level == 0:
                 continue
+            if not result:
+                _info = False
```
PeterAlabaster, line 71: Q: Are there impications to overwriting `_info` to be `False`, isn't that going to bubble up the recursion and blank out any other values which were previously there?

PeterAlabaster: ah, yes sorry, i didn't spot the `info[info_name] = _info` at the end, so it's only going to be changing that depths value to False and not affecting the rest of info. Gotcha

PeterAlabaster (approved):

addons/udes_get_info/tests/test_stock_quant.py
```diff
@@ -30,6 +30,7 @@ def setUpClass(cls):  # noqa: D102
                 "package_id",
                 "product_id",
                 "quantity",
+                "lot_id",
```
jrbirch99, line 33: Q: The change in models.py setting the `_info = False` when there is no result, is that needed to get the test to run? or is it a quality of life improvement. As it seems to me that the test would run fine just with this change if the quant had a lot_id.

addons/udes_get_info/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "website": "http://github/unipartdigital/udes-open",
     "category": "UDES",
     "version": "0.1",
-    "depends": ["base", "stock", "udes_common"],
+    "depends": ["base", "stock", "stock_picking_batch", "udes_common"],
```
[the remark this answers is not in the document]
kdwyer: So it isn't related to the bug that we are trying to fix?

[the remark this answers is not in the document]
armandcela: It was spotted when i was running isolated unit test for udes_get_info module, as was not installing stock_picking_batch therefore was failing on setup.

addons/udes_get_info/tests/test_stock_quant.py
armandcela: The change in models `_info = False` fixed the failing test as the `test_retrieves_expected_fields_from_related_model` which was the failing test dont rely on stock quant expected_keys. 

But we need to change expected keys as well after the change made to _info = False as it was failing the test_stock_quant which makes sense as we have added a new field on Quant._get_info_field_names

jrbirch99: Spoke in a call, I am happy with the change :smile:

PR #679, opened 2022-07-13
Story/2487

kdwyer:


addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -75,7 +72,8 @@ def _find_unreservable_moves(self, batch_size=1000, return_fast=False):
 
                     if product not in stock_for_products.keys():
                         stock_for_products[product] = Quant.get_available_quantity(
-                            product, locations
+                            product,
+                            picking.picking_type_id.default_location_src_id,
```
armandcela, line 76: I think all pickings are grouped by picking_type_id, so practically at the beginning of function can be set a variable to get the location, instead of accessing it for every pick from picking_type_id

At the beginning of method smth like:
`locations = self.picking_type_id.default_location_src_id`

```suggestion
                            locations,
```

jrbirch99 (approved):

addons/udes_stock_cron/models/stock_picking.py
kdwyer: Amended, please review.

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -62,6 +62,7 @@ def _find_unreservable_moves(self, batch_size=1000, return_fast=False):
                 ["product_id", "product_uom_qty", "state"], load="_classic_write"
             )
 
+            location = batched_pickings.picking_type_id.default_location_src_id
```
armandcela, line 65: I think on same place where it was previously 
locations not in the batched_pickings , but in self i believe as they are of the same picking type.
Batched is only splitting in chunks i believe

armandcela (approved):

addons/udes_stock_cron/models/stock_picking.py
kdwyer: Third time lucky? ;-)

armandcela: My bad , sorry :(

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -31,15 +31,13 @@ def _find_unreservable_moves(self, batch_size=1000, return_fast=False):
...
-        Location = self.env["stock.location"]
         Move = self.env["stock.move"]
         Quant = self.env["stock.quant"]
 
         # Create empty record sets for moves.
         unreservable_moves = Move.browse()
 
-        # Get unreserved stock for each product in locations
-        locations = Location.get_available_stock_locations()
+        location = self.picking_type_id.default_location_src_id
```
ws-mps, line 40: Why not `picking.location_id` after line 66 instead of here? Because picking type `default_location_src_id` could be technically different than the one of the picking.

PeterC10: What if a picking had its location manually changed but not the others? The product counts could be misleading at that point.

PeterC10: i.e you have Pick to pick from PICK ZONE

PICK1 - ZONE A
PICK2 - ZONE B
PICK3 - PICK ZONE
PICK4 - PICK ZONE

So probably would be safest to use the picking type location?

PeterC10 (approved):

PR #680, opened 2022-07-13
[IMP] udes_stock: Allow Stock Move to move from any location

LoKKeR96:
User-story: x2421

### DIFF
--- a/addons/udes_stock/data/picking_types.xml
+++ b/addons/udes_stock/data/picking_types.xml
@@ -76,8 +76,8 @@
       <field name="code">internal</field>
       <field name="sequence_id" ref="sequence_internal"/>
       <field name="use_existing_lots" eval="1"/>
-      <field name="default_location_src_id" ref="stock.stock_location_stock"/>
-      <field name="default_location_dest_id" ref="stock.stock_location_stock"/>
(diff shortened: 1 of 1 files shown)


jrbirch99 (approved):

addons/udes_stock/data/picking_types.xml
```diff
@@ -76,8 +76,8 @@
       <field name="code">internal</field>
       <field name="sequence_id" ref="sequence_internal"/>
       <field name="use_existing_lots" eval="1"/>
-      <field name="default_location_src_id" ref="stock.stock_location_stock"/>
-      <field name="default_location_dest_id" ref="stock.stock_location_stock"/>
+      <field name="default_location_src_id" ref="stock.warehouse0_view_location"/>
```
armandcela, line 79: Question: Story says about rockhopper, do we want by default in every possible package or future customer, or the change should be done in udes_rockhopper module.

The change looks good

armandcela (approved):

addons/udes_stock/data/picking_types.xml
LoKKeR96: I think this is fine since we want both adhoc and planned to behave the same so the answer is to have both act the same by default.

PR #681, opened 2022-07-14
User-story: 2500

ws-mps:
[FIX] udes_stock: Missing variable at find_move_lines lambda

* Added missing ml in the lambda function.
* Also have make product_id mandatory, since it does not make sense
to find move lines for a given quantity without specifying product.

### DIFF
--- a/addons/udes_stock/tests/test_stock_move_line.py
+++ b/addons/udes_stock/tests/test_stock_move_line.py
@@ -397,7 +397,7 @@ class TestStockMoveLine(common.BaseUDES):
 class TestStockMoveLinePrepareAndMarkMoveLines(common.BaseUDES):
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -534,6 +534,157 @@ def test_over_mark_error(self):
...
+            f"Move line {banana_ml.id} for product {prod_name} does not have enough "
+            f"quantity: 100 vs 5"
+        )
         self.assertEqual(e.exception.args[0], msg)
+
+
+class TestFindMoveLines(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestFindMoveLines, cls).setUpClass()
```
kdwyer, line 547: Just `super().setUpClass()` will do?

PeterC10 (approved):

PR #684, opened 2022-07-20
Story/2417 Restrict URL access based on menuitem access

PeterAlabaster:


addons/udes_security/controllers/main.py
```diff
@@ -5,12 +5,100 @@
...
+        group_ids_for_access = []
+        # ir.actions.client, ir.actions.act_url don't have groups_id parameter
+        if hasattr(action, "groups_id"):
+            group_ids_for_access += action.groups_id.ids
+        # Can have actions defined in the system with no menu associated. These should be blocked.
+        if not menu:
+            return False
+        group_ids_for_access += menu.groups_id.ids
+        user_has_any_groups = any(g_id in user.groups_id.ids for g_id in group_ids_for_access)
+        if (group_ids_for_access and user_has_any_groups) or not group_ids_for_access:
```
armandcela, line 96: Correct me if i am wrong.
In case that menu has groups `group_ids_for_access` is a list with some ids and user doesnt have any of this groups i think should return false not search to the parent menu isnt it?

addons/udes_security/tests/test_desktop_access.py
```diff
@@ -104,3 +98,140 @@ def test_all_users_can_access_database(self):
...
+        # Settings/Technical/Sequences & Identifiers/Sequences (ir.actions.act_window)
+        sequence_action = self.env.ref("base.ir_sequence_form")
+        # Settings/Users & Companies/Users (ir.actions.act_window)
+        users_action = self.env.ref("base.action_res_users")
+        # Discuss (ir.actions.client)
+        discuss_action = self.env.ref("mail.action_discuss")
+        # Apps/Apps/Third-Party Apps (ir.actions.act_url)
+        third_party_apps_action = self.env.ref("base.menu_third_party")
+        # DEBT: There are no ir.actions.server actions available, in future add one to setUpClass
+        # DEBT: There are no ir.actions.report actions available, in future add one to setUpClass
```
PeterC10, line 180: Nit: The debt has been doubled unfairly.

addons/udes_security/controllers/main.py
```diff
@@ -5,12 +5,100 @@
...
+        parent = menu.parent_id
+        group_ids_for_access = []
+        # ir.actions.client, ir.actions.act_url don't have groups_id parameter
+        if hasattr(action, "groups_id"):
+            group_ids_for_access += action.groups_id.ids
+        # Can have actions defined in the system with no menu associated. These should be blocked.
+        if not menu:
+            return False
+        group_ids_for_access += menu.groups_id.ids
+        user_has_any_groups = any(g_id in user.groups_id.ids for g_id in group_ids_for_access)
```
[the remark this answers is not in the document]
PeterAlabaster: will check

addons/udes_security/controllers/main.py
PeterAlabaster: if a menu/action has groups, and the user does not have any of those groups, it will hit the `else` block and return `False` (no need to continue searching upward)

addons/udes_security/controllers/main.py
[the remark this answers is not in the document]
PeterAlabaster: `user.groups_id.ids` runs this query
```sql
SELECT
    res_groups_users_rel.uid,
    res_groups_users_rel.gid
FROM
    res_groups_users_rel,
    "res_groups"
    LEFT JOIN "ir_translation" AS "res_groups__name" ON (
        "res_groups"."id" = "res_groups__name"."res_id"
        AND "res_groups__name"."type" = 'model'
        AND "res_groups__name"."name" = 'res.groups,name'
        AND "res_groups__name"."lang" = 'en_GB'
        AND "res_groups__name"."value" != ''
    )
WHERE
    1 = 1
    AND res_groups_users_rel.uid IN (2)
    AND res_groups_users_rel.gid = res_groups.id
ORDER BY
    COALESCE("res_groups__name"."value", "res_groups"."name") OFFSET 0
```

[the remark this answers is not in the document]
PeterAlabaster: i'll push it up to parent function too so we don't call it every time round

addons/udes_security/controllers/main.py
```diff
@@ -5,12 +5,100 @@
...
+        :return: ir.ui.menu(x,) recordset, will be empty if a non existent action_id is passed in
+        """
+        IrUiMenu = request.env["ir.ui.menu"]
+
+        if not isinstance(action_id, int):
+            raise ValidationError("Invalid action ID specified.")
+        # Escape the leading and trailing static % symbols with %,
+        # as we don't want to string format them in the execute() call
+        query = "SELECT id FROM ir_ui_menu WHERE action LIKE '%%,%s%%'"
+        request.env.cr.execute(query, (action_id,))
```
kdwyer, line 63: Do you think this would be easier to parse visually?
```python
query = "SELECT id FROM ir_ui_menu WHERE action LIKE '%%,%s%%'" 
request.env.cr.execute(query, (f'%{action_id}%',))
```

addons/udes_security/controllers/main.py
armandcela:
```
         if group_ids_for_access and not user_has_any_groups:
             return False
        else:
            return self._user_has_access_right(action, parent, user) if parent else True
```
How do you think to change the way of condition seems little confusing the first one

addons/udes_security/controllers/main.py
PeterAlabaster: I'm happy to f-string it (originally I had a similar thing), but I had the impression that string formatting outside of the .execute() call was generally frowned upon for security

addons/udes_security/controllers/main.py
PeterAlabaster: yep that works, nice to null out the or condition in the first block, didn't see that one :)

addons/udes_security/controllers/main.py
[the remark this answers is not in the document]
PeterAlabaster: didn't account for postgres caching..

re-ran on a clean session and was
 Planning Time: 0.201 ms
 Execution Time: 0.075 ms

still faster though, and now we aren't running it every parent menuitem

addons/udes_security/tests/test_desktop_access.py
PeterC10: Ignore me, I can't read.

PeterC10: Thought the line was duplicated and making a hilarious joke. :)

addons/udes_security/controllers/main.py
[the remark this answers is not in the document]
PeterAlabaster: `user.groups_id.ids` runs this query
```sql
SELECT
    res_groups_users_rel.uid,
    res_groups_users_rel.gid
FROM
    res_groups_users_rel,
    "res_groups"
    LEFT JOIN "ir_translation" AS "res_groups__name" ON (
        "res_groups"."id" = "res_groups__name"."res_id"
        AND "res_groups__name"."type" = 'model'
        AND "res_groups__name"."name" = 'res.groups,name'
        AND "res_groups__name"."lang" = 'en_GB'
        AND "res_groups__name"."value" != ''
    )
WHERE
    1 = 1
    AND res_groups_users_rel.uid IN (2)
    AND res_groups_users_rel.gid = res_groups.id
ORDER BY
    COALESCE("res_groups__name"."value", "res_groups"."name") OFFSET 0
```
Planning Time: 1.149 ms
Execution Time: 0.276 ms

on a clean postgres session (to avoid caching), can do this which gives same result:
```sql
SELECT ARRAY_AGG(gid) FROM res_groups_users_rel WHERE res_groups_users_rel.uid in (2)
```
Planning Time: 0.201 ms
Execution Time: 0.075 ms

Changed to run once outside the recursive function

addons/udes_security/controllers/main.py
PeterAlabaster: it can come from the user, as it's whatever is against `action=` in the URL bar, but that's why I added the type check above (because odoo core doesn't even type check it)

addons/udes_security/controllers/main.py
PeterAlabaster: just for ref, this is what i originally had the function as.. not confusing at all
```python
        parent = menu.parent_id
        if action and action.groups_id:
            action_access = any(g in user.groups_id.ids for g in action.groups_id.ids)
            if action_access:
                if parent:
                    return self._user_has_access_right(action, parent, user)
                else:
                    return True
            else:
                return False
        if menu and menu.groups_id:
            menu_access = any(g in user.groups_id.ids for g in menu.groups_id.ids)
            if menu_access:
                if parent:
                    return self._user_has_access_right(action, parent, user)
                else:
                    return True
            else:
                return False
        if menu and parent:
            return self._user_has_access_right(action, parent, user)

        return True
```

addons/udes_security/controllers/main.py
kdwyer: Actually on balance I prefer your original approach, as that will blow up if action_id isn't an int.  Best err on the side of safety with this sort of thing.

armandcela (approved):

PeterC10 (approved):

addons/udes_security/controllers/main.py
```diff
@@ -5,12 +5,115 @@
...
+        This ensures that the user has access at least every menuitem in the chain of menuitems
+        leading to the menu/action they are trying to access, following the same logic of the web UI.
+
+        :param: action: ir.actions.act_window(x,) recordset
+        :param: menu: ir.ui.menu(x,) recordset
+        :param: user_group_ids: list() of res.group ids for the user
+
+        :return: bool() whether or not the user should be granted access
+        """
+        # TODO(palabaster): This line makes up 25% of this entire functions cpu time, any way we can improve?
```
PeterAlabaster, line 100: highlighting this, would be good to improve but not necessary

addons/udes_security/tests/test_desktop_access.py
```diff
@@ -104,3 +98,140 @@ def test_all_users_can_access_database(self):
...
+                "has_access": False,
+            },
+            {
+                "user": self.user_2,
+                "action": third_party_apps_action,
+                "has_access": False,
+            },
+        ]
+        access_error_text = "odoo.exceptions.AccessError"
+        for user_action_access_dict in user_action_access_list:
```
LoKKeR96, line 224: Nit: use subTest? It would be nice but I won't block the review if you don't implement it.

LoKKeR96 (approved):

addons/udes_security/tests/test_desktop_access.py
PeterAlabaster: sorry not sure what you're asking - I'm using subTest a few lines later

LoKKeR96: Oh, how did I not see it? 😆

PeterAlabaster: updated it to have a msg either way which is helpful if failures happen

PeterAlabaster:
Ignore the mess above, accidentally pushed 11.0 and 14.0 changes to same branch, will open a new MR on `story/2417_14` branch

PR #687, opened 2022-07-22
Story/2417 11

PeterAlabaster:


addons/udes_security/tests/test_desktop_access.py
```diff
@@ -0,0 +1,201 @@
...
+    def setUp(self):
+        """
+        Use test cursor for environment
+        Create two users for testing and set their passwords so logging in
+        from controllers can be simulated
+        """
+        super().setUp()
+
+        # Use default test cursor for default environment
+        def restore(cr=self.cr, env=self.env):
```
PeterC10, line 18: Is it worth making a new class for both `DesktopAccessBase` and `TestFileControllers` to inherit?

As this will likely be needed if we need to test controllers elsewhere for security fun.

Maybe `ControllerBase` that has the setup simply include this method?

PeterC10: And the `release` method I guess.

Happy for you to make the call though. :)

PeterC10 (approved):

addons/udes_security/tests/test_desktop_access.py
PeterAlabaster: Yeah you're right, i'll try it out

ws-mps (approved):

PR #688, opened 2022-07-22
Story/2417 14

PeterAlabaster:


PeterC10 (approved):

ws-mps (approved):

adamrpatrick (approved):

addons/udes_security/controllers/main.py
```diff
@@ -5,12 +5,135 @@
...
+        request.env.cr.execute(query, (action_id,))
+        result = request.env.cr.fetchone()
+        if not result:
+            return [], []
+        (parent_path, relation_table, inverse, action_id) = result
+        # Get ids in order from parent --> child of the menus
+        parent_path_ids = [int(str_menu_id) for str_menu_id in parent_path.split("/")[:-1]]
+        action_group_ids = []
+        if relation_table and inverse:
+            action_group_ids = self._get_object_group_ids(action_id, relation_table, inverse)
```
jrbirch99, line 97: To me it feels like the `if relation_table and inverse` check should be in the function `_get_object_group_ids()` otherwise I guess we would need to add this check before every function call?

PeterAlabaster: Maybe, however the function won't do anything without them (it's not possible to find the `group_ids` in sql if we don't have this info). If they were kwargs i'd say yeah but because the function is pointless without them, I think they are best left here.

This is a bit of an edge case handler, where we have an action that isn't bound to a menuitem (it can happen on certain actions) and is really only ever going to be hit from this code which is specifically checking if we couldn't find the reltable/inverse based on the action->menu

addons/udes_security/controllers/main.py
```diff
@@ -87,17 +96,36 @@ def _get_menu_info_from_action_id(self, action_id):
...
+                "ir.actions.act_window",
+                "ir.actions.server",
+                "ir.actions.client",
+                "ir.actions.report",
+                "ir.actions.act_url",
+            ]
+            for model in models:
+                action = request.env[model].search([("id", "=", action_id)])
+                if action:
+                    action_group_ids = action.groups_id.ids
```
PeterC10, line 119: Q: Can we try disabling prefetching here? I think `groups_id.ids` might go and collect a load of fields we don't need or want.

addons/udes_security/controllers/main.py
```diff
@@ -87,17 +96,36 @@ def _get_menu_info_from_action_id(self, action_id):
...
+            # This will happen rarely, so not concerned about optimising.
+            models = [
+                "ir.actions.act_window",
+                "ir.actions.server",
+                "ir.actions.client",
+                "ir.actions.report",
+                "ir.actions.act_url",
+            ]
+            for model in models:
+                action = request.env[model].search([("id", "=", action_id)])
```
PeterC10, line 117: Suggestion: Add `order=id` since ordering doesn't matter.

PeterC10: Although follow on question... is it possible to have two `ir.actions.act_window` records with the same action but different groups?

addons/udes_security/controllers/main.py
PeterAlabaster: testing menuitem after a restart and re-log in, it appears to be slightly faster with your suggestion
Settings/Technical/Channels
without prefetch change
```action = request.env[model].search([("id", "=", action_id)])```
Hook took 0.008173465728759766
with prefetch change
```action = request.env[model].with_context(prefetch_fields=False).search([("id", "=", action_id)])```
Hook took 0.006898164749145508

addons/udes_security/controllers/main.py
PeterAlabaster: Users --> Change password wizard
without prefetch change
```action = request.env[model].search([("id", "=", action_id)])```
Hook took 0.010319232940673828
with prefetch change
```action = request.env[model].with_context(prefetch_fields=False).search([("id", "=", action_id)])```
Hook took 0.01175546646118164
with order=id change too
```action = request.env[model].with_context(prefetch_fields=False).search([("id", "=", action_id)], order="id")```
Hook took 0.010631799697875977

PeterC10: Yep let's not worry about prefetching then.

Just suggest to add order by ID as this will be simpler than the ordering set on the models.

PeterC10: Seems like it takes longer to override the default search? Anyway these are very small timings, all good will resolve.

PeterAlabaster: Added order="id" and also a check for groups_id field existing on the line below (in 11 and 14) which i forgot.

PR #689, opened 2022-07-25
[FIX] udes_open: Stop resetting priorities on pickings

caleb-shelton:
Stops done/cancelled pickings having their priorities
set to "Normal"

User-story: 2433

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -823,10 +823,7 @@ class StockPicking(models.Model):
     def action_cancel(self):
         """Extend to set date_done field when cancelling transfers"""
         res = super().action_cancel()
-
-        # To be reviewed later, as our priority approach is different. Having here just to remember
-        # that when we made changes we need to change _action_done method as well on stock module
-        self.write({"date_done": fields.Datetime.now(), "priority": "0"})
+        self.write({"date_done": fields.Datetime.now()})
         return res
 
     def validate_picking(self, create_backorder=False, force_validate=False):
--- a/addons/udes_stock/tests/test_stock_picking.py
+++ b/addons/udes_stock/tests/test_stock_picking.py
@@ -1759,3 +1759,58 @@ class TestStockPickingValidatePicking(common.BaseUDES):
         self.assertEqual(res_pick.move_line_ids, expected_backorder_mls)
         # Original picking in state done
         self.assertEqual(pick.state, "done")
+
+
+class TestStockPickingPriorities(common.BaseUDES):
+    @classmethod
+    def setUpClass(cls):
+        super(TestStockPickingPriorities, cls).setUpClass()
+        cls.create_quant(cls.fig.id, cls.test_stock_location_02.id, 100)
+
+        # first pick
+        cls.pick = cls.create_picking(
+            picking_type=cls.picking_type_pick,
+            products_info=[{"product": cls.fig, "uom_qty": 10}],
+            location_id=cls.test_stock_location_02.id,
+            confirm=True,
+            assign=True,
+            priority="0",
+        )
+
+        # second pick
+        cls.pick2 = cls.create_picking(
+            picking_type=cls.picking_type_pick,
+            products_info=[{"product": cls.fig, "uom_qty": 10}],
+            location_id=cls.test_stock_location_02.id,
+            confirm=True,
+            assign=True,
+            priority="1",
+        )
+
+    def test_priority_zero_remains_on_done_picking(self):
+        """
+        Test that priority 0 is not overwritten on done pickings.
+        """
+        self.complete_picking(self.pick)
+        self.assertEqual(self.pick.priority, "0")
+
+    def test_priority_zero_remains_on_cancelled_picking(self):
+        """
+        Test that priority 0 is not overwritten on cancelled pickings.
+        """
+        self.complete_picking(self.pick)
+        self.assertEqual(self.pick.priority, "0")
+
+    def test_priority_one_remains_on_done_picking(self):
+        """
+        Test that priority 1 is not overwritten on done pickings.
+        """
+        self.complete_picking(self.pick2)
+        self.assertEqual(self.pick2.priority, "1")
+
+    def test_priority_one_remains_on_cancelled_picking(self):
+        """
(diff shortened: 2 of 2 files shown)


kdwyer (commented):
Do we want a unit test for this?

PeterAlabaster (approved):
LGTM, I second kevins comment

addons/udes_stock/models/stock_picking.py
```diff
@@ -826,7 +826,7 @@ def action_cancel(self):
 
         # To be reviewed later, as our priority approach is different. Having here just to remember
```
PeterC10, line 827: Can we remove these comments please, as no longer relevant.

PeterC10:
The first line should have the module name, not the repo. Please can you also reword the text slightly, as this commit doesn't do anything for done pickings.

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1759,3 +1759,34 @@ def test_multi_users_enabled(self):
...
+        """
+        self.create_quant(self.fig.id, self.test_stock_location_02.id, 100)
+        pick = self.create_picking(
+            picking_type=self.picking_type_pick,
+            products_info=[{"product": self.fig, "uom_qty": 10}],
+            location_id=self.test_stock_location_02.id,
+            priority="1",
+        )
+        pick.move_lines[0].quantity_done = 10
+        pick._action_done()
```
PeterC10, line 1777: Q: Is the picking in a done state after calling this? Doesn't it need to be confirmed and assigned stock first?

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1759,3 +1759,34 @@ def test_multi_users_enabled(self):
...
         # Original picking in state done
         self.assertEqual(pick.state, "done")
+
+
+class TestStockPickingPriorities(common.BaseUDES):
+    def test_priority_remains_on_done_picking(self):
+        """
+        Test that priorities are not overwritten on done pickings.
+        """
+        self.create_quant(self.fig.id, self.test_stock_location_02.id, 100)
```
PeterC10, line 1769: You've got a lot of repeated code here, as you basically have the same picking it's just that in one test you complete it and the other you cancel.

Let's move the code that creates the quant and picking into `setUpClass`, that will cut down the tests a fair bit.

addons/udes_stock/tests/test_stock_picking.py
caleb-shelton: Through my testing, this is enough to trigger the code that would reset the priority (prior to the fix)

caleb-shelton: (So without the fix, this UT would fail as the priority would be reset)

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1759,3 +1759,34 @@ def test_multi_users_enabled(self):
...
+        pick = self.create_picking(
+            picking_type=self.picking_type_pick,
+            products_info=[{"product": self.fig, "uom_qty": 10}],
+            location_id=self.test_stock_location_02.id,
+            priority="1",
```
kdwyer, line 1774: There's a case for testing both `priority="1"` _and _ `priority="0"`, as otherwise the tests pass if we just hardcode the priority to "1" in the implementation.  What do you think @PeterC10 ?

addons/udes_stock/tests/test_stock_picking.py
PeterC10: Can you verify that the picking is in a done state? Can either assert it or just debug locally. We should keep our tests as realistic as possible.

caleb-shelton: I can confirm it puts it into state 'done'

addons/udes_stock/tests/test_stock_picking.py
caleb-shelton: I may be misunderstand this, but the problem was it was resetting the priority back to 0. Not the other way around, so not sure how testing if priority is still 0 is beneficial?

PeterC10: I see your point... gets tricky because by default Odoo sets it to zero anyway.

Better to be safe though, so yes let's have a picking for each priority, and assert that it doesn't change when completed or cancelled.

Will be more useful in future when we implement additional priorities.

kdwyer: Then we can be confident that the system is behaving as expected in both cases.

addons/udes_stock/tests/test_stock_picking.py
PeterC10: Can you confirm what the state was before calling `_action_done`?

Just not sure how the pick has move lines unless it has already been assigned stock, but from the earlier lines I can't see where this is being done.

PeterC10:
Note: CI is failing as it is using an older version of the Odoo branch (where the change for not resetting priority on done pickings was made). We can get this sorted once the story has been signed off and merged.

addons/udes_stock/tests/test_stock_picking.py
caleb-shelton: State draft.

PeterC10: Ah I think I see another problem, you are setting the quantity done on the `stock.move` records, not `stock.move.line` records.

`move_lines` are moves  and `move_line_ids` are move lines, bad naming from Odoo but it is what it is.

So let's do this, make the picking confirmed and assigned in `create_picking`.

Remove the code where you are setting the quantity done and calling `_action_don`, and instead call `complete_picking`, this helper method will set the quantity done of all move lines properly and then call action done if validate is true (the default).

caleb-shelton: Pushed changes. Not sure I am confirming and assigning before setting priority, as they are being passed in the same call to create_picking, just in that order... is that what you meant?

PeterC10: Yeah I just meant the ordering, as priority is part of `kwargs` whereas `confirm` and `assign` are args specified in the method.

But looks good, thanks. :)

PeterC10 (approved):

PR #698, opened 2022-08-03
[IMP] udes_stock: Implement delivery slip for multiple source documents

LoKKeR96:
Add source document to each stock move line. Rename table headings and
add picking reference number.

Story: x2422

### DIFF
--- a/addons/udes_stock/report/deliveryslip_templates.xml
+++ b/addons/udes_stock/report/deliveryslip_templates.xml
@@ -7,147 +7,152 @@
 
             <t t-call="udes_stock.external_layout">
 
-                <t t-set="o" t-value="o.with_context({'lang':o.partner_id.lang})" />
+                <t t-set="o" t-value="o.with_context({'lang':o.partner_id.lang})"/>
 
-                    <t t-if="o.state != 'done'">
-                        <h1>
-                            DRAFT
-                        </h1>
-                    </t>
+                <t t-if="o.state != 'done'">
+                    <h1>
+                        DRAFT
+                    </h1>
+                </t>
 
-                    <h2>
-                        Delivery Slip
-                    </h2>
+                <h2>
+                    Delivery Slip
+                </h2>
 
-                    <!-- To avoid to break sale_stock -->
-                    <div name="customer_address" />
-                    <div name="div_sched_date" />
+                <!-- To avoid to break sale_stock -->
+                <div name="customer_address"/>
+                <div name="div_sched_date"/>
 
-                    <table class="table table-condensed">
-                        <thead>
-                            <tr>
-                                <th><strong>Trailer number</strong></th>
-                                <th><strong>Trailer unit ID</strong></th>
-                                <th><strong>Vehicle registration</strong></th>
-                                <th><strong>Driver name</strong></th>
-                            </tr>
-                        </thead>
-                        <tbody>
-                            <tr>
-                                <td>
-                                    <span t-field="o.u_vehicle_sequence"/>
-                                </td>
-                                <td>
-                                    <span t-field="o.u_vehicle_description"/>
-                                </td>
-                                <td>
-                                    <span t-field="o.u_vehicle_registration"/>
-                                </td>
-                                <td>
-                                    <span t-field="o.u_driver_name"/>
-                                </td>
-                            </tr>
-
-                            <t t-if="o.partner_id">
-                                <tr>
-                                    <th colspan="2" style="padding-top: 40px" />
-                                    <th colspan="2" style="padding-top: 40px"><strong>Customer address</strong></th>
-                                </tr>
-                                <tr>
-                                    <td colspan="2" />
-                                    <td colspan="2">
-                                        <span t-field="o.partner_id"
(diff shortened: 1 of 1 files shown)


addons/udes_stock/report/deliveryslip_templates.xml
```diff
@@ -47,69 +62,37 @@
...
-                                  <th style="padding-top: 40px"><strong>Ordered Quantity</strong></th>
-                                  <th style="padding-top: 40px"><strong>Sent Quantity</strong></th>
+                                  <th style="padding-top: 40px"><strong>Ordered Qty.</strong></th>
+                                  <th style="padding-top: 40px"><strong>Sent Qty.</strong></th>
                             </tr>
                         </thead>
                         <tbody>    
                             <tr t-if="not o.move_line_ids" t-foreach="o.move_lines" t-as="move">
-                                <td></td>
+                                <td><span t-field="move.origin"/></td>
```
jrbirch99, line 84: Do we need the pallet number here too?

LoKKeR96: I suspect we want it in case we don't have move lines yet but only stock moves. It seems like a safe assumption. That way if you try to print one you get which order has requested what which is good in my view.

jrbirch99: Not sure if I understood what you meant, but I think if we have no move_lines this should read: 
```suggestion
                                <td><span t-field="move.origin"/></td>
                                <td></td>
```
We need the empty field for the empty pallet number, right?

LoKKeR96: I don't think we need it no, if we have only stock moves then we are displaying stock moves which cannot have pallet so it's redundant. Basically, if your picking is in waiting and you print the delivery slip then you will get this delivery slip which contains the stock moves. Adding a pallet field for this use case doesn't seem necessary in my view.

LoKKeR96: Oh, I see what you're saying. Sorry it wasn't clear. Let me check how it gets displayed.

jrbirch99: sorry I wasn't being very clear :smile:

LoKKeR96: Yep, it was needed, thank you! I added the field now.

jrbirch99 (approved):

addons/udes_stock/report/deliveryslip_templates.xml
```diff
@@ -47,69 +62,38 @@
...
-                                </td>
-                                <td colspan="2" name="td_sched_date">
-                                   <t t-if="o.state == 'done'">
-                                        <span t-field="o.date_done"/>
-                                   </t>
-                                   <t t-if="o.state != 'done'">
-                                        <span t-field="o.scheduled_date"/>
-                                   </t>
-                                </td>
-                            </tr>
```
LoKKeR96, line 87: I tried in both 11 and 14 and I cannot figure out where this is used. I am removing as it is not on the reference document for the story.

adamrpatrick: Assuming that `o` is a `stock.picking`, it should be adding things like the partner info (e.g. address). TBH I would still leave this in unless you're 100% sure it's not used and PO agree

addons/udes_stock/report/deliveryslip_templates.xml
```diff
@@ -7,147 +7,139 @@
...
-                                    <th colspan="2" style="padding-top: 40px"><strong>Customer address</strong></th>
-                                </tr>
-                                <tr>
-                                    <td colspan="2" />
-                                    <td colspan="2">
-                                        <span t-field="o.partner_id"
-                                           t-options='{"widget": "contact", "fields": ["address", "name", "phone"], "no_marker": True}'/>
-                                    </td>
-                                </tr>
-                            </t>
```
LoKKeR96, line 63: This has been removed after a discussion with PO. It is not needed in delivery slips.

addons/udes_stock/report/deliveryslip_templates.xml
```diff
@@ -7,147 +7,139 @@
...
+                                    <span t-esc="o.date_done.strftime('%d-%m-%Y %H:%M')"/>
+                                </t>
+                                <t t-if="o.state != 'done'">
+                                    <span t-esc="o.scheduled_date.strftime('%d-%m-%Y %H:%M')"/>
+                                </t>
+                                </span>
+                            </td>
+                        </tr>
+                    </tbody>
+                </table>
```
LoKKeR96, line 46: This is the only change I made in the new commit so that picking reference name and date are displayed correctly as per reference document on the story.

addons/udes_stock/report/deliveryslip_templates.xml
LoKKeR96: I have updated this to only present the picking name and the date as per discussions with PO.

adamrpatrick (approved):

PR #702, opened 2022-08-08
[ADD] udes_stock: Checking in backend date fields

armandcela:
Entered year should be more than year 1000
Story: 2390

### DIFF
--- a/addons/udes_stock/models/models.py
+++ b/addons/udes_stock/models/models.py
@@ -2,6 +2,9 @@
 
 from odoo import models, api, _
 from lxml import etree
+from odoo.tools import DEFAULT_SERVER_DATE_FORMAT as DATE_FORMAT
+from datetime import datetime
+from odoo.exceptions import UserError
 
 
 class Base(models.AbstractModel):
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/models.py
```diff
@@ -45,3 +48,48 @@ def base_model_detailed_view(self, model, form_view):
...
+                if isinstance(value, str) and len(value) >= 10:
+                    date_field = datetime.strptime(value[:10], DATE_FORMAT)
+                    if date_field.year < 1000:
+                        raise UserError(
+                            _("Year of %s is less than 1000") % value
+                        )
+
+    def get_date_or_datetime_fields(self):
+        """Finding date or datetime fields of current model"""
+        return {
```
PeterC10, line 93: Suggestion: Don't include `create_date` and `write_date` as these will be set by the system, and we can minimise any potential performance hits.

PeterC10:
Can we get a unit test for this change? Can use the `scheduled_date` on `stock.picking` for example.

armandcela:
> Can we get a unit test for this change? Can use the `scheduled_date` on `stock.picking` for example.

Yes sure. I think we have agreed to add in story when Unit test is needed, but happy to add a unit test

PeterAlabaster (approved):

addons/udes_stock/models/models.py
```diff
@@ -45,3 +48,48 @@ def base_model_detailed_view(self, model, form_view):
...
+        return res
+
+    @api.model
+    def validate_values_of_date_fields(self, values, date_fields):
+        """
+        Checking if date or datetime fields that are formatted as string
+        can be converted to a date field and the year is greater than 1000.
+        Raising error in case year is less than 1000
+        """
+        for key, value in values.items():
```
PeterC10, line 82: Suggestion: `date_values = {k: v for k, v in values.items() if k in date_fields and isinstance(v, str) and len(v) >= 10}`

Then can just go through the values in `date_values`.

PeterC10:
> > Can we get a unit test for this change? Can use the `scheduled_date` on `stock.picking` for example.
> 
> Yes sure. I think we have agreed to add in story when Unit test is needed, but happy to add a unit test

It is a process issue atm, I will try and get it so that unit tests are required by default. :)

addons/udes_stock/models/models.py
```diff
@@ -45,3 +48,48 @@ def base_model_detailed_view(self, model, form_view):
...
+            if key in date_fields:
+                if isinstance(value, str) and len(value) >= 10:
+                    date_field = datetime.strptime(value[:10], DATE_FORMAT)
+                    if date_field.year < 1000:
+                        raise UserError(
+                            _("Year of %s is less than 1000") % value
```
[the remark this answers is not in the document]
PeterC10: Lol I can't read or type today sorry Armand, should just be:

`"Date '%s' is not valid."` (drop the a).

PeterC10 (approved):

PR #703, opened 2022-08-09
[ADD] udes_common: Prevent dates in backend when year is less than 1000

armandcela:
Story: 2390


PeterC10:
Can we add a unit test? Would need a model that is available in the base module I guess... maybe `res.partner`, if that has a `date` or `datetime` field?

PeterAlabaster (approved):

PeterC10 (approved):

PR #704, opened 2022-08-09
[IMP] udes_security: Port of redirection domain filter list

LoKKeR96:
Porting of the code that implements the allowlist domain filter model
for blocking redirections from UDES13. This code implements a list to
filter domains and block redirection requests if the domain is not in
the list filter. The list has been renamed from whitelist to allowlist.

Story: x2536

### DIFF
--- a/addons/udes_security/controllers/redirect.py
+++ b/addons/udes_security/controllers/redirect.py
@@ -1,9 +1,70 @@
 import logging
-from odoo import http
+from odoo import http, _
 from odoo.addons.web.controllers import main
 from odoo.http import request
+from werkzeug import urls, url_unquote
+from werkzeug.exceptions import BadRequest
+from fnmatch import fnmatch
 
 _logger = logging.getLogger(__name__)
+
+
+class no_absolute_redirect(object):
+    """Decorator to mark a route as forbidding absolute redirects
+
+    If the decorated route is called with a "redirect" query parameter
+    containing an absolute URL, the request will be terminated with an
+    exception.
+    """
+
+    def __init__(self, func):
+        self.func = func
+
+    @property
+    def __name__(self):
+        """Copy original method name
+
+        This is required since the method name will be introspected as
+        part of calculating the controller inheritance structure.
+        """
+        return self.func.__name__
+
+    def __call__(self, *args, **kwargs):
+        """Call method"""
+        redirect = kwargs.get("redirect")
+
+        if redirect:
+            redirect = url_unquote(redirect)
+            if redirect.startswith("//"):
+                # remove leading /'s which can trick the url parser but is handled by browsers
+                kwargs["redirect"] = "/" + redirect.lstrip("/")
+            else:
+                parsed_url = urls.url_parse(redirect)
+
+                if self.is_absolute_url(parsed_url) and not self.domain_is_allowlisted(parsed_url):
+                    _logger.error("Invalid redirection attempted: %s", redirect)
+                    raise BadRequest(_("Invalid redirection attempted"))
+
+        return self.func(*args, **kwargs)
+
+    @staticmethod
+    def is_absolute_url(parsed_url):
+        """Returns True if URL is an absolute URL, else False"""
+        return True if bool(parsed_url.scheme) or bool(parsed_url.netloc) else False
+
+    @classmethod
+    def domain_is_allowlisted(cls, parsed_url):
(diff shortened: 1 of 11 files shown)


addons/udes_security/controllers/redirect.py
```diff
@@ -1,11 +1,72 @@
...
+
+                if self.is_absolute_url(parsed_url) and not self.domain_is_allowlisted(parsed_url):
+                    _logger.error("Invalid redirection attempted: %s", redirect)
+                    raise BadRequest(_("Invalid redirection attempted"))
+
+        return self.func(*args, **kwargs)
+
+    @staticmethod
+    def is_absolute_url(parsed_url):
+        """Test if a URL is an absolute URL"""
```
PeterC10, line 52: Super nit: Remove the term `test` from the doc-string here and for `domain_is_allowlisted`.

E.g. `Returns True if URL is an absolute URL, else False` or similar.

addons/udes_security/data/res_users.xml
```diff
@@ -13,6 +13,7 @@
                 (4, ref('udes_security.group_can_add_trusted_user')),
                 (4, ref('udes_security.group_trusted_user')),
                 (4, ref('udes_security.group_edit_allowed_file_types')),
+                (4, ref('udes_security.group_edit_allowlist_redirect_domains')),
```
PeterC10, line 16: Can we add this to `user_root` as well please? Technically doesn't need them, but good to keep it consistent. :)

addons/udes_security/tests/test_no_redirect.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+    def test_url_encoded_double_forward_slash_redirects_locally(self):
+        self.authenticate("admin", "admin")
+        self.assertNotFound(
+            self.url_open("/web?redirect=%2F%2Fbaddomain.com"),
+            redirect_url=self._get_url("/baddomain.com"),
+        )
+
+    def test_white_listed_domain_redirects(self):
```
PeterC10, line 84: Supernit: `white` -> `allow`

addons/udes_security/tests/test_no_redirect.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self.assertFailedRedirect(self.url_open("/web?redirect=http://google.com"))
+
+        DomainAllowlist.create({"name": "Google", "domain": "google.com"})
+        self.assertSuccessfulRedirect(
+            self.url_open("/web?redirect=http://google.com"),
+            redirect_url="http://google.com",
+        )
+
+    def test_white_listed_wildcarded_domain_redirects(self):
```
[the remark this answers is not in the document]
LoKKeR96: Can't trust the Ctrl+F Search anymore :) Good spot.

addons/udes_security/views/domain_allowlist_view.xml
```diff
@@ -0,0 +1,67 @@
...
+        </tree>
+      </field>
+    </record>
+
+    <!-- Search filter -->
+    <record id="domain_allowlist_search" model="ir.ui.view">
+      <field name="name">udes_security.domain.allowlist.search</field>
+      <field name="model">udes_security.domain.allowlist</field>
+      <field name="arch" type="xml">
+        <search string="Allowlist Redirect Domains">
```
PeterC10, line 47: Suggestion: Add a filter for archived records.

PeterC10 (approved):

PeterAlabaster (approved):

addons/udes_security/data/res_groups.xml
```diff
@@ -41,4 +41,13 @@
         <field name="u_required_group_id_to_change" ref="base.group_system" />
     </record>
 
+    <record id="group_edit_allowlist_redirect_domains" model="res.groups">
+        <field name="name">Manage Allowed Redirect Domains</field>
+        <field name="category_id" ref="base.module_category_extra" />
+    </record>
+    <!-- Cannot self reference, so we extend the record after we create it -->
+    <record id="udes_security.group_edit_allowlist_redirect_domains" model="res.groups">
```
armandcela, line 49: Do we need to add the group in any of actual user templates, in order that users applied the specific template will have the right to modify allowed list.
I think is more a PO question

armandcela: Got the response i think is needed only for admin :)

armandcela (approved):

PR #705, opened 2022-08-10
[ADD] udes_stock: Move (and partial) cancellation propagation

PeterAlabaster:
This includes a new option `u_propagate_cancel` on `stock.picking.type`
to turn the functionality on.

User-story: 2411

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -28,11 +28,24 @@ class StockMove(models.Model):
         return super().create(vals)
 
     def _action_cancel(self):
+        """
+        Extend _action_cancel to;
+        - Set move_line_ids to have a qty_done of 0
+        - Include a context bypass in the super call to prevent initial demand re-setting
+        - Include a hook to propagate cancellations when the picking
+          type is configured with u_propagate_cancel set to True
+        """
         self.move_line_ids.write({"qty_done": 0})
-
-        return super(
+        # Pull out old move_dest_ids as these _can_ be removed upon cancellation
+        old_dest_ids = {m.id: m.move_dest_ids for m in self}
+        res = super(
             StockMove, self.with_context({"bypass_set_qty_to_initial_demand": True})
         )._action_cancel()
+        for move in self:
+            if move.picking_type_id.u_propagate_cancel:
+                move.propagate_cancellation_to_next_picks(old_dest_ids.get(move.id))
+
+        return res
 
     def _do_unreserve(self):
         """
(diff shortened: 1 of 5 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -118,6 +124,47 @@ def _unreserve_initial_demand(self, new_move):
...
+        """
+        StockMove = self.env["stock.move"].browse()
+        next_moves_filter = lambda m: m.picking_type_id.u_propagate_cancel == True
+        next_moves = self.move_dest_ids.filtered(next_moves_filter)
+        if not next_moves:
+            # NB: Can this ever hit a scenario where cancelling caused
+            # refactoring and thus the old dest ids to be invalid?
+            next_moves = old_dest_ids
+        qty = self.product_uom_qty
+        # NB: This is partly dealing with a symptom rather than a cause.
```
PeterC10, line 147: Let's add the issue reference once we've made it. :)

addons/udes_stock/models/stock_move.py
```diff
@@ -29,10 +29,16 @@ def create(self, vals):
 
     def _action_cancel(self):
```
PeterC10, line 30: Not linked to your story directly but can we add a doc-string here?

PeterC10 (approved):

addons/udes_stock/models/stock_move.py
```diff
@@ -118,6 +131,42 @@ def _unreserve_initial_demand(self, new_move):
...
+                next_moves = StockMove
+            else:
+                next_moves_todo = next_moves
+                next_moves = StockMove  # Clear this as we potentially rebuild inside the loop
+                for move in next_moves_todo:  # Iterate on the non-cleared set
+                    if move.product_uom_qty > qty:
+                        move.product_uom_qty -= qty
+                        # Recursing inside here is only necessary for part cancellations
+                        next_moves |= move.move_dest_ids.filtered(next_moves_filter)
+                    else:  # qty is less, as there was no fully_matched_move
```
jrbirch99, line 167: This else statement is for when `move.product_uom_qty < qty` right? so I think the comment should read: "qty is greater"

addons/udes_stock/models/stock_move.py
```diff
@@ -118,6 +131,42 @@ def _unreserve_initial_demand(self, new_move):
...
+
+        :param: old_dest_ids: stock.move(x) recordset(s), fallback if no move_dest_ids in self.
+                              Can be emptyset
+        """
+        StockMove = self.env["stock.move"].browse()
+        next_moves_filter = lambda m: m.picking_type_id.u_propagate_cancel == True
+        next_moves = self.move_dest_ids.filtered(next_moves_filter)
+        if not next_moves:
+            # NB: This handles a symptom of Issue #2771
+            next_moves = old_dest_ids
```
jrbirch99, line 151: Q: what is the expected behaviour when we have a picking_type down the chain with `u_propagate_cancel =  False`? Will we still expect that picking_types moves to be cancelled?

addons/udes_stock/tests/test_stock_move.py
```diff
@@ -493,6 +495,179 @@ def test_get_uncovered_moves_returns_move_if_only_partially_covered(self):
...
+        cls.quant1 = cls.create_quant(cls.banana.id, cls.test_stock_location_01.id, 5)
+        cls.quant2 = cls.create_quant(cls.banana.id, cls.test_stock_location_02.id, 3)
+        cls.pick = cls.create_picking(
+            cls.picking_type_pick, products_info=cls._pick_info, assign=True
+        )
+        cls.pick_ml1 = cls.pick.move_line_ids.filtered(lambda ml: ml.product_qty == 5)
+        cls.pick_ml2 = cls.pick.move_line_ids.filtered(lambda ml: ml.product_qty == 1)
+
+    @classmethod
+    def create_trailer_route(cls):
```
jrbirch99, line 521: I think this should be added to `UnconfiguredBaseUdes` as it seems quite general

addons/udes_stock/models/stock_move.py
PeterAlabaster: Yeah i think being more explicit is good
```python
else:  # move qty is less, as there was no fully_matched_move
```

jrbirch99:
I like the unit tests

addons/udes_stock/models/stock_move.py
PeterAlabaster: Yeah, thinking about it I think i have the `.filtered(next_moves_filter)` in the wrong places. I'll review and push an update

Having `u_propagate_cancel` set to True on the pick type being cancelled will cancel the next moves, but if those next moves have `u_propagate_cancel` set to False, the next moves of those will not be cancelled

addons/udes_stock/tests/test_stock_move.py
PeterAlabaster: It's specific to this case, if i add it to there it will break most unit tests as they are testing without trailer at the end

addons/udes_stock/models/stock_move.py
jrbirch99: Yeah cool, I just wanted to make sure I understood correctly :smile:

addons/udes_stock/tests/test_stock_move.py
jrbirch99: I think I came across something similar in a story. @PeterC10, maybe the routes in `UnconfiguredBaseUdes` need changing in the future , such that Trailer dispatch is included (so that the tests better reflect the actual routes in UDES)? should I create an issue for it?

addons/udes_stock/models/stock_move.py
PeterAlabaster: I've changed to only do the filter when we have part cancellations, and change `next_moves` inline in the while loop. This is because inside `_action_cancel()` there is a check, so for all other cases it doesn't need filtering as it only calls `propagate_cancellation_to_next_picks()` when those moves picking types have `u_propagate_cancel` set to True

addons/udes_stock/tests/test_stock_move.py
PeterC10: Yep make an issue and it can be discussed later.

jrbirch99 (approved):

PR #707, opened 2022-08-23
User-story: 2818

ws-mps:
Change stock.inventory exhausted field

Avoid potential issues when large number of products in the system:
* Hide check box from the form view
* When changing filter set to exhausted=False

### DIFF
--- a/addons/udes_stock/models/stock_inventory.py
+++ b/addons/udes_stock/models/stock_inventory.py
@@ -58,6 +58,11 @@ class StockInventory(models.Model):
         domain=_get_stock_check_users,
     )
 
+    @api.onchange("filter")
+    def onchange_filter(self):
+        self.exhausted = False
+        super().onchange_filter()
+
     @api.multi
     def action_done(self):
         """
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_inventory.py
```diff
@@ -58,6 +58,11 @@ def _get_stock_check_users(self):
         domain=_get_stock_check_users,
     )
 
+    @api.onchange("filter")
+    def onchange_filter(self):
+        self.exhausted = False
+        super().onchange_filter()
```
PeterC10, line 64: Will the super call not set `exhausted` back to `True` potentially?

PeterC10: Ah okay I think I get it, Odoo only sets exhausted to `True` if filtering is set to `product`, and that will only ever be 1 product right?

PeterC10 (approved):

addons/udes_stock/models/stock_inventory.py
armandcela: In fact i was thinking same thing, maybe safer to set exhausted to False after super.

ws-mps: Yep, only one product line will be created.

PeterC10: @armandcela I think we should leave it as it is. Because if we set `exhausted` after the `super` call, then it means that specifying one product only will not add a line by default in the Inventory Adjustment if it doesn't have stock in the location.

armandcela (approved):

PR #714, opened 2022-08-26
Story/757 - Prevent internal locations from having child locations

PeterC10:
Internal locations locked down to prevent being able to have child locations.

Updated various tests to work nicely with this new constraint.

Added and updated helper methods.

Added sudo parameter for get_or_create for stock mixin.

### DIFF
--- a/addons/udes_stock/models/stock_location.py
+++ b/addons/udes_stock/models/stock_location.py
@@ -2,6 +2,7 @@ import os.path
 from odoo import fields, models, api, _
 from .stock_picking_type import TARGET_STORAGE_FORMAT_OPTIONS
 from odoo.addons.udes_common.models.fields import PreciseDatetime
+from odoo.exceptions import ValidationError
 
 
 class StockLocation(models.Model):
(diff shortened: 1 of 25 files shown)


addons/udes_stock/tests/common.py
```diff
@@ -497,11 +511,25 @@ def create_partner(cls, name, **kwargs):
...
+        Validates the picking unless `validate` is set to False.
+
+        If the destination location on the move line isn't an internal location,
+        then it will be set to the first internal child of that location if possible
+        unless `set_dest_location` is set to False.
+        """
         for move_line in picking.move_line_ids:
             if move_line.state == "assigned":
                 move_line.qty_done = move_line.product_uom_qty
+                if move_line.location_dest_id.usage != "internal":
```
LoKKeR96, line 527: I can see how this works well now that we have blocked down the location structure but is it still a good idea doing this since we are forcing a default behaviour in the tests that is not done in UDES?

LoKKeR96 (commented):
Good job, LGTM, only 1 question before thumbs up for me.

addons/udes_stock/tests/common.py
PeterC10: I'm not sure why you mean by tests not done in UDES. Could you give me an example?

The reason this was needed was because previously view locations were being turned into internal locations, I assume to save time. Without this step the tests that use this helper method will error if using the default location of a picking type, which will almost always be a view location.

LoKKeR96: What I'm trying to say is that this helper method is making automatic a "process" than in UDES doesn't exist.

For example, let's say that we have a picking where the move lines have the done quantities equal to the requested quantities (hence they are done) but the destination location is not set to anything different that the default destination location of the picking (which is usually a location of type view), then when we try to validate the picking UDES will complain saying "You cannot take products from or deliver products to a location of type 'view'". This is because UDES does not automatically set the destination location to the first child of the default destination location. 

So, here in the tests it is not wrong to do anything like this which UDES doesn't do for you but effectively this routine is doing that automatically for all the tests. It's not wrong since it's an helper function and it's job is to make our life easier but I feel like we are hiding this (once explicit behaviour) from the code of each individual test. Again, it's not a real issue but more of a consideration that I'm trying to make.

PeterC10: It also validates the picking automatically, unless `validate` is set to False.

The point of the helper method is to speed up tests. For most tests the specific destination location isn't really important, so this is just saving time.

If it was important, the dev has the same option they have now. Manually setting the destination location on the move line(s) and then calling the helper method to complete the picking (because it only automatically selects a child if the type isn't internal already). And of course they can just set `set_dest_location` to False.

At the end of the day the helper methods are there to make our lives easier in terms of tests like you said. It's still the responsibility of the dev to make sure they understand what the helper methods are doing and whether it is appropriate to use it.

LoKKeR96: Make sense, if that's the case where most tests don't require this level of "hardening" on the requirements then I'm happy with this helper function and I'm happy that we discussed it.

LoKKeR96 (approved):

addons/udes_stock/models/stock_location.py
```diff
@@ -172,14 +173,46 @@ def set_u_heatmap_data_updated(self, vals):
...
+    @api.constrains("usage", "location_id", "child_ids")
+    def _restrict_internal_location_children(self):
+        """
+        Raises a ValidationError if any of the following conditions are met:
+            1. Location is a child of an internal location
+            2. Location is an internal location and has child locations
+        """
+        Location = self.env["stock.location"]
+        for location in self:
+            location = location.with_context(prefetch_fields=False)
```
ws-mps, line 185: Wouldn't `self = self.with_context(prefetch_fields=False)` achieve the same result with only one context change?

PeterC10: Changed.

ws-mps (approved):

PR #716, opened 2022-09-01
Story/949 remove simple location blocking and refs to it

PeterAlabaster:


armandcela:
Do we need to replace manifest files with udes_blocking_rules or udes_stock_blocked if moved the content to the other module.
And do we need to have existing check if added the dependency about u_blocked.
Maybe replacing simple blocking module with the new one on gitlab ci?

PeterC10 (approved):

ws-mps (approved):

armandcela (approved):

PR #721, opened 2022-09-12
Story/2911 - Refactoring with package links

PeterC10:
Take into account package links when refactoring pickings

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -870,9 +870,10 @@ class StockPicking(models.Model):
                 self.move_line_ids.mark_as_done()
             # Check if a backorder is required and can be created:
             # a) Multi users is enabled
-            # b) There are incomplete move lines and create backorder is allowed
+            # b) There are incomplete moves (ie partially available/picked)
+            #    and create backorder is allowed
             multi_users_enabled = self.picking_type_id.u_multi_users_enabled
-            if self.move_line_ids.get_lines_incomplete():
+            if self.move_lines.get_moves_with_quantity_todo():
                 if multi_users_enabled:
                     user = self.env.user
                     mls_to_keep = self.move_line_ids.filtered(
--- a/addons/udes_stock_refactoring/models/stock_picking.py
+++ b/addons/udes_stock_refactoring/models/stock_picking.py
@@ -1,6 +1,4 @@
-from odoo import api, fields, models, _
-from odoo.exceptions import ValidationError
-from collections import defaultdict
+from odoo import api, models
 import logging
 
 _logger = logging.getLogger(__name__)
@@ -13,7 +11,7 @@ class StockPicking(models.Model):
         self, group, picking_type, location, dest_location
     ):
         """
-        Return a dict of values for a new picking based on the supplied 
+        Return a dict of values for a new picking based on the supplied
         Procurement Group Picking Type, Location and Destination Location.
         """
         values = {
@@ -29,7 +27,7 @@ class StockPicking(models.Model):
         """
         If any of the fields in kwargs is set and its value is different
         than the new one, set the field value to False to avoid misleading values.
-        E.g. if picking.origin is 'ASN001' and kwargs contains origin with value 
+        E.g. if picking.origin is 'ASN001' and kwargs contains origin with value
         'ASN002', picking.origin is set to False.
         """
         self.ensure_one()
--- a/addons/udes_stock_refactoring/models/stock_picking.py
+++ b/addons/udes_stock_refactoring/models/stock_picking.py
@@ -41,6 +41,45 @@ class StockPicking(models.Model):
                 if current_value and current_value != value:
                     self[field] = False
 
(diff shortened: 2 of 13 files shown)


addons/udes_stock/models/stock_location.py
```diff
@@ -312,3 +312,11 @@ def get_available_stock_locations(self):
          their children) stock available for fulfilling orders. Should be
         overridden where necessary"""
         return self.env.ref("stock.stock_location_stock")
+
+    def get_supplied_location_is_child(self, location):
```
armandcela, line 316: Would suggest to transfer this in base models, we may need to check for user templates, products etc. Maybe the name should be more generic example `record_is_child(self, child) `

PeterC10: Good idea, not sure on the name though (I didn't like `get_supplied_location_is_child` either).

The problem with `record_is_child` is that it implies that we are just checking if the record in `self` is a child of something else. But instead we are checking that `self` is a parent of the supplied record.

Maybe `record_is_parent`?

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -43,6 +41,43 @@ def _remove_misleading_values(self, **kwargs):
                 if current_value and current_value != value:
                     self[field] = False
 
+    @api.model
+    def _setup_refactored_picking_package_levels(self, picking, move_lines):
```
armandcela, line 45: From what i see you are not using self, and you are on stock picking class. I would suggest to define method:
`def _setup_refactored_picking_package_levels(self, move_lines):`  -- where self is the picking, maybe can add ensure one on self and below when you call to call like: `picking._setup_refactored_picking_package_levels(move_lines)`

addons/udes_stock/models/stock_location.py
armandcela: I think you are better with names, definitely :)

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Good spot, problem with being lazy and copying `_new_picking_for_group`.  Will amend.

addons/udes_stock/models/stock_location.py
PeterC10: I wish that were true, I've added some very curious names in the code!

PeterC10: I think... `record_is_child_of`... that way it is clear from the name that we are checking it is a child of a particular location. :)

addons/udes_stock_refactoring/models/stock_picking.py
```diff
@@ -43,6 +41,43 @@ def _remove_misleading_values(self, **kwargs):
...
+
+        By setting the package levels as needed we avoid extra package levels potentially
+        being created later on via `_check_entire_pack`. This then also avoids the potential for
+        erroneous moves being created when `_generate_moves` is called on a newly generated
+        package level.
+        """
+        PackageLevel = self.env["stock.package_level"]
+
+        move_lines_with_package_levels = move_lines.filtered("package_level_id")
+        package_levels = move_lines_with_package_levels.mapped("package_level_id")
```
ws-mps, line 63: I don't think we need `mapped` for relational fields anymore (same for line 66).

PeterC10: Nope, old habits... fixed.

addons/udes_stock_refactoring/models/stock_picking.py
PeterC10: Fixed.

addons/udes_stock/models/stock_location.py
PeterC10: Went for `record_is_child_of_self` and added unit tests, please review.

armandcela (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -871,8 +871,12 @@ def validate_picking(self, create_backorder=False, force_validate=False):
             # Check if a backorder is required and can be created:
             # a) Multi users is enabled
             # b) There are incomplete move lines and create backorder is allowed
+            # c) There are incomplete moves (ie partially available) and create backorder is allowed
             multi_users_enabled = self.picking_type_id.u_multi_users_enabled
-            if self.move_line_ids.get_lines_incomplete():
+            if (
+                self.move_line_ids.get_lines_incomplete()
+                or self.move_lines.get_moves_with_quantity_todo()
```
LoKKeR96, line 878: Wouldn't we only need get_moves_with_quantity_todo()  ?

PeterC10: Fair point, if we have a move line in a picking and not the move then something elsewhere has gone very wrong and we're all going to be very sad. Have made the change.

LoKKeR96 (approved):
As you mentioned in one of the commits, long term we might need to review the tests but so far so good!

PR #722, opened 2022-09-13
Story #2657 Mobile UI: Allow swaps for tracked products at Pick

jrbirch99:
- Backend changes to allow for tracked products to be swapped in/out
- Tests for the Changes

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -5,7 +5,7 @@ import logging
 from odoo import models, fields, api, _
 from odoo.tools.float_utils import float_compare, float_round
 from odoo.osv import expression
-from odoo.exceptions import ValidationError
+from odoo.exceptions import ValidationError, UserError
 
 _logger = logging.getLogger(__name__)
 
@@ -254,8 +254,8 @@ class StockMoveLine(models.Model):
         return StockMoveLine.search(domain, order=order)
 
     def _find_move_lines(self, uom_qty, product, package=None, lot_name=None, location=None):
-        """ Find a subset of move lines from self matching the subset of the key:
-            product, package, lot and location.
+        """Find a subset of move lines from self matching the subset of the key:
+        product, package, lot and location.
         """
         funcs = [lambda ml: ml.product_id == product]
         if package:
@@ -277,13 +277,13 @@ class StockMoveLine(models.Model):
     def prepare(
         self, product_ids=None, package=None, location=None, result_package=None, location_dest=None
     ):
-        """ Prepare incomplete move lines in self into groups of mls to be updated with
-            the same values. It will split move lines if required.
-            Product_ids is a list of dictionaries with the following keys:
-            * barcode: string
-            * qty: integer
-            * lot_names: list of strings (optional)
-            Example: [{"barcode": "PROD01", "uom_qty":1}]
+        """Prepare incomplete move lines in self into groups of mls to be updated with
+        the same values. It will split move lines if required.
+        Product_ids is a list of dictionaries with the following keys:
+        * barcode: string
+        * qty: integer
+        * lot_names: list of strings (optional)
+        Example: [{"barcode": "PROD01", "uom_qty":1}]
 
         """
         Product = self.env["product.product"]
@@ -302,8 +302,8 @@ class StockMoveLine(models.Model):
                 # TODO: this is probably where swap package should go
                 if (pack_mls & mls) != pack_mls:
                     raise ValidationError(
-                        _("All package %s move lines cannot be found in picking %s") %
-                        (package.name, mls.picking_id.name)
+                        _("All package %s move lines cannot be found in picking %s")
+                        % (package.name, mls.picking_id.name)
                     )
                 res[pack_mls] = vals
             else:
@@ -320,19 +320,17 @@ class StockMoveLine(models.Model):
             is_serial = product_tracking == "serial"
             is_lot = product_tracking == "lot"
             if is_serial and len(lot_names) != quantity:
-                raise ValidationError(
-                    _("Not enough serial numbers provided.")
-                )
+                raise ValidationError(_("Not enough serial numbers provided."))
             elif is_lot and len(lot_names) != 1:
-                raise ValidationError(
-                    _("Too many lots provided.")
-                )
+                raise ValidationError(_("Too many lots provided."))
             qty_done = quantity
             if product_tracking != "none":
                 quantity_fulfilled = 0
                 if is_serial:
                     qty_done = 1
-                res, lot_names = self._swap_tracked_items(mls, quantity, product, location, is_serial, vals, res, lot_names)
+                res, lot_names = self._swap_tracked_items(
+                    mls, quantity, product, location, is_serial, vals, res, lot_names
+                )
                 for lot_name in lot_names:
                     lot_name_val = lot_name
                     # If it is incoming we don't need to match the lot
@@ -343,10 +341,7 @@ class StockMoveLine(models.Model):
                     prod_mls, new_ml = mls._find_move_lines(
                         qty_done, product, package, lot_name, location
                     )
-                    prod_dict = {
-                        "qty_done": qty_done,
-                        "lot_name": lot_name_val
-                    }
+                    prod_dict = {"qty_done": qty_done, "lot_name": lot_name_val}
                     prod_dict.update(vals)
                     res[prod_mls] = prod_dict
                     quantity_fulfilled += qty_done
@@ -356,9 +351,7 @@ class StockMoveLine(models.Model):
                     if quantity_fulfilled == quantity:
                         break
             else:
-                prod_mls, new_ml = mls._find_move_lines(
-                    qty_done, product, package, None, location
-                )
+                prod_mls, new_ml = mls._find_move_lines(qty_done, product, package, None, location)
                 prod_dict = {"qty_done": qty_done}
                 prod_dict.update(vals)
                 res[prod_mls] = prod_dict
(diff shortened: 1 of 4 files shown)


PeterC10:
Nit: Can you reword the commits so that each line doesn't exceed 72 characters? Makes the message not display so nicely on GitHub.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -332,6 +332,84 @@ def prepare(
                 quantity_fulfilled = 0
                 if is_serial:
                     qty_done = 1
+
+                # Swapping the tracked products
+                lot_names_on_picking = self.lot_id.mapped("name")
```
PeterC10, line 337: Can we refactor this out into its own method? `prepare` is already huge and is fast becoming the new `update_picking` from v11!

Maybe we can do something like:

`lot_names = self._get_lot_names`

addons/udes_stock/models/stock_move_line.py
```diff
@@ -332,6 +332,84 @@ def prepare(
...
+                # If there swapping tracked products are allowed and some lot_names have been swapped in
+                if mls.u_picking_type_id.u_allow_swapping_tracked_products and lot_names_swapped_in:
+                    # Find movelines in picking for location scanned in
+                    picking_mls, new_ml = mls._find_move_lines(quantity, product, None, None, location)
+                    picking_mls = picking_mls.filtered(lambda m: m.qty_done < m.product_uom_qty)
+                    mls_need_changing = picking_mls.filtered(lambda mls: mls.lot_id.name not in lot_names) 
+                    quantity_swapping = sum(mls_need_changing.mapped("product_uom_qty"))
+
+                    # Workout the lot_ids that were swapped in and store the ones that were swapped out
+                    lot_ids_swapped_in = self.env["stock.production.lot"].search([("name", "in", lot_names_swapped_in)])
```
PeterC10, line 348: `self.env["stock.production.lot"]` should be in a variable, i.e. `Lot` and placed at the top of the method. Same for `stock.move.line` on line 352.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -332,6 +332,84 @@ def prepare(
...
+                    # Generator expressions to easily iterate over
+                    mls_need_changing = (ml for ml in mls_need_changing)
+                    lot_ids_swapped_in = (lot for lot in lot_ids_swapped_in)
+                    
+                    if is_serial:
+                        # Want to update the movelines on each serial tracked product
+                        # until all the quantity scanned in is zero
+                        while quantity_swapping > 0:
+                            ml, lot_id = next(mls_need_changing), next(lot_ids_swapped_in)
+                            prod_dict = {
```
PeterC10, line 363: Q: Indentation looks a bit off, has the file been black formatted?

addons/udes_stock/models/stock_move_line.py
jrbirch99: I couldn't see a nice way to do this as there will be lots of arguments going into the function and more than one coming out, but I the prepare() method is too long, so I will do it

addons/udes_stock/models/stock_move_line.py
```diff
@@ -332,6 +332,84 @@ def prepare(
                 quantity_fulfilled = 0
                 if is_serial:
                     qty_done = 1
+
+                # Swapping the tracked products
+                lot_names_on_picking = self.lot_id.mapped("name")
+                lot_names_swapped_in = list(set(lot_names) - set(lot_names_on_picking))
+                # If there swapping tracked products are allowed and some lot_names have been swapped in
+                if mls.u_picking_type_id.u_allow_swapping_tracked_products and lot_names_swapped_in:
```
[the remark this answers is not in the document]
jrbirch99: This would not raise an error currently, I think it probably should. This method (as far as I can see) only gets called in the flow, so coming from the flow we could never have that situation, but if it was used elsewhere I guess we could

addons/udes_stock/models/stock_move_line.py
jrbirch99: it had not - but they all have been now

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,103 @@ def prepare(
                     mls |= new_ml
         return dict(res)
 
+    def _swap_tracked_items(
+        self, mls, quantity, product, location, is_serial, vals, res, lot_names
+    ):
+        """
+        This handles the tracked product swaps. This method will produce a dictionary of a dictionary containing
+        update information on different movelines so that the tracked products can be swapped. An adjusted list of
+        lot_names is also returned.
```
PeterAlabaster, line 369: nit: can we describe what a "tracked product swap" is in this docstring?

PeterAlabaster (approved):

addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -540,6 +540,511 @@ def test_over_mark_error(self):
...
+        product_ids = [{"barcode": "productStrawberry", "uom_qty": 3, "lot_names": ["1", "4", "5"]}]
+        location = self.test_stock_location_01
+        pallet_to_pick_onto = self.create_package(name="UDES1")
+
+        res = picking.move_line_ids.prepare(
+            product_ids=product_ids, location=location, result_package=pallet_to_pick_onto
+        )
+        for mls, mls_values in res.items():
+            mls.mark_as_done(mls_values)
+        self.assertIn(picking.move_line_ids.lot_id.mapped("name")[0], ["1", "5", "4"])
```
ws-mps, line 603: What about only one assert equatl with the lto names sorted? We want to make sure we exactly have all three.

addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -540,6 +540,511 @@ def test_over_mark_error(self):
...
+                lot_name=f"{lot_num}",
+            )
+        product_info = [{"product": self.tangerine, "uom_qty": 10}]
+        first_picking = self.create_picking(
+            self.picking_type_pick, products_info=product_info, assign=True
+        )
+        second_picking = self.create_picking(
+            self.picking_type_pick, products_info=product_info, assign=True
+        )
+
```
ws-mps, line 660: Would be good to check that lot of the first picking move line is "1" and of the second picking is "2".

addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -540,6 +540,511 @@ def test_over_mark_error(self):
...
+        )
+        self.create_quant(
+            self.tangerine.id, location_id=self.test_stock_location_01.id, qty=5, lot_name="3"
+        )
+
+        product_info = [{"product": self.tangerine, "uom_qty": 10}]
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=product_info, assign=True
+        )
+
```
ws-mps, line 697: Would be good to check that we have one line of 10 in the picking.

addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -540,6 +540,511 @@ def test_over_mark_error(self):
...
+        )
+        self.create_quant(
+            self.tangerine.id, location_id=self.test_stock_location_01.id, qty=10, lot_name="3"
+        )
+
+        product_info = [{"product": self.tangerine, "uom_qty": 10}]
+        picking = self.create_picking(
+            self.picking_type_pick, products_info=product_info, assign=True
+        )
+
```
ws-mps, line 745: Would be good to check we have 2 move lines in picking.

PeterC10 (approved):

PeterAlabaster (approved):

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,152 @@ def prepare(
...
+                )
+
+                # Workout the lot_ids that were swapped in and store the ones that were swapped out
+                lot_ids_swapped_in = StockProductionLot.search(
+                    [("name", "in", lot_names_swapped_in)]
+                )
+                lot_ids_swapped_out = picking_mls.lot_id
+
+                # If any of the lot_ids_swapped_in belong to another move_line
+                swapped_lot_on_assigned_mls = StockMoveLine.search(
```
ws-mps, line 414: I don't think we care of the order of the move lines, so probably better to order by id.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,152 @@ def prepare(
...
+                            "lot_name": lot_id.name,
+                            "lot_id": lot_id.id,
+                        }
+                        prod_dict.update(vals)
+                        res[ml] = prod_dict
+                        quantity_swapping -= qty_done
+                else:
+                    qty_done = quantity_swapping
+                    # Update the first lot with all scanned in lot information
+                    ml, lot_id = next(mls_need_changing), next(lot_ids_swapped_in)
```
ws-mps, line 460: Can we use `move_lines_for_qty()` to find the move lines we need for that lot and qty?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,152 @@ def prepare(
...
+                            prod_dict = {"product_uom_qty": abs(quantity), "qty_done": 0}
+                            res[ml] = prod_dict
+                
+                # Remove swapped in lot from old move line and replace it with swapped out lot
+                # Can zip as lot_ids_swapped_out will never have fewer item then swapped_lot_on_assigned_mls
+                for old_lot, mls_to_change in zip(lot_ids_swapped_out, swapped_lot_on_assigned_mls):
+                    # If we partially pick a lot that is reserved on another picking then we need to update the 
+                    # mls on that picking to have a new reserved quantity (old quantity - quantity swapped out) and create another moveline with
+                    # for the lot that was swapped out.
+                    if not is_serial and mls_to_change.product_uom_qty > amount_to_be_swapped:
```
ws-mps, line 489: Can `mls_to_change` contain more than one move line?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,152 @@ def prepare(
...
+                    res[ml] = prod_dict
+                    quantity_swapping -= reserved_quantity
+                    # If the lot quantity covers multiple movelines then delete these movelines
+                    # Unless it has partially picked some moveline then update the reserved qty on that moveline
+                    while quantity_swapping > 0:
+                        ml = next(mls_need_changing)
+                        reserved_quantity = ml.product_uom_qty
+                        quantity_swapping -= reserved_quantity
+                        if quantity_swapping >= 0:
+                            ml.unlink()
```
ws-mps, line 478: I think we should be able to cope without unlinking by just preserving original `product_uom_qty` and updating the `lot_id` of the move lines.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -366,10 +360,152 @@ def prepare(
...
+                # Can zip as lot_ids_swapped_out will never have fewer item then swapped_lot_on_assigned_mls
+                for old_lot, mls_to_change in zip(lot_ids_swapped_out, swapped_lot_on_assigned_mls):
+                    # If we partially pick a lot that is reserved on another picking then we need to update the 
+                    # mls on that picking to have a new reserved quantity (old quantity - quantity swapped out) and create another moveline with
+                    # for the lot that was swapped out.
+                    if not is_serial and mls_to_change.product_uom_qty > amount_to_be_swapped:
+                        prod_dict = {"lot_name": mls_to_change.lot_id.name, "lot_id": mls_to_change.lot_id.id, "product_uom_qty": mls_to_change.product_uom_qty - amount_to_be_swapped, "qty_done": 0.0}
+                        mls_to_change.write({"product_uom_qty": mls_to_change.product_uom_qty - amount_to_be_swapped})
+                        res[mls_to_change] = prod_dict
+                        new_mls = mls_to_change.copy()
```
ws-mps, line 493: Are we not using `_split()` because we don't want to `bypass_reservation_update` ?

addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -540,6 +540,795 @@ def test_over_mark_error(self):
...
+        res = first_picking.move_line_ids.prepare(
+            product_ids=product_ids, location=location, result_package=pallet_to_pick_onto
+        )
+        for mls, mls_values in res.items():
+            mls.mark_as_done(mls_values)
+        self.assertEqual(second_picking.move_line_ids.mapped("lot_id.name"), ["2","1"])
+        self.assertEqual(second_picking.move_line_ids.mapped("product_uom_qty"), [5.0, 5.0])
+        self.assertEqual(second_picking.move_line_ids.mapped("qty_done"), [0.0,0.0])
+
+        self.assertEqual(Quant.search([("lot_id.name", "=", "1")]).reserved_quantity, 10.0)
```
ws-mps, line 719: We could use a helper function that checks for a given lot name what the reserved quantity should be. Then use it in all the tests, since the previous ones we are only checking if the move line info has been swapped. Also consider checking the reserved quantity of the original lot before the swap.

addons/udes_stock/models/stock_move_line.py
jrbirch99: The lot information we have here does not necessarily have a moveline so we can't use `move_lines_for_qty()`

addons/udes_stock/models/stock_move_line.py
jrbirch99: no :smile:

PR #724, opened 2022-09-21
[IMP] udes_stock: Track changes to location.

kdwyer:
story/2432

Track changes to a location's usage type and parent location id to make analysing support issues easier.

### DIFF
--- a/addons/udes_stock/views/stock_location_views.xml
+++ b/addons/udes_stock/views/stock_location_views.xml
@@ -51,6 +51,14 @@
                 <attribute name="domain">[('usage', '!=', 'internal')]</attribute>
             </xpath>
 
+            <!-- Add message log to display attribute tracking messages. -->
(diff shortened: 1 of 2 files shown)


addons/udes_stock/views/stock_location_views.xml
```diff
@@ -51,6 +51,14 @@
                 <attribute name="domain">[('usage', '!=', 'internal')]</attribute>
             </xpath>
 
+            <!-- Add message log to show archiving events. -->
```
PeterC10, line 54: Super super nit: Could we make this comment more generic? As it isn't just archiving it will show.

PeterC10 (approved):

addons/udes_stock/views/stock_location_views.xml
kdwyer: Oops, copypasta :blush:. Comment updated.

jrbirch99 (approved):

PR #729, opened 2022-09-29
[IMP] udes_stock, udes_stock_picking_batch: Add logging for monitor

LoKKeR96:
Story: x1666

Signed-off-by: Lorenzo Cucurachi <lorenzo.cucurachi@unipart.com>

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -8,6 +8,8 @@ from odoo.addons.udes_common.models.fields import PreciseDatetime
 from odoo.exceptions import ValidationError
 from lxml import etree
 
+from ..utils import UDES_STATISTICS_LOG_FORMAT
+
 _logger = logging.getLogger(__name__)
 
 
@@ -850,7 +852,27 @@ class StockPicking(models.Model):
         ):
             pick._create_own_procurement_group()
         res = super().action_confirm()
+        return res
 
+    def _action_done(self):
+        """Extend to collect stats for the UDES monitor"""
+        # Collect action_done stats for monitor
+        with self.statistics() as stats:
+            res = super()._action_done()
+
+        # Ensure when multiple picking types are involved that the names are combined
+        picking_type_names = ",".join(self.mapped("picking_type_id.name"))
+        # Write the log
+        _logger.info(
+            UDES_STATISTICS_LOG_FORMAT,
+            self._name,
+            self.ids,
+            picking_type_names,
+            "action_done",
+            stats.elapsed,
+            stats.count,
+            stats.count / stats.elapsed
+        )
         return res
 
     def validate_picking(self, create_backorder=False, force_validate=False):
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -850,7 +850,28 @@ def action_confirm(self):
...
+        return res
+
+    def _action_done(self):
+        """Extend to collect stats for the UDES monitor"""
+        # Collect action_done stats for monitor
+        with self.statistics() as stats:
+            res = super()._action_done()
 
+        # Ensure when multiple picking types are involved that the names are combined
+        picking_type_names = ""
```
armandcela, line 862: Do picking_type_names need a comma to separate them in case there are more than one.
```suggestion
        picking_type_names = ",".join(self.mapped("picking_type_id.name")
```
Remove the for loop in that case

addons/udes_stock/models/stock_picking.py
```diff
@@ -899,11 +920,20 @@ def validate_picking(self, create_backorder=False, force_validate=False):
             # but still needs more access rights for the flow
             # We don't want odoo stock module to create backorders. That is handed in UDES layer.
             self.sudo().with_context({"cancel_backorder": True})._action_done()
+        # Write the log
+        # It should never happen but ensure when multiple picking types
+        # are involved that the names are combined
+        picking_type_names = ""
```
armandcela, line 926: Same here

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -730,67 +730,81 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+                    # but still needs more access rights for the flow
+                    to_add.sudo().write({"batch_id": self.id})
 
-            with self.statistics() as stats:
                 picks_todo.sudo().with_context(tracking_disable=True)._action_done()
 
+            # Write the log
+            # It should never happen but ensure when multiple picking types
+            # are involved that the names are combined
+            picking_type_names = ""
```
armandcela, line 797: Also here

addons/udes_stock/models/stock_picking.py
LoKKeR96: Comma works fine, and yes perfect, thank you, I was missing how to do this. Will change soon.

kdwyer (approved):

PeterAlabaster (approved):

jrbirch99 (approved):

addons/udes_stock/models/stock_picking.py
armandcela: Yep happy with changes, just not needed the if because join doesnt add comma when length is 1.
Happy to resolve all threads after that

armandcela (approved):

addons/udes_stock/models/stock_picking.py
LoKKeR96: Addressed

PR #730, opened 2022-09-29
[IMP] udes_stock: Add logging for monitoring

LoKKeR96:
Add logging for action_done, mark_as_done and drop_off_picked.
These will be used by the UDES monitor to generate stats and present
these to our developers.

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -1,10 +1,15 @@
 # -*- coding: utf-8 -*-
 from itertools import groupby
+
 from odoo import api, fields, models, tools, _
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools.float_utils import float_compare, float_round
 from copy import deepcopy
+import logging
 from collections import Counter, defaultdict
+from ..utils import UDES_STATISTICS_LOG_FORMAT
+
+_logger = logging.getLogger(__name__)
 
 
 class StockMoveLine(models.Model):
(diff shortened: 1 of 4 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -465,37 +465,57 @@ def action_done(self):
...
+            self.env.ref("udes_stock.picking_done").with_context(
+                active_model=picks._name,
+                active_ids=picks.ids,
+                action_filter="picking.action_done",
+                **extra_context
+            ).run()
+
+        # Write the log
+        # Ensure when multiple picking types are involved that the names are combined
+        picking_type_names = ""
```
armandcela, line 503:
```suggestion
       picking_type_names = ",".join(self.mapped("picking_type_id.name")
```

addons/udes_stock/models/stock_picking_batch.py
```diff
@@ -793,60 +793,73 @@ def drop_off_picked(self, continue_batch, move_line_ids, location_barcode, resul
...
+                # Add backorders to the batch
+                to_add.write({"batch_id": self.id})
 
-            with self.statistics() as stats:
                 picks_todo.sudo().with_context(tracking_disable=True).action_done()
 
+            # Write the log
+            # It should never happen but ensure when multiple picking types
+            # are involved that the names are combined
+            picking_type_names = ""
```
armandcela, line 852: Same here

addons/udes_stock/models/stock_move_line.py
```diff
@@ -164,135 +167,145 @@ def mark_as_done(
...
+                )
+                self._trigger_print_for_move_line_mark_as_done(
+                    active_model=picking._name,
+                    active_ids=picking.ids,
+                    print_records=print_mls,
+                    action_filter="move_lines.mark_as_done",
+                )
+        # Write the log
+        _logger.info(
+            "(%s) %s (%s) mark_as_done in %.2fs, %d queries, %.2f q/s",
```
kdwyer, line 301: It might be worth abstracting this message format, as it's common? ('mark_as_done' could become a `%s`)

kdwyer (approved):

PeterAlabaster (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -465,37 +465,59 @@ def action_done(self):
...
+                active_ids=picks.ids,
+                action_filter="picking.action_done",
+                **extra_context
+            ).run()
+
+        # Write the log
+        # Ensure when multiple picking types are involved that the names are combined
+        picking_type_names = (
+            self.picking_type_id.name
+            if len(self) == 1
```
ws-mps, line 505: Why do we need this if? Join shouldn't add "," if there is only one element in the list.

jrbirch99 (approved):

addons/udes_stock/models/stock_move_line.py
LoKKeR96: We could make a module and abstract these calls or do other clever things but, I didn't go further than what I've done right now to try and limit the work otherwise there's always something to improve. Happy to do more work as long as TL agrees.

armandcela (approved):

addons/udes_stock/models/stock_move_line.py
kdwyer: I think it would be helpful to prevent the formats from accidentally diverging, but happy for someone else to make the final call.

LoKKeR96: @ws-mps any thoughts?

ws-mps: It is a good idea to go to that direction, but it is out of scope of this story to create a new module etc. If we want to predefine somewhere the format (constant?) where we 'mark_as_done' (and others) become a %s, I'm fine with that.

addons/udes_stock/models/stock_picking.py
LoKKeR96: Addressed

addons/udes_stock/models/stock_move_line.py
LoKKeR96: I will abstract and store the message format somewhere in udes_stock so that it is easier to change and track and then import it where it's used. How does it sound? @kdwyer

PR #733, opened 2022-10-12
[FIX] udes_security: Handle javascript `do_action("extid")` properly

PeterAlabaster:
Signed-off-by: Peter Alabaster <peter.alabaster@unipart.com>

User-story: 3029

### DIFF
--- a/addons/udes_security/controllers/main.py
+++ b/addons/udes_security/controllers/main.py
@@ -77,7 +77,10 @@ class SecureAction(Action):
             try:  # If you middle click a menu, action_id is a string..
                 action_id = int(action_id)
             except ValueError:
-                raise ValidationError("Invalid action ID specified.")
+                try:  # JS `do_action("action_extid")` results in the parameter coming in as a string here
+                    action_id = request.env.ref(action_id).id
+                except ValueError:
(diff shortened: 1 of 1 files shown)


addons/udes_security/controllers/main.py
```diff
@@ -77,7 +77,10 @@ def _get_menu_info_from_action_id(self, action_id):
             try:  # If you middle click a menu, action_id is a string..
                 action_id = int(action_id)
             except ValueError:
-                raise ValidationError("Invalid action ID specified.")
+                try:  # JS `do_action("action_extid")` results in the parameter coming in as a string here
```
LoKKeR96, line 80: It's probably not an issue but wouldn't it be better to have a separate if statement and try and use the string and except separately?

PeterAlabaster: we only want to run the second check if it fails the int pass, because it's more performant. The int could be wrapped in a string like "11" so can't do `type()` or `isinstance()` in an if check, instead we just have to let it fail and only if it fails conversion run this second pass

LoKKeR96 (approved):

kdwyer (approved):

PR #734, opened 2022-10-19
Prevent non-done backorder pickings from having date done set

armandcela:


addons/udes_stock/models/stock_picking.py
```diff
@@ -214,6 +215,8 @@ def create(self, vals):
         picking = super().create(vals)
         if not picking.backorder_id:
             picking.write({"u_original_picking_id": picking.id})
+        # Just in case date_done field is set to datetime.now but picking is not done
```
PeterAlabaster, line 218: nit: move comment to docstring on function

addons/udes_stock/models/stock_picking.py
```diff
@@ -214,6 +215,8 @@ def create(self, vals):
...
+        picking.safe_prevent_date_done_on_not_done_pickings()
```
PeterAlabaster, line 219: nit: maybe should be called something slightly different, like
```python
picking.clear_date_done_if_not_done()
```

addons/udes_stock/models/stock_picking.py
```diff
@@ -3093,3 +3098,43 @@ def open_back_orders(self):
...
         )
+
+    @api.multi
+    def safe_prevent_date_done_on_not_done_pickings(self):
+        """
+        Set date_done field to empty in case that not all moves are in state done
+        """
+        for picking, moves in self.mapped("move_lines").groupby("picking_id"):
+            if picking.date_done and moves and not all(
+                    move.state in ("done", "cancel") for move in moves
```
PeterAlabaster, line 3109: does this mean, that if all move lines are cancelled - the picking won't have its date done cleared?

we might need a secondary check - `if all(move.state == "cancel" for move in moves)` then date_done should be cleared

addons/udes_stock/tests/test_picking_prevent_date_done.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-
```
PeterAlabaster, line 1: nit: can be removed

addons/udes_stock/models/stock_picking.py
```diff
@@ -3093,3 +3098,43 @@ def open_back_orders(self):
         return self.open_related_pickings(
             self.u_created_back_orders.ids, "Created Back Orders"
         )
+
+    @api.multi
```
PeterC10, line 3102: Nit: Not needed.

addons/udes_stock/models/stock_picking.py
```diff
@@ -3093,3 +3098,43 @@ def open_back_orders(self):
...
+    def safe_prevent_date_done_on_not_done_pickings(self):
+        """
+        Set date_done field to empty in case that not all moves are in state done
+        """
+        for picking, moves in self.mapped("move_lines").groupby("picking_id"):
```
PeterC10, line 3107: Q: We already have the pickings in self, why do we need to group by the move lines then by picking again?

We could instead loop through the pickings filtering those with date_done set, then do a `search_count` for moves of that picking ID that are not done, and if there are any, clear the date.

addons/udes_stock/models/stock_picking.py
```diff
@@ -3093,3 +3098,43 @@ def open_back_orders(self):
...
+        In case that exists create an issue with the traceback on it, otherwise first create
+        the project, after create issue for the created project.
+        """
+        # Give sudo access rights not to have access rights issue
+        self = self.sudo()
+        Project = self.env["project.project"]
+        Issue = self.env["project.task"]
+        self.ensure_one()
+        traceback_message = traceback.extract_stack()
+        prevent_name = "Prevent non-done pickings having date done set"
```
PeterC10, line 3129: This is a bit long for the project name, which we usually overview with kanban cards. But tricky to think of a title for this... maybe `Open Pickings with Date Done`?

addons/udes_stock/models/stock_picking.py
```diff
@@ -3093,3 +3098,43 @@ def open_back_orders(self):
...
+        prevent_project = Project.search([
+            ("name", "=", prevent_name)
+        ], limit=1)
+        if not prevent_project:
+            prevent_project = Project.create({"name": prevent_name})
+        vals = {
+            "project_id": prevent_project.id,
+            "name": ("[%s] date_done set on a not completed picking" % self.name),
```
PeterC10, line 3137: Suggestion: `... date_done set on open picking`

addons/udes_stock/tests/test_picking_prevent_date_done.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestPickingPreventDateDone, cls).setUpClass()
+        cls.prevent_name = "Prevent non-done pickings having date done set"
+
+    @classmethod
+    def prepare_move_vals(cls, product, qty, picking_type, state="draft", **kwargs):
+        vals = {
+            'product_id': product.id,
```
PeterC10, line 19: Can we black format this file?

addons/udes_stock/models/stock_picking.py
PeterC10: We could simplify it further, e.g. `check_date_done`.

addons/udes_stock/models/stock_picking.py
armandcela: self.safe_prevent_date_done_on_not_done_pickings(), self can be multiple pickings, as i am understanding from  action_done method, needed to handle that scenario i think

PeterC10: Yes, so `self` with have multiple pickings. So we can do `for picking in self` - then check the moves from there right?

PeterC10: Or probably better to do `for picking in self.filtered("date_done")` for example

PeterC10:
```
for picking in self.filtered("date_done"):
    open_move_count = Move.search_count([("picking_id", "=", picking.id), ("state", "!=", "done")])
    if open_move_count:
        # clear picking done date
```

Something like this? It would avoid needing to group or reference all moves, which on big systems can have an impact on performance.

armandcela: Not agree with this as maybe that some moves are on state cancel, and we dont want to remove date_done on those, as @PeterAlabaster suggested which i agree should check that at least one should be done.
Pushed another change for that if happy.
` open_move_count = Move.search_count([("picking_id", "=", picking.id), ("state", "!=", "done")])`

Agree about mapping and groupping

armandcela: The reason why i did that is because i wanted moves to and instead assigning them again like moves = picking.move_lines, thought better to group all moves by picking id

PeterAlabaster: agreed, but I think we could at least remove the groupby, just do `for picking in self: moves = picking.mapped("move_line_ids")`

addons/udes_stock/models/stock_picking.py
```diff
@@ -3099,15 +3100,14 @@ def open_back_orders(self):
...
-    def safe_prevent_date_done_on_not_done_pickings(self):
+    def check_date_done(self):
         """
         Set date_done field to empty in case that not all moves are in state done
         """
         for picking, moves in self.mapped("move_lines").groupby("picking_id"):
-            if picking.date_done and moves and not all(
-                    move.state in ("done", "cancel") for move in moves
-            ):
+            picking_states = moves.mapped("state")
```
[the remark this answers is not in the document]
armandcela: move_states better :P

addons/udes_stock/models/stock_picking.py
PeterC10: We should avoid `mapped` here, on big pickings it will be ineffective.

addons/udes_stock/models/stock_picking.py
```diff
@@ -481,7 +482,7 @@ def action_done(self):
 
             res = super(StockPicking, self.with_context(lock_batch_state=True)).action_done()
             # Just in case date_done field is set to datetime.now but picking is not done
```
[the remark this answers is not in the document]
armandcela: The previous suggestion was on def create, this is in action_done. Don't think can transfer on docstring, and i think is better to have a comment here. Maybe the comment is not very well structured

addons/udes_stock/models/stock_picking.py
PeterC10: Didn't consider cancelled moves.

Maybe:

```
Move = self.env["stock.move"].with_context(prefetch_fields=False)
...
for picking in self.filtered("date_done"):
    clear_date_done = False
    none_done_moves = Move.search([("picking_id", "=", picking.id), ("state", "!=", "done")], order="id")
    if none_done_moves:
        none_done_moves.read(["state"])
        for move in none_done_moves:
            if move.state != "cancel":
                clear_date_done = True
                break
    if clear_date_done:
        picking.write({"date_done": False})
```

PeterC10: Or we could search for all moves of any picking in self first, then group by those by picking ID and then we still avoid mapped. That would avoid calling `search` multiple times if `self` has more than 1 picking.

PeterC10: So we could do:

```
none_done_moves = Move.search([("picking_id", "in", self.filtered("date_done").ids), ("state", "!=", "done")], order="id")

for picking, move in none_done_moves.groupby("picking_id"):
    ...
```

I don't really mind too much with this, just the `mapped` calls we should avoid if possible, as we've seen many times where this has been a bit of a bottleneck.

addons/udes_stock/models/stock_picking.py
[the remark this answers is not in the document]
PeterC10: Will never forgive Odoo for their naming of the `Move` and `Move Line` fields on pickings... `move_lines` vs `move_line_ids`...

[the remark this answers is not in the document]
armandcela: move_lines -- stock.move but is not as they define one2many fields with _ids at the end.
Yeah confusing the name of models they have picked up.

addons/udes_stock/models/stock_picking.py
armandcela: If we filter out the moves of a pick that are done, and if there are cancel moves in the same picking we maybe empty the field date_done where we shouldnt. I think we shouldnt filter out the moves that are done

PeterC10 (approved):

PeterAlabaster (approved):

PR #736, opened 2022-10-26
[REM] udes_suggest_location: Removing suggested locations on prepare …

armandcela:
…move lines

Removing the set of destination location on prepare move lines 
Removing unit tests related to that change

Story: 2923


PeterC10:
Nit: Can we use `REM` for the commit tag? As the commit is only removing, not adding anything.

PeterC10 (approved):

adamrpatrick (approved):

PR #739, opened 2022-11-11
[ADD] udes_stock: Allow source location to be changed on open pickings

armandcela:
Story: 3204


addons/udes_stock/models/stock_picking.py
```diff
@@ -456,6 +456,16 @@ def assert_not_lot_restricted(self):
                             )
                         )
 
+    def button_draft(self):
+        """
+        Button that passes the picking to draft state. Adding extra check on backend that should be
```
kdwyer, line 461: This wording might more idiomatic:

A button that moves the picking to the draft state. Adds an extra check in the backend that should be
triggered for only one transfer at time and the state is in waiting or confirmed
when the button is clicked.

kdwyer (approved):

adamrpatrick (approved):

addons/udes_stock/views/stock_picking.xml
```diff
@@ -127,6 +127,11 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
 
+            <!--Add Set To Draft button visible when picking has been marked as to do-->
+            <xpath expr="//button[@name='action_confirm']" position="before">
+                <button name="button_draft" attrs="{'invisible': [('state', 'not in', ['waiting', 'confirmed'])]}" string="Set To Draft" type="object" class="oe_highlight" groups="base.group_user"/>
```
PeterC10, line 132: Nit: `groups` should be `stock.group_stock_user`

PeterC10: Nevermind, didn't realise other buttons were also using `base.group_user` (including unreserve), only a few use the stock group... I guess the logic is that if you have access to the pickings then you can press the buttons. :)

addons/udes_stock/models/stock_picking.py
```diff
@@ -456,6 +456,16 @@ def assert_not_lot_restricted(self):
...
+        triggered only for one transfer at time and state is on waiting or confirmed
+        when button is clicked.
+        """
+        self.ensure_one()
+        if self.state in ("waiting", "confirmed"):
+            self.write({"state": "draft"})
```
PeterC10, line 467: Probably a silly question, but does this also unreserve the stock?

addons/udes_stock/views/stock_picking.xml
PeterC10: Yep, I only checked after adding the comment. :sweat_smile: 

Although with the example you gave there is another super super nit... `Set to Draft` to match with `Mark as Todo`? (to isn't capitalised in a title) - happy for you to leave that though :smile:

addons/udes_stock/models/stock_picking.py
PeterC10: Ah I see, so if stock has been reserved then the picking won't get put back to draft. The user only sees the button once they've unassigned stock.

armandcela: Very good question i think, i am assuming that shouldnt be any stock reserved when the button is clicked. Otherwise we shouldnt click the button set to draft.
Or maybe just in case to add an extra check if move lines exist to pass

addons/udes_stock/views/stock_picking.xml
PeterC10, line 132: Sorry last one!

Just a suggestion:

1. Remove `class="oe_highlight"` - as this isn't a happy path or natural progression for the picking (same like how unreserve button isn't highlighted)
2. Put the Set to Draft button at the end, like the unreserve button is, again because it isn't a natural progression for the picking

PeterC10 (approved):

PR #740, opened 2022-11-15
story #3033 Picking - Improve error messaging/logging when picking can't be validated

jrbirch99:
- A more detailed error message with information on the move lines in the "done" state are given
- Unit tests were added to test the error message
- Formatting

User-story: 3033

Signed-off-by: Jack Birch <jack.birch@unipart.com>

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -35,14 +35,14 @@ class StockMoveLine(models.Model):
             lines.write({"u_picking_type_id": pick_type_id})
 
     def get_lines_todo(self):
-        """ Return the move lines in self that are not completed,
-            i.e., quantity done < quantity todo
+        """Return the move lines in self that are not completed,
+        i.e., quantity done < quantity todo
         """
         return self.filtered(lambda ml: ml.qty_done < ml.product_uom_qty)
 
     def get_lines_done(self):
-        """ Return the move lines in self that are completed,
-            i.e., quantity done == quantity todo
+        """Return the move lines in self that are completed,
+        i.e., quantity done == quantity todo
         """
         return self.filtered(lambda ml: ml.qty_done == ml.product_uom_qty)
 
@@ -143,9 +143,7 @@ class StockMoveLine(models.Model):
             # Error when trying to mark_as_done a full package or setting result package
             # when result storage format is products
             if result_package:
-                raise ValidationError(
-                    _("Invalid parameters for products target storage format.")
-                )
+                raise ValidationError(_("Invalid parameters for products target storage format."))
 
         return (result_package, parent_package)
 
@@ -157,17 +155,17 @@ class StockMoveLine(models.Model):
         package=None,
         product_ids=None,
     ):
-        """ Marks as done the move lines in self and updates location_dest_id
-            and result_package_id if they are set.
+        """Marks as done the move lines in self and updates location_dest_id
+        and result_package_id if they are set.
 
-            When product_ids is set, only matching move lines from self will
-            be marked as done for a specific quantity.
+        When product_ids is set, only matching move lines from self will
+        be marked as done for a specific quantity.
 
-            - location_dest = string or id
-            - result_package = string or id
-            - package = string or id
-            - product_ids = list of dictionaries, whose keys will be
-                              barcode, qty, lot_names
+        - location_dest = string or id
+        - result_package = string or id
+        - package = string or id
+        - product_ids = list of dictionaries, whose keys will be
+                          barcode, qty, lot_names
         """
         # Collect mark_as_done stats for monitor
         with self.statistics() as stats:
@@ -193,7 +191,8 @@ class StockMoveLine(models.Model):
                     raise ValidationError(
                         _(
                             "The location '%s' is not a child of the picking destination "
-                            "location '%s'" % (loc_dest_instance.name, picking.location_dest_id.name)
+                            "location '%s'"
+                            % (loc_dest_instance.name, picking.location_dest_id.name)
                         )
                     )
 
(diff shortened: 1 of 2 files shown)


jrbirch99:
I did formatting and changes in one commit... so the PR is a bit ugly. The only non-formatting change in stock_move_line.py are in `_prevent_write_after_done`.

PeterAlabaster:
> I did formatting and changes in one commit... so the PR is a bit ugly. The only non-formatting change in stock_move_line.py are in `_prevent_write_after_done`.

nit: can we separate the black format into its own commit

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1110,14 +1115,35 @@ def _drop_off_criterion_summary(self):
         "qty_done",
     )
     def _prevent_write_after_done(self):
-        done_lines = self.filtered(lambda x: x.state == "done")
+        done_lines = self.filtered(lambda ml: ml.state == "done")
+
+        # When a moveline is updated through desktop only one "done" move line is passed into self even when there are multiple
+        # Get all the done move lines on the picking if there is only one "done" move line
+        if len(done_lines) == 1 and self.mapped("picking_id"):
+            done_lines = self.mapped("picking_id.move_line_ids").filtered(
```
PeterC10, line 1123: Suggestion: This mapped could be pretty expensive on big databases, and this constraint will be checked fairly often.

Can we try using `search` or `search_read` here?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1110,14 +1115,35 @@ def _drop_off_criterion_summary(self):
...
+                _logger.error(
+                    f"Move Line id: {line.id} in state 'done' cannot be written to. Product id: {line.product_id.id}, Quantity: {line.qty_done}, Destination Location id: {line.location_dest_id.id}, Lot/Serial id: {line.lot_id if line.lot_id else False}, Package id: {line.result_package_id.id if line.result_package_id else False}"
+                )
+                move_line_strings.append(
+                    f"Product: {line.product_id.name}, Quantity: {line.qty_done}, Location: {line.location_dest_id.name}, Lot/Serial Number: {line.lot_name if line.lot_name else False}, Package Name: {line.result_package_id.name if line.result_package_id else False}"
+                )
+
+            raise ValidationError(
+                _("Cannot update move lines that are already 'done': \n%s")
+                % ("\n".join(move_line_strings))
```
PeterC10, line 1141: Can we include the Picking name?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1110,14 +1115,35 @@ def _drop_off_criterion_summary(self):
...
+                lambda ml: ml.state == "done"
+            )
+
         if done_lines:
-            raise ValidationError(_("Cannot update move lines that are already 'done'."))
+            # For each done line
+            # Create error message
+            move_line_strings = list()
+            for line in done_lines:
+                _logger.error(
```
PeterC10, line 1132: Picking name and/or ID would be useful here too.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1110,14 +1115,35 @@ def _drop_off_criterion_summary(self):
...
+                    f"Move Line id: {line.id} in state 'done' cannot be written to. Product id: {line.product_id.id}, Quantity: {line.qty_done}, Destination Location id: {line.location_dest_id.id}, Lot/Serial id: {line.lot_id if line.lot_id else False}, Package id: {line.result_package_id.id if line.result_package_id else False}"
```
PeterC10, line 1133: Nit: For logged statements we should use `%` for strings.

addons/udes_stock/models/stock_move_line.py
jrbirch99: I had thought about this, but I didn't think there would be much of a performance dip as the mapped will only ever be called if self contains `done_lines` which might be quite rare.   

Happy to change it though

addons/udes_stock/models/stock_move_line.py
jrbirch99: changed

addons/udes_stock/models/stock_move_line.py
jrbirch99: changed

addons/udes_stock/models/stock_move_line.py
jrbirch99: changed

PeterAlabaster (approved):

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1120,20 +1120,26 @@ def _prevent_write_after_done(self):
         # When a moveline is updated through desktop only one "done" move line is passed into self even when there are multiple
         # Get all the done move lines on the picking if there is only one "done" move line
         if len(done_lines) == 1 and self.mapped("picking_id"):
-            done_lines = self.mapped("picking_id.move_line_ids").filtered(
-                lambda ml: ml.state == "done"
+            done_lines = self.search(
+                [("picking_id", "in", self.mapped("picking_id.id")), ("state", "=", "done")]
```
PeterC10, line 1124: Couple of minor suggestions:

```
if len(done_lines) == 1:
    picking_ids = self.mapped("picking_id.id")
    if picking_ids:
        done_lines = self.search([("picking_id", "in", self.mapped("picking_id.id")), ("state", "=", "done")], order="id")
```

This avoids us doing `mapped` twice. And `order="id"` means we avoid using the default `_order` on the move line model, as the order of records doesn't matter here.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -1119,10 +1119,12 @@ def _prevent_write_after_done(self):
...
         # Get all the done move lines on the picking if there is only one "done" move line
-        if len(done_lines) == 1 and self.mapped("picking_id"):
-            done_lines = self.search(
-                [("picking_id", "in", self.mapped("picking_id.id")), ("state", "=", "done")]
-            )
+        if len(done_lines) == 1:
+            picking_ids = self.mapped("picking_id")
+            if picking_ids:
+                done_lines = self.search(
+                    [("picking_id", "in", self.mapped("picking_id.id")), ("state", "=", "done")], order="id"
```
PeterC10, line 1126: Mistake on my comment sorry, should be `("picking_id", "in", picking_ids)` - as we already have the picking_ids from line 1123.

PeterC10 (approved):

PR #743, opened 2022-12-01
[IMP] udes_stock: Make source document field editable outside 'draft'.

jrbirch99:
- Added a unit test to check updates to the source document are added to the message thread

User-story: 3206

Signed-off-by: Jack Birch <jack.birch@unipart.com>

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -40,6 +40,7 @@ class StockPicking(models.Model):
 
     _order = "priority desc, scheduled_date asc, sequence asc, id asc"
 
+    origin = fields.Char(track_visibility = "onchange")
     # Adding fields to be hidden by default from form view.
     DetailedFormViewFields = [
         "u_first_picking_ids",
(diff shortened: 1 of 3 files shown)


adamrpatrick (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -214,11 +214,14 @@ def create(self, vals):
...
         """
-        picking = super().create(vals)
-        if not picking.backorder_id:
-            picking.write({"u_original_picking_id": picking.id})
-        picking.check_date_done()
-        return picking
+        pickings = super().create(vals)
+        for picking in pickings:
+            if vals.get("origin"):
+                picking.message_post(body=f"<ul><li>{picking._fields['origin'].string}: {vals.get('origin')}</li></ul>")
```
PeterAlabaster, line 220: q: can't we just add `track_visibility="onchange"` to the `origin` field? Or does that not go into enough detail?

addons/udes_stock/views/stock_picking.xml
```diff
@@ -98,7 +98,7 @@
                 Origin also set to readonly unless the picking is in draft state
             -->
             <xpath expr="//field[@name='origin']" position="attributes">
-                <attribute name="attrs">{'readonly': [('state', '!=', 'draft')],  'required': [('picking_type_code', '=', 'incoming')]}</attribute>
+                <attribute name="attrs">{'required': [('picking_type_code', '=', 'incoming')]}</attribute>
```
armandcela, line 101: Question: Does this mean that can be changed when picking is cancelled or done?
Do we want to be able to change on those states?

addons/udes_stock/models/stock_picking.py
jrbirch99: Changed

PeterAlabaster (approved):

adamrpatrick (approved):

PR #744, opened 2022-12-02
[ADD] delivery_control: Pickings with no partners able to be cancelled

armandcela:
Story: 3208

### DIFF
--- a/addons/udes_delivery_control/views/stock_picking.xml
+++ b/addons/udes_delivery_control/views/stock_picking.xml
@@ -21,7 +21,7 @@
             </xpath>
             <!-- Make partner field mandatory -->
             <xpath expr="//field[@name='partner_id']" position="attributes">
-                <attribute name="attrs">{'required': [('u_is_delivery_control', '=', True)], 'readonly': [('state', '!=', 'draft')]}</attribute>
+                <attribute name="attrs">{'required': [('state', '=', 'draft'), ('u_is_delivery_control', '=', True)], 'readonly': [('state', '!=', 'draft')]}</attribute>
             </xpath>
             <!-- Add page to notebook -->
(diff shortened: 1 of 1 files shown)


addons/udes_delivery_control/views/stock_picking.xml
```diff
@@ -21,7 +21,7 @@
             </xpath>
             <!-- Make partner field mandatory -->
             <xpath expr="//field[@name='partner_id']" position="attributes">
-                <attribute name="attrs">{'required': [('u_is_delivery_control', '=', True)], 'readonly': [('state', '!=', 'draft')]}</attribute>
+                <attribute name="attrs">{'required': [('state', '=', 'draft'), ('u_is_delivery_control', '=', True)], 'readonly': [('state', '!=', 'draft')]}</attribute>
```
PeterAlabaster, line 24: do says `Partner is not a required field.` - are we sure it should be required in draft?

armandcela: It says that we should be able to cancel pickings if they by mistake have been created without partner. I suppose from files(EDI or importing somewhere else).

In draft we can delete it etc, so in other states should be able to cancel.
I removed in the UI to be required in other states as in fact will be readonly but that will allow to cancel

armandcela: Checked even the comments that Peter C has noted and i think we just want a way to be able to cancel when is ready. But i think we need to have the functionality to be required when created a delivery control

PeterAlabaster: Ok, resolving

PeterAlabaster (approved):

adamrpatrick (approved):

PR #745, opened 2022-12-02
[FIX] udes_stock_refactoring: Refactor partial moves correctly.

kdwyer:
story/2918

### DIFF
--- a/addons/udes_stock_cron/models/stock_picking.py
+++ b/addons/udes_stock_cron/models/stock_picking.py
@@ -95,8 +95,8 @@ class StockPicking(models.Model):
         """
         # Assign at the move level disabling refactoring
         moves = self.move_lines
-        moves._action_assign()
-        return moves.picking_id
+        assigned_moves = moves._action_assign()
+        return assigned_moves.mapped("picking_id")
 
     def reserve_stock(self):
         """
--- a/addons/udes_stock_refactoring/models/stock_move.py
+++ b/addons/udes_stock_refactoring/models/stock_move.py
@@ -258,7 +258,7 @@ class StockMove(models.Model):
                 group_pickings |= move.picking_id
                 move_lines = ml_group.filtered(lambda l: l.move_id == move)
 
-                if move_lines != move.move_line_ids:
+                if move_lines != move.move_line_ids or move.state == "partially_available":
                     # The move is not entirely contained by the move lines
                     # for this grouping. Need to split the move.
                     group_moves |= move.split_out_move_lines(move_lines)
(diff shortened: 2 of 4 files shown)


addons/udes_stock_refactoring/tests/test_refactoring.py
```diff
@@ -508,6 +508,29 @@ def test_2_pallets_over_backorder_preserves_package_levels_in_picking_chain(self
             picking._action_done()
             self._assert_package_levels(goods_out_pickings)
 
+    def test_refactors_partially_unavailable_stock_into_new_move(self):
+        """The system should split moves into available and unavailable."""
+        products_info = [
+            {"product": self.apple, "uom_qty": 15},
+        ]
+        moves = self.create_move(self.picking, products_info)
+        original_move = moves.filtered(lambda m: m.product_id == self.apple)
```
armandcela, line 517: Nit: I believe dont need original_move by filtering moves with product apple.
From what i see moves = original_move so in the asserts below can use directly moves and refactored moves.

armandcela (approved):

jrbirch99 (approved):

addons/udes_stock_refactoring/tests/test_refactoring.py
kdwyer: Good point - changed, please review.

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -95,8 +95,8 @@ def _reserve_stock_assign(self):
         """
         # Assign at the move level disabling refactoring
         moves = self.move_lines
-        moves._action_assign()
-        return moves.picking_id
+        refactored_moves = moves._action_assign()
```
ws-mps, line 98: I wouldn't call this one refactored, since there could be other code in the future that splits/changes/returns different recordset of moves.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -151,13 +151,11 @@ def _action_confirm(self, *args, **kwargs):
...
         """
         Extend _action_assign to trigger refactor action.
-        n.b. _action_assign does not return anything in core Odoo, so we
-        don't return any extra moves that may have been created
-        by refactoring.
         """
         res = super(StockMove, self)._action_assign()
 
-        self._action_refactor(stage="assign")
+        refactored_moves = self._action_refactor(stage="assign")
```
ws-mps, line 157: Shouldn't this be `refactored_moves = res._action_refactor(stage="assign")` ?

kdwyer: Fixed in https://github.com/unipartdigital/udes-open/pull/745/commits/ff1084d3a25aaa287caf20b0ef2590fbff124df4

ws-mps (approved):

PR #749, opened 2022-12-20
[ADD] udes_stock: Adding a config for current quantity on wharehouse

armandcela:
Story: 3515

### DIFF
--- a/addons/udes_stock/models/stock_warehouse.py
+++ b/addons/udes_stock/models/stock_warehouse.py
@@ -129,6 +129,12 @@ class StockWarehouse(models.Model):
         default=False,
         help="Logs details when picking is added to batch picking",
     )
+    u_show_current_quantity = fields.Boolean(
+        string="Show Current Quantity",
+        default=False,
+        help="Show current quantity on old mobile look up. "
+             "If this is set True the field to compute current quantity on stock should be added",
+    )
 
     @lazy_property
     def reserved_package_name(self):
@@ -191,6 +197,7 @@ class StockWarehouse(models.Model):
             "u_stock_investigation_picking_type": self.u_stock_investigation_picking_type.id,
             "u_show_rpc_timing": self.u_show_rpc_timing,
             "u_allow_create_picking_reserved_package": self.u_allow_create_picking_reserved_package,
+            "u_show_current_quantity": self.u_show_current_quantity,
         }
 
     def get_info(self):


kdwyer (approved):

jrbirch99:
Q: why do we set this attribute at the warehouse level instead of the location level? It feels like (from reading the story) that it should be configurable per location

jrbirch99:
Q: Should this be in the hds repo? On the story they seem to only want this functionality for hds :slightly_smiling_face:

armandcela:
> Q: why do we set this attribute at the warehouse level instead of the location level? It feels like (from reading the story) that it should be configurable per location not just per warehouse

I think reasons are as below:
1. In story doesnt say that some location should have and some not to set it per location.
2. In story says needed only for HDS, thats why need smth to distinguish per client level , which warehouse distinguishes 
3. We may even scan package on lookup which shows products with quantities and we need to show current quantity in that case too.
4. Setting per location means that on every location need to set True or False which doesnt make sense i think

jrbirch99:
> > Q: why do we set this attribute at the warehouse level instead of the location level? It feels like (from reading the story) that it should be configurable per location not just per warehouse
> 
> I think reasons are as below:
> 
>     1. In story doesnt say that some location should have and some not, to set it per location.
> 
>     2. In story says needed only for the customer, thats why need smth to distinguish per client level , which warehouse distinguishes
> 
>     3. We may even scan package on lookup which shows products with quantities and we need to show current quantity in that case too.
> 
>     4. Setting per location means that on every location need to set True or False which doesnt make sense i think

For point 4. You could have it so that if it was set to `True` only a parent location it propagated it to its child locations, which would mean for 2. we could just set it in data on say the "stock" location for HDS. But I think it is fine, as the flag is only used for showing things in the frontend, and we do that per picking type anyway which implies some filtering by location :slightly_smiling_face:

adamrpatrick:
Typo: "Warehouse" instead of "Wharehouse"

addons/udes_stock/views/stock_warehouse.xml
```diff
@@ -54,6 +54,7 @@
                             <field name="u_inventory_adjust_reserved" />
                             <field name="u_allow_create_picking_reserved_package" />
                             <field name="u_log_batch_picking" />
+                            <field name="u_show_current_quantity" />
```
adamrpatrick, line 57: Let's move this into the customer module.

jrbirch99 (approved):

PR #762, opened 2023-02-16
Story #2298 Rockhopper/Gentoo - Check into a Pallet

jrbirch99:
- `find_move_lines` now return move lines for parent packages


PeterAlabaster (approved):

kdwyer (approved):

jrbirch99:
Turns out we can just use `child_of`, so changed it back

armandcela (approved):

PR #763, opened 2023-02-24
[ADD] udes_stock: Visibility of archived rules for procurement routes

armandcela:
Story: 3830

### DIFF
--- a/addons/udes_stock/models/stock_location_route.py
+++ b/addons/udes_stock/models/stock_location_route.py
@@ -6,3 +6,13 @@ class Route(models.Model):
 
     # Disable translation instead of renaming.
     name = fields.Char(translate=False)
+
+    def write(self, values):
+        """
+        When in a route active field is changed, set the context with active_test = False which will
+        ensure that will search even the inactive records and activate/inactivate them where needed.
+        """
+
+        if "active" in values and "active_test" not in self._context:
+            self = self.with_context(active_test=False)
+        return super(Route, self).write(values)
--- a/addons/udes_stock/views/stock_location_route_views.xml
+++ b/addons/udes_stock/views/stock_location_route_views.xml
@@ -27,6 +27,21 @@
         </field>
     </record>
 
+    <record id="view_procurement_rule_tree_inherit" model="ir.ui.view">
+        <field name="name">udes_stock.procurement.rule.tree</field>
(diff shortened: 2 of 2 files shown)


addons/udes_stock/models/stock_location_route.py
```diff
@@ -6,3 +6,13 @@ class Route(models.Model):
 
     # Disable translation instead of renaming.
     name = fields.Char(translate=False)
+
+    def write(self, values):
+        """
+        When a rute active field is changed set the context with active_test = False which will
```
adamrpatrick, line 12: Typo: "rute" --> "route"

adamrpatrick (approved):

addons/udes_stock/models/stock_location_route.py
```diff
@@ -6,3 +6,13 @@ class Route(models.Model):
...
+
+    def write(self, values):
+        """
+        When in a route active field is changed, set the context with active_test = False which will
+        ensure that will search even the inactive records and activate/inactivate them where needed.
+        """
+
+        if "active" in values and "active_test" not in self._context:
+            self = self.with_context(active_test=False)
+        return super(Route, self).write(values)
```
jrbirch99, line 18: nit: `super(Route, self)` -> `super()`

jrbirch99 (approved):

addons/udes_stock/models/stock_location_route.py
armandcela: The reason why i added with self is because i changed self and wanted to make it more visible that will be running as self.
In fact should change anything , but i think we have freedom on this part to use or not self as we are commonly using both ways in our code.

jrbirch99: yeah it won't change anything and I understand it makes it more visible, I just thought we had a rule where we didn't pass in `Route, self` when we didn't need to, as I have been asked to change this a lot in the past. Either way it shouldn't hold up the story as is only a nit :slightly_smiling_face:

jrbirch99: will resolve the thread

armandcela: Thanks, if you can add tag reviewed on the story i would appreciate it.

PR #766, opened 2023-03-03
[IMP] udes_stock: Reserve unbatched pickings in groups.

kdwyer:
story/3625

The stock reservation job processes pickings one at a time, unless the picking is in a batch, in which case all the pickings in the batch are processed together. This works well for environments where batches are used, as it allows us to take into account batch-level concerns, for example atomic batch reservation; however it's inefficient if batches are not used, as stock is reserved one picking at a time.

This patch separates the processing of batched and not-batched pickings, allowing pickings that aren't members of batches to be processed in groups rather than singly.

It has the pleasing side-effect of reducing the number of log messages generated by the job in environments where batches are not used.

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -2787,11 +2787,24 @@ class StockPicking(models.Model):
             processed = Picking.browse()
             by_type = lambda x: x.picking_type_id == picking_type
 
+            # Get the number of pickings that are not in batches.
+            domain = base_domain[:]
+            domain.append(("batch_id", "=", False))
+            num_unbatched = Picking.search(domain, count=True)
+
             while reserve_all or to_reserve > 0:
 
                 if self:
                     # Removed processed pickings from self
                     pickings = self.filtered(by_type) - processed
+                elif num_unbatched > 0:
+                    # Process pickings that are not in batches in groups,
+                    # rather than singly.
+                    domain = base_domain[:]
+                    if processed:
+                        domain.extend([("id", "not in", processed.ids), ("batch_id", "=", False)])
+                    pickings = Picking.search(domain, limit=1000)
+                    num_unbatched -= len(pickings)
                 else:
                     domain = base_domain[:]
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -2787,11 +2787,24 @@ def raise_insufficiency_error(moves):
             processed = Picking.browse()
             by_type = lambda x: x.picking_type_id == picking_type
 
+            # Get the number of pickings that are not in batches.
+            domain = base_domain[:]
+            domain.append(("batch_id", "=", False))
```
armandcela, line 2792:
```suggestion
            domain = expression.AND([domain, [("batch_id", "=", False)]])
```

Its better to get used to use expression.And or expression.Or when appending to an existing domain. If there is any Or operator in existing domain it may not work as expected in some cases.

addons/udes_stock/models/stock_picking.py
```diff
@@ -2787,11 +2787,24 @@ def raise_insufficiency_error(moves):
...
                 if self:
                     # Removed processed pickings from self
                     pickings = self.filtered(by_type) - processed
+                elif num_unbatched > 0:
+                    # Process pickings that are not in batches in groups,
+                    # rather than singly.
+                    domain = base_domain[:]
+                    if processed:
+                        domain.extend([("id", "not in", processed.ids), ("batch_id", "=", False)])
+                    pickings = Picking.search(domain, limit=1000)
```
armandcela, line 2806: If there arent any processed yet domain will not have bath_id = False, and the comment says to process the one not in batches i think

armandcela:
Complex to understand why we need to handle separately pickings not linked with a batch. Will have a look if i can understand smth more about the reason.

kdwyer:
@armandcela The stock reservation code was written primarily to correctly handle pickings in batches. The fact that unbatched pickings are handled individually is a side-effect of the need to handle batches individually rather than a deliberate design decision (if I remember correctly).

PR #767, opened 2023-03-06
Story/3625

kdwyer:


addons/udes_stock/models/stock_picking.py
```diff
@@ -2784,130 +2784,137 @@ def raise_insufficiency_error(moves):
...
-                    )
-                    if not picking_type.u_handle_partials:
-                        # Add to processed to skip reservation.
-                        processed |= pickings
-
-                        if self:
-                            raise_insufficiency_error(unreservable_moves)
+                # "pickings" is plural as the individual picking may be
+                # extended if it's in a batch.
+                for pickings in pickings_to_process:
```
armandcela, line 2811: Read the comment but still think is confusing.
Maybe better for picking in pickings_to_process: and after if-s if pickings is not assigned 
pickings = picking somewhere before line 2829 i think

addons/udes_stock/models/stock_picking.py
```diff
@@ -2784,130 +2784,137 @@ def raise_insufficiency_error(moves):
             to_reserve = picking_type.u_num_reservable_pickings
             reserve_all = to_reserve == -1
             base_domain = [("picking_type_id", "=", picking_type.id), ("state", "=", "confirmed")]
-            limit = 1
+            limit = 100 if reserve_all else picking_type.u_num_reservable_pickings
```
armandcela, line 2787: 100 stands for ?

kdwyer: We're not setting `reserve_all` to -1, you missed an `=` I think.

100 is the number of pickings we will query for if `to_reserve` is -1, which means we must reserve as many pickings as possible.

addons/udes_stock/models/stock_picking.py
kdwyer: I'm not very fond of it either, so I'm willing to go with whatever the team decides.  But I'm not sure I like the idea of changing the name part way through the loop :man_shrugging:

addons/udes_stock/models/stock_picking.py
```diff
@@ -2783,130 +2784,137 @@ def raise_insufficiency_error(moves):
...
 
-                # MPS: mimic Odoo's retry behaviour
-                tries = 0
-                while True:
-
-                    try:
-                        with self.env.cr.savepoint():
-                            pickings = pickings._reserve_stock_assign()
-                            batch._compute_state()
+                    batch = pickings.mapped("batch_id")
```
armandcela, line 2815: It seems like now we are doing batch by batch, maybe we should group by the picks who have batch and who hasnt.

addons/udes_stock/models/stock_picking.py
armandcela: Missed that true

addons/udes_stock/models/stock_picking.py
armandcela: I believe that is very obvious that pickings iterator is not the right name

addons/udes_stock/models/stock_picking.py
armandcela: Why exactly 100 :)

kdwyer: > Why exactly 100 :)

It was @ws-mps's suggestion.

addons/udes_stock/models/stock_picking.py
ws-mps: Probably we can use `picking` here and then use `to_assign` later on instead of pickings.

addons/udes_stock/models/stock_picking.py
```diff
@@ -2783,130 +2784,137 @@ def raise_insufficiency_error(moves):
...
-
-                        if self:
-                            raise_insufficiency_error(unreservable_moves)
-                        continue
-                    elif not moves_todo:
-                        # Add to processed to skip reservation.
-                        processed |= pickings
+                # "pickings" is plural as the individual picking may be
+                # extended if it's in a batch.
+                for pickings in pickings_to_process:
```
armandcela, line 2811: I think we should break in case to_reserve is less than 0 inside the for loop.
Currently it seems like if limit is 20 because of u_num_reservable_pickings and suppose all pickings are different batches we go in this for loop and reserve all batches one by one without breaking in case pickings reserved have passed the limit.
Maybe a while loop and to add in the while to_reserve to would be a suggestion

armandcela: Or maybe at line 2917 another if if to_reserve <= 0: break the for loop

addons/udes_stock/models/stock_picking.py
```diff
@@ -2783,130 +2784,137 @@ def raise_insufficiency_error(moves):
...
+                            continue
+
+                    # MPS: mimic Odoo's retry behaviour
+                    tries = 0
+                    while True:
+
+                        try:
+                            with self.env.cr.savepoint():
+                                pickings = pickings._reserve_stock_assign()
+                                batch._compute_state()
```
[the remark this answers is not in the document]
kdwyer: I added `assert not batch` after this line and got 10 test failures, so I don't believe that the batch is empty here.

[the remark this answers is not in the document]
armandcela: Maybe all pickings that you have set up in unit test have batches. Try with a picking without batch :)

addons/udes_stock/models/stock_picking.py
kdwyer: Applied @ws-mps's suggestion in 8da3aba7

addons/udes_stock/models/stock_picking.py
kdwyer: I just indented the existing code.  Fixed by 1f0282b2a.

addons/udes_stock/models/stock_picking.py
kdwyer: Fixed by e62d7c743.

armandcela (approved):

jrbirch99 (approved):

PR #768, opened 2023-03-06
Story/3390 Update Gentoo Functionality

jrbirch99:
- Added stock.picking.print.strategy functionality
- Added the ability to print reports when a user is assigned a picking


addons/udes_stock/models/stock_picking.py
```diff
@@ -860,6 +860,14 @@ def _action_done(self):
         with self.statistics() as stats:
             res = super()._action_done()
 
+            # Trigger the Print Job for the done pickings
+            self.env.ref("udes_stock.picking_done").with_context(
+                active_model=self._name,
+                active_ids=self.ids,
+                action_filter="picking._action_done",
+                skip_printing=True if self.picking_type_id.u_disable_automated_printing else False
```
kdwyer, line 868: Same as above

addons/udes_stock/README.md
```diff
@@ -278,6 +278,25 @@ Model used to register users that can access the system in the way has been defi
 | assign_picking_to_users       | Assigning a picking to users                                 |
 | _unassign_pickings_from_user  | Unassigning pickings from users                              |
 
+### Stock Picking Print Strategy
+This model contains records for printing during pickings. It inherits from the "print.strategy" model in "print". In the code there is
```
kdwyer, line 282: "In the code there_are_ hardcoded calls..."

armandcela:
Before reviewing i would suggest to remove responsibilities from udes_stock maybe to create another module for printing and to add required dependences. 

Was one of the improvements we discussed in udes14 to make it lighter udes_stock

jrbirch99:
> Before reviewing i would suggest to remove responsibilities from udes_stock maybe to create another module for printing and to add required dependences.
> 
> Was one of the improvements we discussed in udes14 to make it lighter udes_stock

Yep, just discussed with @adamrpatrick, and it will be moved into its own module. I will close this PR and make @kdwyer changes as I create the new module.

PR #769, opened 2023-03-07
[TEST] udes_stock: Parameterised stock reservation tests.

kdwyer:
story/3625

Add unit tests for StockPicking.reserve_stock to test the outcoms described in the table below.

+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
| Batch  | Reserve Batch | Handle Partial |    Stock    |    No Stock     | Line Not Available  | Line Partially Available |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   Y    |       Y       |        Y       | reserve all | reserve nothing | reserve available   | reserve available        |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   Y    |       Y       |        N       | reserve all | reserve nothing | reserve nothing     | reserve nothing          |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   Y    |       N       |        Y       | reserve all | reserve nothing | reserve available   | reserve available        |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   Y    |       N       |        N       | reserve all | reserve nothing | reserve fully       | reserve fully            |
|        |               |                |             |                 | reservable pickings | reservable pickings      |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   N    |       Y       |        Y       | reserve all | reserve nothing | reserve available   | reserve available        |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   N    |       Y       |        N       | reserve all | reserve nothing | reserve nothing     | reserve nothing          |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   N    |       N       |        Y       | reserve all | reserve nothing | reserve available   | reserve available        |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+
|   N    |       N       |        N       | reserve all | reserve nothing | reserve nothing     | reserve nothing          |
+--------+---------------+----------------+-------------+-----------------+---------------------+--------------------------+

Headings:

- Batch: whether a picking is in a batch (we assume that either all pickings are in batches or none of them are).
- Reserve Batch: value of StockPickingType.u_reserve_batch
- Handle Partial: value of StockPickingType.u_handle_partials
- Stock: there exists enough stock for the all pickings
- No Stock: there is no stock for any pickings
- Line Not Available: there is no stock for one move line
- Line partially Available: there is enough stock for only part of one move line

### DIFF
--- a/addons/udes_stock/tests/test_stock_reservation.py
+++ b/addons/udes_stock/tests/test_stock_reservation.py
@@ -5,6 +5,7 @@ The method itself cannot be tested because it contains a commit statement, but
 we can test related methods.
 """
 import logging
+from unittest import mock
 from .common import BaseUDES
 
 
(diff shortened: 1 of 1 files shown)


addons/udes_stock/tests/test_stock_reservation.py
```diff
@@ -95,3 +96,222 @@ def test_logs_debug_message_if_more_one_picking_processed(self):
...
+
+# Test configuration, stock levels and expected outputs, keyed on:
+# Pickings are in batches, PickingType.u_reserve_batches, PickingType.u_handle_partials.
+# Values order: all available, none available, partially available, line partially available
+# In-values order:
+# quantity in stock, expected reservation
+# each picking has two identical lines (2 x 10)
+# each batch has two identical pickings
+RESERVATION_MATRIX = {
+    (True, True, True): [(40, 40), (0, 0), (30, 30), (35, 35)],
```
armandcela, line 122: Seems like could use binary in mapping 7,6,5,4,3,2,1,0 and convert to binary , maybe no need for classes but a loop in range (8)

addons/udes_stock/tests/test_stock_reservation.py
```diff
@@ -95,3 +96,222 @@ def test_logs_debug_message_if_more_one_picking_processed(self):
...
+
+        with mock.patch.object(self.pick.env.cr, "commit", return_value=None):
+            self.pick.reserve_stock()
+
+        quant = self.Quant.search([("reserved_quantity", ">", 0)])
+        self.assertEqual(quant.reserved_quantity, expected)
+
+    def test_handles_partially_available_lines(self):
+        """The system will correctly handle the case when a line is partially available."""
+        initial_qty, expected = self.config[2]
```
armandcela, line 169: Config mostly is getting the third element while when defined it seems has 4 elements. Do we need the fourth one or i am missing smth

kdwyer: Each value in the `RESERVATION_MATRIX` dict consists for four tuples, that correspond to the states of 

- enough stock to reserve
- no stock to reserve
- enough stock for all lines except one
- enough stock for ll lines except one that we can partially reserve

Each of the tests in `ReservationMixin` tests one of these states, and fetches the corresponding element element from the value list.

addons/udes_stock/tests/test_stock_reservation.py
```diff
@@ -270,48 +273,45 @@ def setUpClass(cls):
...
-        with mock.patch.object(self.pick.env.cr, "commit", return_value=None):
-            self.pick.reserve_stock()
-
-        quant = self.Quant.search([("reserved_quantity", ">", 0)])
-        self.assertEqual(quant.reserved_quantity, expected)
+        for value, expected in zip(values, expected):
+            with self.subTest(u_num_reservable_pickings=value):
+                with self.savepoint():
+                    self.picking_type_pick.u_num_reservable_pickings = value
+                    quant = self.create_quant(self.apple.id, self.test_location_01.id, qty=30)
```
kdwyer, line 306: I can probably hoist the quant and picking creation outside of the loop?

kdwyer:
I could probably run each row of the reservation matrix in a single subtest with a custom error message for each "column". Then `ReservationMixin` would have only one test method. But perhaps  that's too much abstraction?

PeterAlabaster (approved):

addons/udes_stock/tests/test_stock_reservation.py
```diff
@@ -5,6 +5,7 @@
 we can test related methods.
 """
 import logging
+from unittest import mock
```
jrbirch99, line 9: Typo in the commit `outcoms` should be `outcomes`

addons/udes_stock/tests/test_stock_reservation.py
kdwyer: Ah I see - that last test should be `[3]` (thanks @ws-mps )

armandcela (approved):

kdwyer:
> I could probably run each row of the reservation matrix in a single subtest with a custom error message for each "column". Then `ReservationMixin` would have only one test method. But perhaps that's too much abstraction?

It turns out that doing this and using a savepoint to revert the changes after each iteration of the subtest reduces the overall duration of the tests in this file from ~60s to ~40s, so I will make this change.

addons/udes_stock/tests/test_stock_reservation.py
kdwyer: Fixed by 5e286d88

addons/udes_stock/tests/test_stock_reservation.py
kdwyer: Fixed by 0c72542e2, but I've since reworked that test case so that it hast just one method which loops over the config values.

PR #775, opened 2023-03-28
Story #4035 Desktop backorder refactoring bug

jrbirch99:
- Add a context flag to skip refactoring

### DIFF
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -24,8 +24,8 @@ class StockMove(models.Model):
     u_grouping_key = fields.Char("Key", compute="compute_grouping_key")
 
     def _unreserve_initial_demand(self, new_move):
-        """ Override stock default function to keep the old move lines,
-            so there is no need to create them again.
+        """Override stock default function to keep the old move lines,
+        so there is no need to create them again.
         """
         self.mapped("move_line_ids").filtered(lambda x: x.qty_done == 0.0).write(
             {"move_id": new_move, "product_uom_qty": 0}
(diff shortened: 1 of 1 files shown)


armandcela (approved):

PeterAlabaster (approved):

addons/udes_stock/models/stock_move.py
```diff
@@ -687,3 +689,16 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
                 new_pickings += picking._backorder_movelines(mls, dest_picking)
 
         return self | new_pickings.mapped("move_lines")
+
+    def _split(self, qty, restrict_partner_id=False):
+        """
+        If the moves have movelines we do not want to refactor the moves when we split them, as
```
PeterAlabaster, line 695:
```suggestion
        Extend _split to disable move refactoring when the move has move_lines linked to it
        ..then something here as to why, i don't follow what we mean by `as the movelines will not be refactored`
```

addons/udes_stock/models/stock_move.py
```diff
@@ -687,3 +687,14 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
                 new_pickings += picking._backorder_movelines(mls, dest_picking)
 
         return self | new_pickings.mapped("move_lines")
+    
+    def _split(self, qty, restrict_partner_id=False):
+        """
+        If the moves have movelines we do not want to refactor the moves when we split them, as
+        the movelines will not be refactored.
+        """
+        if self.mapped("move_line_ids"):
```
armandcela, line 698: It looks like self is always 1 record as even in core odoo is treated like that self.product_uom_qty etc.
I think don't need the mapped in this case,

addons/udes_stock/models/stock_move.py
jrbirch99: changed :slightly_smiling_face:

addons/udes_stock/models/stock_move.py
jrbirch99: changed :slightly_smiling_face:

PR #776, opened 2023-03-31
[IMP] udes_suggest_location: Tidy up tests.

kdwyer:
story/2419-tidyup


PeterAlabaster (approved):

addons/udes_suggest_location/tests/test_stock_move_line.py
```diff
@@ -6,14 +6,16 @@
...
-class TestStockMoveLine(common.SuggestedLocations):
+class CommonTestCase(common.SuggestedLocations):
+    """Tests that don't depend on specific policy behaviours."""
+
     @classmethod
     def setUpClass(cls):
-        super(TestStockMoveLine, cls).setUpClass()
+        super().setUpClass()
         cls.MoveLine = cls.env["stock.move.line"]
         # Set policies, use by_product for pick
```
jrbirch99, line 16: I think this comment needs changing to `by_empty_location`

addons/udes_suggest_location/tests/test_stock_move_line.py
```diff
@@ -57,42 +60,34 @@ def test_get_policy_class(self):
         # Fetch by product policy
```
jrbirch99, line 60: same here `by product` -> `by empty location`

addons/udes_suggest_location/tests/test_stock_move_line.py
```diff
@@ -105,48 +100,44 @@ def test_suggest_locations_errors_with_no_location_policy(self):
...
+        pick = self.create_picking(
+            self.picking_type_internal,
+            products_info=[{"product": self.apple, "qty": 5}],
+            confirm=True,
+            assign=True,
+        )
+        # Check mls has correct product
+        self.assertEqual(pick.move_line_ids.product_id, self.apple)
+
+    @unittest.skip("Fails for by_empty_location")
```
jrbirch99, line 125: Do we have stock in `self.test_goodsout_location_02`? At the moment I don't think we have set the test up to produce an invalid destination location when setting the policy to `by_empty_location`, as `self.test_goodsout_location_02` is actually empty and therefore a suggestible location

jrbirch99 (commented):
Just from scanning the code, I cannot see a situation when the tests actually use `cls.mock_validate_location_dest`, so I think we could remove it. `cls.mock_suggested_location` is used in test `test_validate_location_dest_view`, but I think we could change the test to check that `None` is returned when `validate_location_dest()` is called so that we do not have to use mocks.

addons/udes_suggest_location/tests/test_stock_move_line.py
kdwyer: There's no stock there, but I'm surprised that we can suggest a location that is not a child of `picking_type.default_location_dest_id`.

jrbirch99: yeah, I am not sure if there will be situation when we don't want that condition to be true. Maybe an issue does need to be created for it? It should be a simple change I think, just adding another check in `validate_location_dest()`

ws-mps: When creating the picking we can set any destination location, `picking_type.default_location_dest_id` is only in case `location_dest_id` is not set. Therefore, `test.test_goodsout_location_02` is (in general) a valid destination location for any new picking regarding its picking type.

kdwyer: Fixed by passing a destination location that already contains stock.

jrbirch99 (approved):

kdwyer:
> Just from scanning the code, I cannot see a situation when the tests actually use `cls.mock_validate_location_dest`, so I think we could remove it. `cls.mock_suggested_location` is used in test `test_validate_location_dest_view`, but I think we could change the test to check that `None` is returned when `validate_location_dest()` is called so that we do not have to use mocks.

On balance I think it's reasonable to keep these tests. They demonstrate which responsibilities lie with `StockMoveLine` rather than the policy objects.

PR #777, opened 2023-04-03
[IMP] udes_stock: Improve mark_picked_lines to handle multiple lots

armandcela:
with different quantities

Story: 3427

### DIFF
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -12,21 +12,21 @@ class ProductProduct(models.Model):
     # Add tracking for archiving.
     active = fields.Boolean(tracking=True)
 
-    def assert_serial_numbers(self, serial_numbers):
+    def assert_tracking_unique(self, serial_numbers):
         """
-        If the product in self is tracked by serial numbers, check if
+        If the product in self is tracked, check if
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/product_product.py
```diff
@@ -14,19 +14,19 @@ class ProductProduct(models.Model):
...
     def assert_serial_numbers(self, serial_numbers):
         """
-        If the product in self is tracked by serial numbers, check if
+        If the product in self is tracked, check if
         the ones in serial_numbers are currently in use in the system.
         """
         Lot = self.env["stock.production.lot"]
         self.ensure_one()
-        if self.tracking == "serial":
+        if self.tracking != "none":
```
adamrpatrick, line 22: Given that we're now asserting any kind of tracking, not just serial numbers, we should also change the name of the method e.g. to `assert_tracking_unique`. 
Q. In which case would we want a lot number to be unique?

armandcela: In all cases product_id lot number is unique for serial products and for products tracked with lot.
I agree about naming the method, you are right

adamrpatrick (approved):

PR #779, opened 2023-04-13
[ADD] udes_stock_refactoring: Refactoring method returns consistency …

armandcela:
…moves

Story: 3710

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -636,6 +636,7 @@ class StockPicking(models.Model):
                 product = product_info.get("product")
                 uom_qty = product_info.get("uom_qty") or product_info.get("qty")
                 uom_id = product_info.get("uom_id") or product.uom_id.id
+                expected_date = product_info.get("expected_date")
                 vals = {
                     "product_id": product.id,
                     "name": product.name,
@@ -648,6 +649,10 @@ class StockPicking(models.Model):
                     "picking_type_id": picking.picking_type_id.id,
                     "description_picking": product._get_description(picking.picking_type_id),
                 }
+                if expected_date:
+                    vals.update({
+                        "date": expected_date,
+                    })
                 vals.update(kwargs)
                 move_values.append(vals)
         return move_values
--- a/addons/udes_stock_refactoring/tests/test_refactoring.py
+++ b/addons/udes_stock_refactoring/tests/test_refactoring.py
@@ -1,5 +1,6 @@
 from odoo.addons.udes_stock.tests import common
 from odoo.osv import expression
+import random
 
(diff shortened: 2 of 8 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -636,6 +636,7 @@ def _prepare_move(self, pickings, products_info, **kwargs):
                 product = product_info.get("product")
                 uom_qty = product_info.get("uom_qty") or product_info.get("qty")
                 uom_id = product_info.get("uom_id") or product.uom_id.id
+                expected_date = product_info.get("expected_date")
```
ws-mps, line 639: I would move this in its own commit, so we have the tests in its own commit and changes on two different modules in separate commits.

addons/udes_stock_refactoring/tests/test_refactoring.py
```diff
@@ -840,3 +841,66 @@ def test_disable_move_refactor_context(self):
...
+                    # Set the date here because create_picking creates the picking
+                    # without moves then adds the moves, overwriting the `scheduled_date`.
+                    dict(product=product, qty=1, expected_date=date)
+                    for product in [self.apple, self.banana]
+                ],
+            )
+            for date, sequence in picks
+        }
+
+        self.picks = Picking.union(*(self.picks_by_key.values()))
```
kdwyer, line 889: Nice, I didn't realise you could do this instead of 

```
picks = Pick.browse()
for p in some_picks:
    picks |= p
```

addons/udes_stock_refactoring/tests/test_refactoring.py
```diff
@@ -840,3 +841,66 @@ def test_disable_move_refactor_context(self):
...
+        picks = self.picks.filtered(
+            lambda x: x.picking_type_id == self.picking_type_pick
+        )
+        batches_before_refactoring = picks.batch_id
+        self.assertEqual(len(batches_before_refactoring), 0)
+        # It returns the moves sent to refactor
+        refactored_moves = picks.move_lines._action_refactor()
+        self.assertEqual(refactored_moves, picks.move_lines)
+        batches_after_refactoring = picks.batch_id
+        self.assertEqual(len(batches_after_refactoring), 2)
```
kdwyer, line 906: Should we be asserting that the results are actually grouped by date?

armandcela: I just checked that it refactored in 2 batches, just to test that refactored worked. I think was not part of this story to check that refactoring works as expected. if needed can be added more asserts later

kdwyer (approved):

ws-mps (approved):

PR #787, opened 2023-06-29
Story/4528

kdwyer:


addons/udes_stock_refactoring/models/stock_location.py
```diff
@@ -0,0 +1,40 @@
...
+class StockLocation(models.Model):
+    """Stock Location model."""
+
+    _inherit = "stock.location"
+
+    u_is_picking_zone = fields.Boolean(
+        string="Is A Picking Zone",
+        help="Picking Zones are the level to which warehouse-wide Picks are broken down.",
+    )
+    # TODO do we need an ondelete here?
```
ws-mps, line 15: I don't think so

addons/udes_stock_refactoring/models/stock_location.py
```diff
@@ -0,0 +1,40 @@
...
+
+from odoo import fields, models
+
+
+class StockLocation(models.Model):
+    """Stock Location model."""
+
+    _inherit = "stock.location"
+
+    u_is_picking_zone = fields.Boolean(
```
ws-mps, line 11: This could go to `udes_stock` module, since could be a useful field for other modules.

addons/udes_stock_refactoring/models/stock_location.py
```diff
@@ -0,0 +1,40 @@
...
+        If a record is a picking zone, its is set to itself, otherwise successive parent
+        locations are checked, and the first one which is a picking zone is set.
+        If no picking zone is found, False is set.
+        """
+        for record in self:
+            record.u_picking_zone_id = False
+            location = record
+            while location and not record.u_picking_zone_id:
+                if location.u_is_picking_zone:
+                    record.u_picking_zone_id = location
```
JRBirch, line 38: `record.u_picking_zone_id = location.id`

jrbirch99, line 38: `record.u_picking_zone_id = location.id`

addons/udes_stock_refactoring/models/stock_location.py
```diff
@@ -0,0 +1,40 @@
...
+
+        A picking zone is a location with u_is_picking_zone == True.
+        If a record is a picking zone, its is set to itself, otherwise successive parent
+        locations are checked, and the first one which is a picking zone is set.
+        If no picking zone is found, False is set.
+        """
+        for record in self:
+            record.u_picking_zone_id = False
+            location = record
+            while location and not record.u_picking_zone_id:
```
PeterAlabaster, line 36: I'm not sure if it's faster, or exactly the same behaviour, but maybe instead of a while loop, we could do a `Location.search_read([("u_is_picking_zone", "=", True), ("id", "parent_of", record.id)], ["id"])`
Wouldn't work if we need to be able to set multiple picking zone ids within a hierarchy though

addons/udes_stock_refactoring/tests/test_stock_location.py
```diff
@@ -0,0 +1,30 @@
+"""Unit trests for the Stock Location model."""
```
PeterAlabaster, line 1:
```suggestion
"""Unit tests for the Stock Location model."""
```

addons/udes_stock_refactoring/views/stock_location.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+  <data>
+
+    <!-- Changes to stock.view_location_form -->
+    <record id="view_location_form" model="ir.ui.view">
+      <field name="name">Changes to location form</field>
```
PeterAlabaster, line 7:
```suggestion
      <field name="name">udes_stock_refactoring.stock.location.form</field>
```

addons/udes_stock_refactoring/views/stock_location.xml
```diff
@@ -0,0 +1,34 @@
...
+        <xpath expr="//field[@name='u_limit_orderpoints']" position="after">
+          <field name="u_is_picking_zone" attrs="{'invisible': [('usage', '!=', 'view')]}" />
+          <field name="u_picking_zone_id" attrs="{'invisible': [('u_picking_zone_id', '=', False)]}"/>
+        </xpath>
+      </field>
+    </record>
+
+    <!-- Changes to search dropdown -->
+    <record id="view_location_search" model="ir.ui.view">
+      <field name="name">udes_stock.stock.location.search</field>
```
PeterAlabaster, line 21:
```suggestion
      <field name="name">udes_stock_refactoring.stock.location.search</field>
```

PeterAlabaster (changes requested):
Couple nits, one suggestion that may or may not be better

addons/udes_stock_refactoring/tests/test_refactoring.py
```diff
@@ -1078,3 +1078,160 @@ def test_split_mixed_products_partial_reserve(self):
...
+    def test_refactors_by_picking_zone(self):
+        """The system will refactor by picking zone."""
+        origin = "TEST_ORIGIN"
+        # One picking, stock in each zone.
+        products_info = [{"product": self.apple, "qty": 1}, {"product": self.cherry, "qty": 1}]
+        original_picking = self.create_picking(
+            self.picking_type_pick, products_info=products_info, origin=origin, confirm=True
+        )
+
+        original_picking.action_assign()
```
jrbirch99, line 1125: You should be able to pass in `assign=True` to `create_picking()` and it should product an assigned picking

addons/udes_stock_refactoring/models/stock_location.py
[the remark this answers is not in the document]
PeterAlabaster: I see miquel has suggestion we don't need one, so happy to resolve this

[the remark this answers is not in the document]
ws-mps: I think default is set null, so that should be enough.

addons/udes_stock_refactoring/__manifest__.py
```diff
@@ -7,14 +7,13 @@
     "license": "LGPL-3",
     "application": True,
     "version": "14.0.2.1.0",
-    "depends": [
-       "base", "stock", "udes_common", "stock_picking_batch", "udes_stock"
-    ],
+    "depends": ["base", "stock", "udes_common", "stock_picking_batch", "udes_stock"],
```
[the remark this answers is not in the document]
kdwyer: -0 (I'd rather let black handle it if we exceed max line length) but ok :smiley: .

addons/udes_stock_refactoring/tests/test_refactoring.py
kdwyer: Ah I was asserting the length of `original_picking` at first, but then removed as I don't like asserting pre-conditions.  I'll change.

addons/udes_stock_refactoring/models/stock_location.py
kdwyer: Will do (both fields will need to move, as refactoring depends on stock).

addons/udes_stock_refactoring/models/stock_location.py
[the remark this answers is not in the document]
kdwyer: I was thinking `'set null'`, FWIW.

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
PeterAlabaster: if you change the line to `"depends": ["base", "stock", "udes_common", "stock_picking_batch", "udes_stock",],` does black autoformat it to line per list item though?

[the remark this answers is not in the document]
PeterAlabaster: It's preference, i just like it cause it makes it easier to tell what changed on the MR, but i'm happy to let it go :smile:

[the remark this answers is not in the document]
kdwyer: > if you change the line to `"depends": ["base", "stock", "udes_common", "stock_picking_batch", "udes_stock",],` does black autoformat it to line per list item though?

Yes:
```
$ black -c '[1, 2, 3,]'
[
    1,
    2,
    3,
]
```

addons/udes_stock_refactoring/tests/test_stock_location.py
kdwyer: Fixed by 1e344ffe

addons/udes_stock_refactoring/__manifest__.py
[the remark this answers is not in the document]
kdwyer: Fixed by 29c9c691

addons/udes_stock_refactoring/models/stock_location.py
kdwyer: Removed in 67683c12

addons/udes_stock_refactoring/models/stock_location.py
kdwyer: Implemented in 0b5369ae

jrbirch99 (approved):

addons/udes_stock_refactoring/models/stock_location.py
kdwyer: Addressed in ff893c2bf7.
(I'll split out the stock and refactoring changes if we are happy with this commit).

PeterAlabaster (approved):

adamrpatrick (approved):

PR #788, opened 2023-07-06
[ADD] udes_stock: Merging move lines

armandcela:
Story: 4416

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -4,6 +4,8 @@ import logging
 
 from odoo import models, fields, api, _
 from odoo.tools.float_utils import float_compare, float_round
+from itertools import groupby
+from operator import itemgetter
 from odoo.osv import expression
 from odoo.exceptions import ValidationError, UserError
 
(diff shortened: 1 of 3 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -621,3 +623,72 @@ def _split_and_group_mls_by_quantity(self, maximum_qty):
...
+            move_lines_to_unlink.sudo().unlink()
+        return (self | MoveLine.concat(*move_lines_to_merge)) - move_lines_to_unlink
+
+    @api.model
+    def _prepare_merge_move_lines_distinct_fields(self):
+        """
+        Prepare merge move lines distinct fields
+        """
+        return [
+            "product_id", "location_id", "location_dest_id", "package_id", "lot_id", "state"
```
ws-mps, line 673: Do we need state? all move lines of the picking should be in the same state.

Do we want move_id? we could have 2 move lines with the same info but different move_id, so they cannot be merged.

armandcela: Added move_id as suggested, added even _merge_moves because noticed that was created a new move for same product. Can you have a look please?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -621,3 +623,72 @@ def _split_and_group_mls_by_quantity(self, maximum_qty):
             remaining_mls -= mls_used
             grouped_mls.append(mls_used)
         return grouped_mls
+
+    def _merge_move_lines(self):
+        """ This method will, for each move line in `self`, go up in their linked picking and try to
+        find in their existing move line a candidate into which we can merge the move.
+        :return: Recordset of move lines passed to this method.
+        If some of passed move lines were merged into another existing one, return this one and
+        not the (now unlinked) original.
```
jrbirch99, line 632: nit: Can we move this comment above `:return: Recordset of move lines passed to this method`?

addons/udes_stock/models/stock_move_line.py
```diff
@@ -621,3 +623,72 @@ def _split_and_group_mls_by_quantity(self, maximum_qty):
...
+        """
+        MoveLine = self.env["stock.move.line"]
+        distinct_fields = self._prepare_merge_move_lines_distinct_fields()
+
+        candidate_move_line_list = []
+        for picking in self.mapped("picking_id"):
+            candidate_move_line_list.append(picking.move_line_ids)
```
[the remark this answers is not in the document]
armandcela: I referred to the merge_moves method, and just changed some variable names. Yes i think is better , and maybe no need for mapped as well

addons/udes_stock/models/stock_move_line.py
jrbirch99: oh I see, this is part of the description of the `:return:`. I will resolve then :slightly_smiling_face:

jrbirch99 (approved):

kdwyer (approved):

PR #789, opened 2023-08-01
Story #4725 UDES 14 - All flows to handle floats

jrbirch99:
- Updating various parts of the backend code to handle floats

### DIFF
--- a/addons/udes_stock_picking_batch/models/stock_move_line.py
+++ b/addons/udes_stock_picking_batch/models/stock_move_line.py
@@ -80,11 +80,9 @@ class StockMoveLine(models.Model):
 
             task["type"] = "package"
             task["package_id"] = info[0]
-            
+
         if picking.picking_type_id.u_reserve_pallet_per_picking:
-            task.update({
-                "reserved_pallet": picking.u_reserved_pallet
-            })
+            task.update({"reserved_pallet": picking.u_reserved_pallet})
 
         return task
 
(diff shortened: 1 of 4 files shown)


addons/udes_stock/models/uom_uom.py
```diff
@@ -0,0 +1,18 @@
+import decimal
+from odoo import api, fields, models
+
+
+class UoM(models.Model):
+    _inherit = "uom.uom"
+
+    decimal_places = fields.Integer(
```
armandcela, line 8: By standard field should have u_ at the beginning

addons/udes_stock_picking_batch/models/stock_move_line.py
```diff
@@ -62,7 +62,7 @@ def _prepare_task_info(self):
             task["type"] = "product"
             task["pick_quantity"] = sum(self.mapped("product_qty"))
             task["initial_demand"] = sum(self.move_id.mapped("u_uom_initial_demand"))
-            task["quant_ids"] = self.get_quants().get_info(level=2)
+            task["quant_ids"] = self.get_quants().get_info(level=2, extra_fields={"decimal_places"})
```
armandcela, line 65: I think decimal places is quite important to have on get_info of uom.uom so maybe better to add in get_info_fields and you dont need to specify in extra_fields

addons/udes_stock/models/uom_uom.py
jrbirch99: I think I am going to move this to `udes_uom` too

armandcela (approved):

PeterAlabaster (approved):

PR #790, opened 2023-08-18
[IMP] Strip EXIF data from image after upload

JayshreeMahajan:
user-story:4777


addons/udes_security/models/ir_attachment.py
```diff
@@ -101,8 +102,34 @@ def create(self, vals_list):
...
+            if "attachment" in vals and "attachment_filename" in vals:
+                processed_attachment_data = self._process_attachment_data(
+                    vals["attachment"], vals["attachment_filename"]
+                )
+                vals["attachment"] = processed_attachment_data
         return super().create(vals_list)
 
+    def _process_attachment_data(self, attachment_data, attachment_filename):
+        # Check if the attachment is an image based on its file extension
+        allowed_image_extensions = [".jpg", ".jpeg", ".png"]
```
kdwyer, line 114: Apparently (at least according to Wikipedia) `.webp` images support EXIF, so we should probably include those here too.

addons/udes_security/tests/test_stripped_exif_data.py
```diff
@@ -0,0 +1,41 @@
...
+    @classmethod
+    def create_attachment(cls, filename):
+        file_path = cls._get_file_path(filename)
+
+        attachment = base64.b64encode(file_path.read_bytes())
+
+        return attachment
+
+    def test_remove_exif_data(self):
+        attachment = self.create_attachment("image_with_exif_data.jpg")
```
kdwyer, line 35: It might be useful to assert that the original image _has_ exif data, since it isn't obvious unless we open the image in some kind of exif-reading app.

kdwyer (changes requested):

jrbirch99:
The commit needs to include the module name: 
```
[IMP] udes_security: Strip EXIF data from image after upload
```

addons/udes_security/tests/__init__.py
```diff
@@ -3,3 +3,4 @@
 from . import test_trusted_user
 from . import test_no_redirect
 from . import test_password_complexity  # noqa: F401
+from . import test_stripped_exif_data
```
jrbirch99, line 6: nit: no newline at the end of the file. You need to run black formatting here

PR #791, opened 2023-08-23
Story/4776

rajeeshknambiar:
udes_security: strip dangerous HTML tags from "Send message", "Log note", "Conversations" and email

### DIFF
--- a/addons/udes_security/models/mail_thread.py
+++ b/addons/udes_security/models/mail_thread.py
@@ -1,0 +1,34 @@
+from odoo import models, fields, _, api
+from odoo.exceptions import AccessError, UserError, ValidationError
+
+from lxml.html.clean import Cleaner
+
+import logging
+_logger = logging.getLogger(__name__)
+
+HTML_TAGS_PREVENTED = ['a', 'script']
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _message_create(self, values_list):
+        """Extend _message_create to strip specified tags from the message body"""
+        if not isinstance(values_list, (list)):
+            values_list = [values_list]
+
+        for vals in values_list:
+            body_content = vals.get('body', False)
+            #NOTE: not easy to distinguish email/Send message/Log note
+            #Strip prevented html tags before storing to DB
+            if '<' in body_content and '>' in body_content:
+                vals['body'] = self._strip_prevented_html_tags(body_content)
+
+        return super(MailThread, self)._message_create(values_list)
+
+    def _strip_prevented_html_tags(self, html_text):
+        """
+        Remove specified tags from a html text.
+        param: html_text: str() representation of html contents.
+        """
+        cleaner = Cleaner(javascript=True, scripts=True, remove_tags=HTML_TAGS_PREVENTED)
+        return cleaner.clean_html(html_text)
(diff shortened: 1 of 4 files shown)


addons/udes_security/models/mail_thread.py
```diff
@@ -0,0 +1,73 @@
...
+        for vals in values_list:
+            body_content = vals.get('body', False)
+            #NOTE: message_type of email is also 'comment'! Check email specific fields
+            is_email = vals.get('message_type', False) == 'email' or \
+                       vals.get('mail_server_id', False)
+            #Strip html tags from 'Send message', 'Log note', 'Conversations'; except email
+            if body_content and not is_email:
+                vals['body'] = self._strip_html_tags(body_content)
+
+        return super(MailThread, self)._message_create(values_list)
```
adamrpatrick, line 27: Looks like we have test failures because we also add messages into mail.thread from other parts of the system. Stripping HTML at this level could mean that we're also losing that functionality (as per the unit tests). Can we strip the HTML closer in the stack trace to "Conversations", "Log note", "Send message" etc to avoid any undesirable impacts?

addons/udes_security/models/mail_thread.py
```diff
@@ -0,0 +1,29 @@
...
+
+import logging
+_logger = logging.getLogger(__name__)
+
+HTML_TAGS_PREVENTED = ['a', 'script']
+
+class MailThread(models.AbstractModel):
+    _inherit = "mail.thread"
+
+    def _message_create(self, values_list):
```
PeterAlabaster, line 14:
```suggestion
    def _message_create(self, values_list):
        """Extend _message_create to strip specified tags from the message body"""
```

addons/udes_security/models/mail_thread.py
```diff
@@ -0,0 +1,29 @@
...
+        for vals in values_list:
+            body_content = vals.get('body', False)
+            #NOTE: not easy to distinguish email/Send message/Log note
+            #Strip prevented html tags before storing to DB
+            if '<' in body_content and '>' in body_content:
+                vals['body'] = self._strip_prevented_html_tags(body_content)
+
+        return super(MailThread, self)._message_create(values_list)
+
+    def _strip_prevented_html_tags(self, html_text):
```
PeterAlabaster, line 27:
```suggestion
    def _strip_prevented_html_tags(self, html_text):
        """
        Remove specified tags from a html text.
        param: html_text: str() representation of html contents.
        """
```

addons/udes_security/static/src/js/mail/utils.js
```diff
@@ -0,0 +1,30 @@
...
+	*/
+        return node.textContent;
+    }
+    if (node.tagName === "A") return node.outerHTML;
+    transformChildren();
+    return node.outerHTML;
+  }
+
+  // Override mail.utils function
+  mail_utils.addLink = udes_addLink;
```
PeterAlabaster, line 28: Ah, Odoo JS fun :smile:

addons/udes_security/static/src/js/mail/utils.js
```diff
@@ -0,0 +1,30 @@
+odoo.define("udes_security.mail_utils", function (require) {
+  "use strict";
+
+  var mail_utils = require("mail.utils");
+
+  function udes_addLink(node, transformChildren) {
+    if (node.nodeType === 3) {  // text node
+	/* Don't add automatic link to url text
```
PeterAlabaster, line 8: Is this comment/commented code needed?

rajeeshknambiar: Yes, absolutely! Otherwise odoo will add back link to the plain text urls when displaying in the UI/frontend.

adamrpatrick (approved):

PR #792, opened 2023-08-24
[SECURITY] - udes_security: hide "schedule activity" for non admin users

vvasoyav:
security related chages in udes_security to hide "schedule activity" button for non-admin users only.

Story/4774

Signed-off-by: vishal.vasoya@unipart.com


addons/udes_security/static/src/xml/custom_chatter_topbar.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates>
+    <t t-name="ChatterTopbarExtension" t-inherit="mail.ChatterTopbar" t-inherit-mode="extension" owl="1">
+        <xpath expr="//button[hasclass('o_ChatterTopbar_buttonScheduleActivity')]" position="replace">
+        <!-- Only show this button to admin users. -->
+            <t t-if="chatter.env.session.is_admin">
```
armandcela, line 6: Do we need to have both conditions on the t-if the previous one has_activities

armandcela (approved):

addons/udes_security/models/__init__.py
```diff
@@ -4,4 +4,4 @@
 from . import ir_attachment
 from . import ir_module_category
 from . import domain_allowlist
-from . import res_company  # noqa: F401
+from . import res_company  # noqa: F401
```
PeterAlabaster, line 7: no newline at end of file

addons/udes_security/static/src/xml/custom_chatter_topbar.xml
PeterAlabaster: the target button we replace is already inside the `has_activities` t-if so both will be evaluated

PeterAlabaster (approved):

PR #796, opened 2023-08-29
[IMP] udes_stock_picking_batch: Retrieve information about a tasks result_package_id.

jrbirch99:
- Store extra information about a tasks `result_package_id` for later processing

### DIFF
--- a/addons/udes_stock_picking_batch/models/stock_move_line.py
+++ b/addons/udes_stock_picking_batch/models/stock_move_line.py
@@ -63,6 +63,7 @@ class StockMoveLine(models.Model):
             task["pick_quantity"] = sum(self.mapped("product_qty"))
             task["initial_demand"] = sum(self.move_id.mapped("u_uom_initial_demand"))
             task["quant_ids"] = self.get_quants().get_info(level=2)
+            task["result_package"] = bool(self.result_package_id)
         else:
             package = self.package_id
             package.ensure_one()


addons/udes_stock_picking_batch/models/stock_move_line.py
```diff
@@ -63,6 +63,7 @@ def _prepare_task_info(self):
             task["pick_quantity"] = sum(self.mapped("product_qty"))
             task["initial_demand"] = sum(self.move_id.mapped("u_uom_initial_demand"))
             task["quant_ids"] = self.get_quants().get_info(level=2)
+            task["result_package_id"] = self.result_package_id
```
armandcela, line 66: The data is transferred through jsonrpc, so i think need to care a little about performance. In the other MR-s we are just checking if true or not, so maybe result_package_id can be changed to a boolean. In case that we may need in future the values would be better to return result_package_id.get_info()

PeterAlabaster (approved):

armandcela (approved):

addons/udes_stock_picking_batch/models/stock_move_line.py
jrbirch99: I have changed it to `result_package` in this commit https://github.com/unipartdigital/udes-open/pull/796/commits/1d9f3655b878f6a910220b0c4dc54aab30d8eb0a. I have also changed it for the other MR

adamrpatrick (approved):

PR #799, opened 2023-09-01
[IMP] udes_stock:  Sales order priorities are propagated through to p…

JayshreeMahajan:
udes_sale_stock:  Sales order priorities are propagated through to pickings

SET-128

### DIFF
--- a/addons/udes_sale_stock/models/stock_move.py
+++ b/addons/udes_sale_stock/models/stock_move.py
@@ -54,7 +54,7 @@ class StockMove(models.Model):
         same priority as the sales order that the move is created from
         """
         picking = super(StockMove, self)._search_picking_for_assignation()
-        if len(self.sale_line_id.order_id) == 1:
+        if self.sale_line_id:
             if picking and picking.priority != self.sale_line_id.order_id.priority:
                 return self.env['stock.picking']
         return picking
--- a/addons/udes_sale_stock/models/stock_move.py
+++ b/addons/udes_sale_stock/models/stock_move.py
@@ -50,14 +50,14 @@ class StockMove(models.Model):
 
     def _search_picking_for_assignation(self):
         """
-        When searching for pickings to assign moves too, the picking should have the
+        When searching for picking to assign moves too, the picking should have the
         same priority as the sales order that the move is created from
         """
-        pickings = super(StockMove, self)._search_picking_for_assignation()
-        if self.sale_line_id:
-            priority = self.sale_line_id.order_id.priority
-            pickings = pickings.filtered(lambda p: p.priority == priority)
-        return pickings
+        picking = super(StockMove, self)._search_picking_for_assignation()
+        if len(self.sale_line_id.order_id) == 1:
+            if picking and picking.priority != self.sale_line_id.order_id.priority:
+                return self.env['stock.picking']
+        return picking
 
     def _get_new_picking_values(self):
         """
@@ -65,7 +65,7 @@ class StockMove(models.Model):
         as sales order from which it is created.
         """
         values = super(StockMove, self)._get_new_picking_values()
-        if self.sale_line_id:
+        if len(self.sale_line_id.order_id) == 1:
             # Update the values dictionary with the priority of the associated sale order
             values.update({"priority": self.sale_line_id.order_id.priority})
         return values
--- a/addons/udes_sale_stock/models/stock_move.py
+++ b/addons/udes_sale_stock/models/stock_move.py
@@ -8,7 +8,6 @@ _logger = logging.getLogger(__name__)
 
 
 class StockMove(models.Model):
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -530,3 +530,14 @@ def _action_assign(self):
...
+    def _search_picking_for_assignation(self):
+        pickings = super(StockMove, self)._search_picking_for_assignation()
+        priority = self.sale_line_id.order_id.priority
+        return pickings.filtered(lambda p: p.prirotiy == priority)
+
+
+    def _get_new_picking_values(self):
+        values = super(StockMove, self)._get_new_picking_values()
+        values.update({'priority': self.sale_line_id.order_id.priority})
+        return values
```
PeterAlabaster, line 543: nit: no newline at end of file

addons/udes_stock/models/stock_move.py
```diff
@@ -530,3 +530,14 @@ def _action_assign(self):
         """
         super()._action_assign()
         return self
+
+    def _search_picking_for_assignation(self):
+        pickings = super(StockMove, self)._search_picking_for_assignation()
+        priority = self.sale_line_id.order_id.priority
+        return pickings.filtered(lambda p: p.prirotiy == priority)
```
PeterAlabaster, line 537: typo:
```suggestion
        return pickings.filtered(lambda p: p.priority == priority)
```

addons/udes_stock/models/stock_move.py
```diff
@@ -530,3 +530,14 @@ def _action_assign(self):
         """
         super()._action_assign()
         return self
+
+    def _search_picking_for_assignation(self):
```
jrbirch99, line 534: Can you add a comment explaining why we need this change please :slightly_smiling_face: and for `_get_new_picking_values`

PeterAlabaster (approved):

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -48,3 +48,24 @@ def _prepare_procurement_values(self):
...
         values["sale_line_id"] = self.sale_line_id.id
         return values
+    
+    def _search_picking_for_assignation(self):
+        """
+        Override the method to filter and retrieve pickings based on the priority
+        of the associated sale order.
+        """
+        pickings = super(StockMove, self)._search_picking_for_assignation()
+        priority = self.sale_line_id.order_id.priority
```
jrbirch99, line 58: Sorry just noticed something, please can this be:
```suggestion
        if self.sale_line_id:
               priority = self.sale_line_id.order_id.priority
               pickings.filtered(lambda p: p.priority = priority)
        return pickings
```
As this might get called when the picking is not created from a sales order.

Likewise for `_get_new_picking_values()`:
```suggestion
        if self.sale_line_id:
               values.update({'priority': self.sale_line_id.order_id.priority})
        return values
```

addons/udes_stock/models/stock_move.py
jrbirch99: Sorry the comment does not explain why we really need the changes, 
for `_search_picking_for_assignation` - `When searching for pickings to assign moves too, the picking should have the same priority as the sales order. `
for `_get_new_picking_values` - `When creating a picking from a move, the picking should have the same priority as sales order from which it is created.`

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -48,3 +47,25 @@ def _prepare_procurement_values(self):
...
+
+    def _get_new_picking_values(self):
+        """
+        When creating a picking from a move, the picking should have the same priority
+        as sales order from which it is created.
+        """
+        values = super(StockMove, self)._get_new_picking_values()
+        if self.sale_line_id:
+            # Update the values dictionary with the priority of the associated sale order
+            values.update({"priority": self.sale_line_id.order_id.priority})
```
armandcela, line 70: From the code it looks like that self can be multi recordset which eventually will raise error when trying to access if moves we are creating pickings are from different sale orders. Dont know if is a real scenario, saying just looking at the code

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -48,3 +47,25 @@ def _prepare_procurement_values(self):
...
+            pickings = pickings.filtered(lambda p: p.priority == priority)
+        return pickings
+
+    def _get_new_picking_values(self):
+        """
+        When creating a picking from a move, the picking should have the same priority
+        as sales order from which it is created.
+        """
+        values = super(StockMove, self)._get_new_picking_values()
+        if self.sale_line_id:
```
armandcela, line 68:
```suggestion
        if len(self.sale_line_id.order_id) == 1:
```

PR #800, opened 2023-09-11
[IMP] udes_sale_stock: allow reducing quantity of sale orders in editable state

rajeeshknambiar:
SET-143

Allow editing sale orders via EDI, if order is in editable state. Order can be edited if it is not in progress, done or cancelled. Order is considered 'in progress' if linked picking is done or it is assigned and attached to a batch.

Stock move quantities are adjusted by Odoo if quantity is higher. If quantity was lower in EDI file, manually set quantities, if moves are still editable.

### DIFF
--- a/addons/udes_sale_stock/models/sale_order_line.py
+++ b/addons/udes_sale_stock/models/sale_order_line.py
@@ -3,6 +3,7 @@
 from datetime import datetime
 
 from odoo import api, models, fields, _
+from odoo.tools import float_compare
 from odoo.exceptions import ValidationError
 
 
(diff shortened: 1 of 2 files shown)


armandcela:
I think these changes should be on udes_interfaces module. Udes sale stock can be installed without EDI , and doesnt look right to be in a module where EDI is not installed.

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -134,3 +135,35 @@ def _set_ui_is_cancelled(self):
             )
 
         lines_to_cancel.action_cancel()
+
+    def write(self, values):
+        """ If quantity is reduced, stock moves are not updated for them (but warning
+        issued). Update linked stock moves if those are still editable.
+        """
+        lines_qty_reduced  = self.env["sale.order.line"]
```
armandcela, line 143:
```suggestion
        lines_qty_reduced  = self.browse()
```

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -134,3 +135,35 @@ def _set_ui_is_cancelled(self):
...
+        moves_editable = self.env["stock.move"]
```
armandcela, line 144: Definition of class at very beginning, due to our standards of defining classes

```suggestion
        StockMove = self.env["stock.move"]
        moves_editable = Moves.browse()
```

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -134,3 +135,35 @@ def _set_ui_is_cancelled(self):
...
+            precision = self.env['decimal.precision'].precision_get('Product Unit of Measure')
+            moves_editable = self.move_ids.filtered(
+                lambda m: m.state not in ("draft", "done", "cancel")
+            ).sorted()
+            # Lines with quantity reduced and linked moves still editable
+            lines_qty_reduced = self.filtered(
+                lambda r: r.state == 'sale' and not r.is_expense and 
+                    float_compare(r.product_uom_qty, values['product_uom_qty'], precision_digits=precision) == 1 and
+                    (not r.move_ids or r.move_ids.sorted() == moves_editable)
+                    #(not r.move_ids or all(m.state not in ("draft", "done", "cancel") #TODO: 'any'?
```
armandcela, line 156: Is the comment needed?

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -134,3 +135,35 @@ def _set_ui_is_cancelled(self):
...
+        lines_qty_reduced  = self.env["sale.order.line"]
+        moves_editable = self.env["stock.move"]
+
+        if 'product_uom_qty' in values:
+            precision = self.env['decimal.precision'].precision_get('Product Unit of Measure')
+            moves_editable = self.move_ids.filtered(
+                lambda m: m.state not in ("draft", "done", "cancel")
+            ).sorted()
+            # Lines with quantity reduced and linked moves still editable
+            lines_qty_reduced = self.filtered(
```
armandcela, line 152: It sounds confusing to me,
I would suggest to have a method which returns lambda with a doc string.
```suggestion
            lines_qty_reduced = self.filtered(lines_qty_reduced_lambda)
```

rajeeshknambiar:
> I think these changes should be on udes_interfaces module. Udes sale stock can be installed without EDI , and doesnt look right to be in a module where EDI is not installed.

Thanks. Agree @adamrpatrick?

PR #801, opened 2023-09-11
Fixing Odd Desktop Backorder Behaviour

jrbirch99:
- If a picking has `u_validate_real_time = True` with 3 move lines, when a move line is updated with qty_done and saved via the desktop, Odoo will send 3 commands to the backend. One updating the move line with qty_done, and another two re-writing the other 2 move lines to the picking. This meant that even when the qty_done was written a backorder would be created (which the 2 other move lines would be written to) they would then be written to the original picking due to the extra commands sent to the backend. The fix was to filter out these commands in this specific situation.
 - Another weird behaviour that happened with goods in, meant we could have extra move lines attached to a backorder when the picking was validated through the desktop. These extra move lines were in the `assigned` state, but had no reserved quantity (which does not make any sense). The fix was to check and remove those move lines.

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -17,6 +17,8 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     # Adding fields to be hidden by default from form view.
+    # NB JB: Fields added here will cause test Form class to break. Not sure
+    # on fix, will just have to not use the test Form class.
     DetailedFormViewFields = [
         "u_first_picking_ids",
         "u_prev_picking_ids",
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -5,6 +5,7 @@ from odoo import api, models, fields, _
 from odoo.exceptions import ValidationError, UserError
 from .common import get_next_name
 from odoo.addons.udes_common.models.fields import PreciseDatetime
+from odoo.addons.udes_common.tools import RelFieldOps
 from lxml import etree
 
 from ..utils import UDES_STATISTICS_LOG_FORMAT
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -1074,3 +1091,33 @@ def unlink(self):
...
+        for picking in pickings_to_check:
+            move_lines_on_pick = set(picking.move_line_ids.ids)
+            actions = vals.get("move_line_ids")
+            if actions:
+                new_actions = list()
+                for action in actions:
+                    command = action[0]
+                    move_line_id = action[1]
+                    # 4 is command for appending one record to another recordset
+                    if not (command == 4 and move_line_id in move_lines_on_pick):
```
armandcela, line 1119: Maybe it looks easier to read in this way
```suggestion
                    if command != 4 and move_line_id not in move_lines_on_pick:
```

addons/udes_stock/models/stock_picking.py
```diff
@@ -1074,3 +1091,33 @@ def unlink(self):
...
+        to the picking even when they already exist on the picking. This causes issues
+        when an order has some move lines completed (with real_time flag = True) and some
+        needing completion. In UDES we will assign any move lines needing completion
+        to the backorder picking but core Odoo will then reassign these move lines to the
+        original picking (due to the extra actions in vals).
+        """
+
+        # Only consider removing commands when validating in real time and picking is in assigned state
+        pickings_to_check = self.filtered(
+            lambda p: p.picking_type_id.u_validate_real_time and self.state == "assigned"
```
armandcela, line 1108:
```suggestion
            lambda p: p.picking_type_id.u_validate_real_time and p.state == "assigned"
```

armandcela:
Do we need a unit test when we edit from desktop a picking with picking type validate real time True

jrbirch99:
> Do we need a unit test when we edit from desktop a picking with picking type validate real time True

I could not get a unit test working for this situation. I attempted to use the Form test class as suggested on the story, but I kept bumping into errors due to `DetailedFormViewFields`.

addons/udes_stock/models/stock_picking.py
jrbirch99: changed https://github.com/unipartdigital/udes-open/pull/801/commits/042397f7cd7a41306c6e6373258a44d16272d0c7

addons/udes_stock/models/stock_picking.py
jrbirch99: changed https://github.com/unipartdigital/udes-open/pull/801/commits/f31ddbdd1d8c31e02ac6481c68c1ed1c65d9c5be

addons/udes_stock/models/stock_picking.py
jrbirch99: Sorry just checked and this is wrong, because `not (command == 4 and move_line_id in move_lines_on_pick) ` is not the same logic as `command != 4 and move_line_id not in move_lines_on_pick`

jrbirch99: I am going to drop this commit

addons/udes_stock/models/stock_picking.py
```diff
@@ -1074,3 +1091,33 @@ def unlink(self):
...
+        )
+        for picking in pickings_to_check:
+            move_lines_on_pick = set(picking.move_line_ids.ids)
+            actions = vals.get("move_line_ids")
+            if actions:
+                new_actions = list()
+                for action in actions:
+                    command = action[0]
+                    move_line_id = action[1]
+                    # 4 is command for appending one record to another recordset
```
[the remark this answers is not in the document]
jrbirch99: changed https://github.com/unipartdigital/udes-open/pull/801/commits/72f460b68198809c0c71c69d3ff905a799f777e1

addons/udes_stock/models/stock_picking.py
jrbirch99: image

kdwyer (approved):

PR #807, opened 2023-09-19
[IMP] udes_stock_refactoring: Merge possible moves & lines after refactoring

rajeeshknambiar:
SET-297

Merge possible stock moves and move lines after refactoring. Do this at
the end of `_action_assign()` as merging earlier may cause currently
processing move to get merged, cancelled and unlinked in between; resulting
in error.


addons/udes_stock/models/stock_move.py
```diff
@@ -37,7 +37,8 @@ def _action_cancel(self):
         - Include a hook to propagate cancellations when the picking
           type is configured with u_propagate_cancel set to True
         """
-        self.move_line_ids.write({"qty_done": 0})
+        if self.move_line_ids:  # Merging moves may assign move lines to another move and cancel this, skip those
```
rajeeshknambiar, line 40: This change was required to avoid an infinite loop; and it also avoids pointless chain of calls.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -155,7 +155,13 @@ def _action_assign(self):
         res = super(StockMove, self)._action_assign()
 
         refactored_moves = res._action_refactor(stage="assign")
-        res |= refactored_moves
+
+        # Merge possible moves (if same group/location/destination) and their move lines
+        # Note that _merge_moves() may return same or different move(s) after merging other moves (or no merging)
+        refactored_moves = refactored_moves._merge_moves()
```
armandcela, line 161: Can we say that merging moves is a process which is related with refactoring rather than assigning? 
My suggestion in that case maybe to move the _merge_moves and _merge_move_lines inside _action_refactor.

rajeeshknambiar: Makes sense. I have moved the merging into `_action_refactor` and tested to be working well.

One question: should the moves be merged only if `new_moves != stage_moves`? Can it happen that even if refactoring didn't change/create moves, there are other possible moves that could be merged?

CC: @ws-mps

addons/udes_stock_picking_batch/models/stock_picking.py
```diff
@@ -56,7 +56,8 @@ def action_assign(self):
         but the constraint on state has not proven to work correctly.
         """
         res = super().action_assign()
-        self.batch_id._compute_state()
+        if self.exists():   # This picking can get deleted after merge
```
rajeeshknambiar, line 59: Found in unit tests that this picking may get unlinked after merging take place in `action_assign()`.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -119,6 +119,15 @@ def _action_refactor(self, stage=None, refactor_action=False):
                     )
 
                     new_moves = refactor_class.do_refactor(stage_moves)
+
+                    # Merge possible moves (same group/location/destination/product...) and their
+                    # move lines. Note that _merge_moves() may return same or different move(s) after
+                    # merging other moves (no merging can happen too)
+                    # TODO: merge only if new_moves != stage_moves? It's possible there are other
```
rajeeshknambiar, line 126: Question: should we merge only if refactoring changed the stock moves?

addons/udes_stock_refactoring/models/stock_move.py
ws-mps: Sounds reasonable to move it. Moves do not necessarily will change, they might get move into another picking only. I Wouldn't expect to have other moves that could be merged, since everything should get merged when it was refactored in the first place.

rajeeshknambiar: Ok, IIUC, even if refactor doesn't change the set of moves; those moves could be potentially merged; so we proceed to merge (as is).

PeterAlabaster (approved):

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -103,6 +103,9 @@ def _action_refactor(self, stage=None, refactor_action=False):
         refactor_moves = moves.filtered(refactor_lam)
 
         for picking_type, picking_type_moves in refactor_moves.groupby("picking_type_id"):
+            # TECH-DEBT: temporary fix for issue #5172 <https://taiga.unipart.digital/project/udes-x/issue/5172>
+            picking_type_moves = picking_type_moves.filtered(lambda m: m.state != "cancel")
```
ws-mps, line 107: You can do this outside.

ws-mps: As per discussion this might be sorted by changing the order of the `refactor_lam`.

addons/udes_stock/tests/test_stock_move.py
```diff
@@ -605,6 +605,8 @@ def test_cancel_picking_propagates_to_next_pickings(self):
         for next_picking in next_pickings:
             with self.subTest():
                 self.assertEqual(next_picking.move_lines.product_qty, 6)
+        self.pick.move_lines.write({"propagate_cancel": True})
```
PeterAlabaster, line 608:
```suggestion
        (self.pick.move_lines | next_pickings.move_lines).write({"propagate_cancel": True})
```

PR #809, opened 2023-09-25
Set 160

armandcela:


kdwyer (approved):

addons/udes_stock/models/stock_move.py
```diff
@@ -449,17 +457,6 @@ def update_orig_ids(self, origin_moves):
                 updated_origin_moves |= previous_mls.move_id
             if quantity_todo > 0:
                 unlinked_qty_by_move[move] = quantity_todo
-            if quantity_todo < 0:
```
adamrpatrick, line 452: I assume that by removing this, the condition can no longer be met?

armandcela: This can be met for putaway, thats why was raising warning. But its not met in outbound where initially this method was created for.
We have added that when made the changes mostly for outbound, in case that will be met in any scenario In outbound we havent met that scenario but met in putaway when under putaway which is what we expect because the goods in which is origin of putaway has done more qty. Thats why in putaway is hitting every time when we under putaway.

PeterAlabaster (approved):

addons/udes_stock/models/stock_move.py
adamrpatrick: So you're comfortable that we can safely remove this and we don't benefit from the logging that it provides us at outbound or putaway?

armandcela: We can safely remove this, added at first to hit a possible not forecasted case and to investigate when that happened.
Now we know that is hitting always when we are under putaway, but that is expected.

adamrpatrick (approved):

PR #815, opened 2023-10-12
[META] udes_warehouse_classification: Fix warning

PeterAlabaster:
Add _description to wizard to prevent warning

### DIFF
--- a/addons/udes_warehouse_classification/wizard/add_product_classification.py
+++ b/addons/udes_warehouse_classification/wizard/add_product_classification.py
@@ -11,6 +11,7 @@ class AddProductClassification(models.TransientModel):
     """Add product classification action wizard."""
 
     _name = "udes_warehouse_classification.add_product_classification"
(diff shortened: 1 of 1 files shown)


addons/udes_warehouse_classification/wizard/add_product_classification.py
```diff
@@ -11,6 +11,7 @@ class AddProductClassification(models.TransientModel):
     """Add product classification action wizard."""
 
     _name = "udes_warehouse_classification.add_product_classification"
+    _description = "Warehouse Classification - Add product classification wizard"
```
armandcela, line 14: Nit: Maybe name should be shorter as name of model can be display name of many2one field of model.model i think.

PeterAlabaster: It won't be shown as m2o, it's a wizard. Ignoring :smile:

PR #817, opened 2023-10-25
[IMP] udes_stock: Set AUDIT_LOG_PRINTING to True on stock.picking

PeterAlabaster:
SE-147

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -16,6 +16,10 @@ _logger = logging.getLogger(__name__)
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    # Set this on a model to have automated printing and desktop printing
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -16,6 +16,10 @@
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    # Set this on a model to have automated printing and desktop printing
+    # trigger audit log entries on the record.
+    AUDIT_LOG_PRINTING=True
```
PeterAlabaster, line 21: QQ: Do we want this here or in another module?

rajeeshknambiar (approved):

addons/udes_stock/models/stock_picking.py
PeterAlabaster: Moving to another module

PR #820, opened 2023-11-22
[ADD] udes_stock: Write destination location before qty done when picking the items

armandcela:
SE-527

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -262,12 +262,9 @@ class StockMoveLine(models.Model):
                     )
                 else:
                     ml_values["qty_done"] = ml.product_qty
+                if loc_dest_instance is not None:
+                    ml_values["location_dest_id"] = loc_dest_instance.id
                 mls_done |= ml._mark_as_done(ml_values)
-
-            if loc_dest_instance is not None:
-                # HERE(ale): updating the dest loc here to have a single
-                # invocation of its constraint handler (see below)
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -652,6 +651,8 @@ def _mark_as_done(self, values, split=True):
             )
 
         if split:
+            if values.get("location_dest_id"):
+                self.location_dest_id = values["location_dest_id"]
```
rajeeshknambiar, line 655: Isn't this trying to set `location_dest_id` object with integer id value?

rajeeshknambiar (approved):
LGTM

addons/udes_stock/models/stock_move_line.py
armandcela: Yes when you write a many2one field it handles both integer or recordset

addons/udes_stock/models/stock_move_line.py
```diff
@@ -262,13 +262,12 @@ def mark_as_done(
                     )
                 else:
                     ml_values["qty_done"] = ml.product_qty
+                if loc_dest_instance is not None:
+                    # HERE(ale): updating the dest loc here to have a single
+                    # invocation of its constraint handler (see below)
```
adamrpatrick, line 267: Comment no longer needed

adamrpatrick (approved):

addons/udes_stock/models/stock_move_line.py
armandcela: Pushed the suggested change

PR #821, opened 2023-12-11
[IMP] udes_stock: Change the tracking of a product if it has no stock

JayshreeMahajan:
SE-85

### DIFF
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -85,6 +85,4 @@ class ProductProduct(models.Model):
             ]
         )
 
-        if self.get_quant_counts() > 0 or bool(stock_pickings):
-            return True
-        return False
+        return bool(stock_pickings) or self.get_quant_counts() > 0
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -85,4 +85,4 @@ class ProductProduct(models.Model):
             ]
         )
 
-        return bool(stock_pickings) or self.get_quant_counts() > 0
+        return bool(stock_pickings) or self.get_quant_counts()
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -63,3 +63,28 @@ class ProductProduct(models.Model):
         """Override superclass to prevent deletion."""
         raise ValidationError(_("Products may not be deleted. Please archive them instead."))
 
+    def get_quant_counts(self):
+        warehouse = self.env.ref("stock.warehouse0")
+        Quant = self.env["stock.quant"]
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/product_product.py
```diff
@@ -63,3 +63,28 @@ def unlink(self):
...
+        picking_type = self.env.ref("stock.picking_type_in")
+        stock_pickings = Picking.search(
+            [
+                ("move_lines.product_id", "in", self.ids),
+                ("state", "=", "assigned"),
+                ("picking_type_id", "=", picking_type.id),
+            ]
+        )
+
+        if self.get_quant_counts() > 0 or bool(stock_pickings):
```
ws-mps, line 88: I would swap the conditions so the quant search is not triggered if there are pickings. Also we can directly return this condition.

ws-mps (approved):

addons/udes_stock/models/product_product.py
```diff
@@ -63,3 +63,26 @@ def unlink(self):
...
+        picking_type = self.env.ref("stock.picking_type_in")
+        stock_pickings = Picking.search(
+            [
+                ("move_lines.product_id", "in", self.ids),
+                ("state", "=", "assigned"),
+                ("picking_type_id", "=", picking_type.id),
+            ]
+        )
+
+        return bool(stock_pickings) or self.get_quant_counts() > 0
```
armandcela, line 88: Nit:
```suggestion
        return bool(stock_pickings) or self.get_quant_counts()
```

armandcela (approved):

addons/udes_stock/models/product_template.py
```diff
@@ -49,3 +49,24 @@ def _domain_height_category(self):
     def unlink(self):
         """Override superclass to prevent deletion."""
         raise ValidationError(_("Products may not be deleted. Please archive them instead."))
+
+    @api.onchange("tracking")
+    def onchange_tracking(self):
```
PeterAlabaster, line 54: this and `constrain_tracking()` do the same thing, we can abstract the contents to a helper function

PeterAlabaster:
```python
    @api.onchange("tracking")
    def onchange_tracking(self):
        self.check_tracking()

    @api.constrains("tracking")
    def constrain_tracking(self):
        self.check_tracking()

    def check_tracking(self):
        for product in self:
            if product.product_variant_ids.has_goods_in_transit_or_stock():
                # If there is stock, raise an error to prevent changing the tracking
                raise ValidationError(
                    _(
                        "Cannot change tracking for product '%s' with stock or move lines in ready state."
                    )
                    % product.name
                )
```

PeterAlabaster (approved):

PR #822, opened 2023-12-21
[IMP]udes_suggest_location : set limit to none when calling suggest_l…

JayshreeMahajan:
…ocation from validate_location_dest()

SE-511

### DIFF
--- a/addons/udes_suggest_location/models/stock_move_line.py
+++ b/addons/udes_suggest_location/models/stock_move_line.py
@@ -115,7 +115,7 @@ class StockMoveLine(models.Model):
                     # Allow view destination locations on create
                     continue
                 # Get the suggested locations for comparison
-                suggested_locations = mls_validation_grp.suggest_locations()
+                suggested_locations = mls_validation_grp.suggest_locations(limit=None)
                 if not suggested_locations:
                     raise ValidationError(_("There are no valid locations to drop stock"))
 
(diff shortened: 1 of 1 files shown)


ws-mps (approved):

addons/udes_suggest_location/models/stock_move_line.py
```diff
@@ -115,7 +115,7 @@ def validate_location_dest(self, locations=None):
                     # Allow view destination locations on create
                     continue
                 # Get the suggested locations for comparison
-                suggested_locations = mls_validation_grp.suggest_locations()
+                suggested_locations = mls_validation_grp.suggest_locations(limit=None)
```
PeterAlabaster, line 118: @ws-mps if we remove limits on locations, say there are 1,000 locations suggested - do we really want to show them on the mobile UI? Or do we want to show a subset (say 30) but allow the user to scan if it's one of the 1,000?

ws-mps: This is the function that validates once it gets updated, so it should be fine. The limit should be applied somewhere else.

PeterAlabaster: OK Cool, just wanted to check it hadn't been missed :smile:

PeterAlabaster (approved):

PR #826, opened 2024-02-14
SE-669: float round issue

rajeeshknambiar:


addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -2088,3 +2088,40 @@ def test_passes_partner_id_to_moves_without_package_ids_at_creation_time(self):
...
+        Scenario: Goods in for 1.2 m. Product received in 12 batches of 0.1 m. Rounding error could
+        change 0.1 to 0.11 and split to extra move line with qty 0.01 and block validating the pick
+        """
+        product_info = [{"product": self.cable, "uom_qty": 1.2}]
+        picking = self.create_picking(
+            self.picking_type_goods_in, products_info=product_info, assign=True
+        )
+
+        product_ids = [{"barcode": "productCable", "uom_qty": 1.2,
+                        "lot_names": [f"l:02" for l in range(12)],      #["01", "02",..., "12"]
```
kdwyer, line 2113: We could just do this? Similarly for `lot_quantities` and in the asserts.
```python
['1.02] * 12
```

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -2088,3 +2088,40 @@ def test_passes_partner_id_to_moves_without_package_ids_at_creation_time(self):
...
+        pallet_to_pick_onto = self.create_package(name="UDES1")
+
+        res = picking.move_line_ids.prepare(
+            product_ids=product_ids, location_dest=location, result_package=pallet_to_pick_onto
+        )
+        for mls, mls_values in res.items():
+            mls.mark_as_done(mls_values)
+        picking.button_validate()   #Ensure all quantity received correctly
+        self.assertEqual(len(picking.move_line_ids), 12)
+        self.assertEqual(sorted(picking.move_line_ids.mapped("lot_name")), [f"l:02" for l in range(12)])
```
kdwyer, line 2125: Again, are we missing `{}`?

addons/udes_stock/tests/test_stock_picking.py
PeterAlabaster: Currently this would be:
```
>>> f"l:02" for l in range(12)]
['l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02', 'l:02']
```
I think it's just missing {}, and needs slicing up 1:
```
>>> [f"{l:02}" for l in range(1, 13)]
['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12']
```
although `1, 13` might not be necessary if we don't care if a lot is named `"00"`, but the comment suggests otherwise

PeterAlabaster: Maybe also make it a variable, and use that in the assert too

kdwyer: Also maybe don't use `l` as the loop variable so my eyes can read it properly :-)

rajeeshknambiar: > We could just do this? Similarly for `lot_quantities` and in the asserts.
> 
> ```python
> ['1.02] * 12
> ```

I had a faint memory of repeating list element so...

> Or are we missing the `{}` in the f-strings?

Thanks on both counts.

kdwyer: @PeterAlabaster I'm happy that this has been addressed - please resolve it if you are too.

kdwyer (approved):

armandcela (approved):

PeterAlabaster (approved):

PR #831, opened 2024-03-22
[ADD] udes_stock: Abstract individual move value prep to function

PeterAlabaster:
This allows other modules to extend the function to modify an individual
moves values based on product_info.

SE-928

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -627,6 +627,38 @@ class StockPicking(models.Model):
         PickingBatch = self.env["stock.picking.batch"]
         PickingBatch.create({"picking_ids": [(6, 0, pickings.ids)]})
 
+    def _prepare_individual_move_values(self, product_info, **kwargs):
+        """
+        Return values for an individual move. This can be extended to provide
+        additional mappings from product_info to move values when calling create_picking()
+        :args:
+            - product_info: dict of move data.
+        :returns:
+            Move values dict
+        """
+        self.ensure_one()
+        product = product_info.get("product")
+        uom_qty = product_info.get("uom_qty") or product_info.get("qty")
+        uom_id = product_info.get("uom_id") or product.uom_id.id
+        expected_date = product_info.get("expected_date")
+        vals = {
+            "product_id": product.id,
+            "name": product.name,
+            "product_uom": uom_id,
+            "product_uom_qty": uom_qty,
+            "location_id": self.location_id.id,
+            "location_dest_id": self.location_dest_id.id,
+            "picking_id": self.id,
+            "priority": self.priority,
+            "picking_type_id": self.picking_type_id.id,
+            "description_picking": product._get_description(self.picking_type_id),
+        }
+        if expected_date:
+            vals.update({"date": expected_date})
+        vals.update(kwargs)
+        return vals
+
+
     def _prepare_move(self, pickings, products_info, **kwargs):
         """Return a list of the move details to be used later in creation of the move(s).
         The purpose of this is to allow for multiple moves to be created at once.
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -627,14 +635,17 @@ def _create_batch(self, pickings):
         PickingBatch = self.env["stock.picking.batch"]
         PickingBatch.create({"picking_ids": [(6, 0, pickings.ids)]})
 
-    def _prepare_move(self, pickings, products_info, **kwargs):
+    def _prepare_move(self, pickings, products_info, extra_product_info_keys=[], **kwargs):
```
kdwyer, line 638: Is it safe to use a mutable default argument here?

PeterAlabaster: Good point: c33d4d9687c19d86e4e8f9d4c4e784fe6a03cb01

addons/udes_stock/models/stock_picking.py
```diff
@@ -627,25 +635,37 @@ def _create_batch(self, pickings):
         PickingBatch = self.env["stock.picking.batch"]
         PickingBatch.create({"picking_ids": [(6, 0, pickings.ids)]})
 
-    def _prepare_move(self, pickings, products_info, **kwargs):
+    def _prepare_move(self, pickings, products_info, extra_product_info_keys=None, **kwargs):
```
armandcela, line 638: Can we handle it with super and set a context in case we want to set additional fields on the move when calling create_picking?

The change do not seem very generic to add in udes open, and increases complexity to my opinion

PeterAlabaster: the change would be mostly the same with context as the change is in the inner loop.

We lose information by calling super and modifying the result externally (i tried this initially, but the relation of products_info and pickings and how they end up in the result of this function is confusing)

PeterAlabaster: I don't understand what is not generic about this change

addons/udes_stock/models/stock_picking.py
```diff
@@ -627,25 +635,37 @@ def _create_batch(self, pickings):
...
                 uom_id = product_info.get("uom_id") or product.uom_id.id
                 expected_date = product_info.get("expected_date")
+                # Unpack additional keys from product_info for when we need to
+                # set additional fields on the move when calling create_picking()
+                kwargs_by_proxy = {
+                    k: v
+                    for k, v in product_info.items()
+                    if k in extra_product_info_keys
+                }
                 vals = {
```
armandcela, line 669: Maybe better create vals in a method which can be easily overriden smth like:

vals = picking.get_move_vals_from_product_info(product_info)

addons/udes_stock/models/stock_picking.py
armandcela: Suggested below a solution i think

armandcela: It becomes very complex to remember to follow the code 
product_info, extra_product_info_keys , and changing kwargs with product_info[key] if key in extra_product_info_keys. I believe the suggested way to get the move vals into a method would be easier to override with super and of course less code change

addons/udes_stock/models/stock_picking.py
PeterAlabaster: Yeah this can probably work. Didn't think of abstracting the inner part, i'll make the change

addons/udes_stock/models/stock_picking.py
PeterAlabaster: c8075a7402d85b91ecdb40b94a78459c3fb04a9e

addons/udes_stock/models/stock_picking.py
PeterAlabaster: No longer using this method, will resolve

addons/udes_stock/models/stock_picking.py
```diff
@@ -627,6 +627,38 @@ def _create_batch(self, pickings):
         PickingBatch = self.env["stock.picking.batch"]
         PickingBatch.create({"picking_ids": [(6, 0, pickings.ids)]})
 
+    def _prepare_individual_move_values(self, picking, product_info, **kwargs):
```
armandcela, line 630: Nit: Self can be picking i think
```suggestion
    def _prepare_individual_move_values(self, product_info, **kwargs):
```

PeterAlabaster: Oops, yeah it can

PR #833, opened 2024-05-10
[FIX] udes_priorities: Compute priority sequence on picking.

kdwyer:
SE-822

In cases where we want to order by priority, we should use the priority's sequence attribute. Make this available on the picking to enable ordering where required.

### DIFF
--- a/addons/udes_priorities/models/stock_picking.py
+++ b/addons/udes_priorities/models/stock_picking.py
@@ -10,9 +10,12 @@ class StockPicking(models.Model):
         return self.env.ref("udes_priorities.normal").reference
 
     priority = fields.Selection(selection="get_priorities_for_selection", default=_default_priority)
+    u_priority_id = fields.Many2one(comodel_name="udes_priorities.priority")
+    u_priority_sequence = fields.Integer(
+        string="Priority Sequence", compute="_compute_priority_sequence", store=False
+    )
 
     def _priority_and_priority_group_domain(self, picking_type_ids=None):
-
         domain = []
         if picking_type_ids is None:
             picking_type_ids = []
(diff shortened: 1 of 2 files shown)


addons/udes_priorities/models/stock_picking.py
```diff
@@ -10,9 +10,12 @@ def _default_priority(self):
         return self.env.ref("udes_priorities.normal").reference
 
     priority = fields.Selection(selection="get_priorities_for_selection", default=_default_priority)
+    u_priority_id = fields.Many2one(comodel_name="udes_priorities.priority")
+    u_priority_sequence = fields.Integer(
+        string="Priority Sequence", compute="_compute_priority_sequence", store=False
```
rajeeshknambiar, line 15: Would `related=u_priority_id.sequence` work instead of compute?

addons/udes_priorities/models/stock_picking.py
```diff
@@ -10,9 +10,12 @@ def _default_priority(self):
         return self.env.ref("udes_priorities.normal").reference
 
     priority = fields.Selection(selection="get_priorities_for_selection", default=_default_priority)
+    u_priority_id = fields.Many2one(comodel_name="udes_priorities.priority")
```
armandcela, line 13: Is u_priority_id field used? I think only  u_priority_sequence is used, and can set that field on def create or def write in the same way you are doing for u_priority_id, or dont need to have it computed or related, maybe can set the value of u_priority_sequence in create and write

armandcela: Ignore my suggestion, from discussions had on channel

addons/udes_priorities/models/stock_picking.py
kdwyer: No, because an additional requirement (which I have no implemented) is that the priority sequence should not be updated on cancelled or completed pickings, and `related=` doesn't let us filter by a search domain.

Good question though :+1:

addons/udes_priorities/models/stock_picking.py
```diff
@@ -45,6 +48,14 @@ def _priority_group_domain(self, picking_type_ids=None):
         # To allow for this they are proxies to the default function
         return self._priority_and_priority_group_domain(picking_type_ids=picking_type_ids)
 
+    @api.depends("u_priority_id.sequence")
+    def _compute_priority_sequence(self):
+        """Set the sequence associated with a picking's priority."""
+        for record in self:
```
armandcela, line 54: Nit; maybe can filter out self with state done or cancel, and no need to check inside the loop after

armandcela (approved):

PeterAlabaster (approved):

addons/udes_priorities/models/stock_picking.py
kdwyer: Good point!  Made the change

PR #835, opened 2024-05-24
Se 822 - Fix cache issue

kdwyer:


addons/udes_priorities/fields.py
```diff
@@ -0,0 +1,28 @@
+"""Dodgy field"""
```
PeterAlabaster, line 1: nit: update :laughing:

adamrpatrick (approved):

kdwyer:
We found a different solution for this

PR #839, opened 2024-06-07
[FIX] udes_security: mail thread could be html in binary form, fix comparison

rajeeshknambiar:
SE-1231


PeterAlabaster:
nit commit typo : `udes_secutiry` -> `udes_security`

PeterAlabaster (approved):

PR #840, opened 2024-07-02
[FIX] udes_priorities: Propagate priority sequence on write instead o…

PeterAlabaster:
…f compute.

On large databases with many pickings, Odoo spends a very long time preparing the pickings for compute and this can lead to hour(s) long wait times if moving a priority sequence up or down. This change just propagates to the relevant pickings on write, which means the precomputation is no longer necessary and the speed improvement is orders of magnitude faster.

The only downside, is that if the picking priority changes, then we need to handle updating the sequence there with a write/create hook.

I have removed the `u_priority_id` computed field along with this as it is not needed.

SE-1268


armandcela (approved):

adamrpatrick (approved):

addons/udes_priorities/models/priority.py
```diff
@@ -166,6 +166,20 @@ def _validate_transfer_types_changes(self, values):
...
+        """
+        Propagate sequence changes to all relevant stock transfers with the current priority.
+        Done and Cancelled transfers are excluded for audit traceability.
+        """
+        StockPicking = self.env["stock.picking"]
+        for priority in self:
+            pickings_to_update = StockPicking.search([
+                ("priority", "=", priority.reference),
+                ("state", "not in", ["cancel", "done"])
+            ])
```
kdwyer, line 180: It might be more efficient to use a positive `in` filter, so that for large tables the query uses the index on `stock_picking.state`:

```psql
dbd# EXPLAIN ANALYZE select id from stock_picking where state in ('draft', 'confirmed', 'assigned');
                                                                  QUERY PLAN                                                                   
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 Index Scan using stock_picking_state_index on stock_picking  (cost=0.43..301.23 rows=278 width=4) (actual time=0.816..3.540 rows=472 loops=1)
   Index Cond: ((state)::text = ANY ('{draft,confirmed,assigned}'::text[]))
 Planning Time: 15.644 ms
 Execution Time: 3.591 ms
(4 rows)

Time: 20.208 ms
db# EXPLAIN ANALYZE select id from stock_picking where state not in ('draft', 'confirmed', 'assigned');
                                                       QUERY PLAN                                                        
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 Seq Scan on stock_picking  (cost=0.00..74476.10 rows=1113111 width=4) (actual time=3.216..546.793 rows=1106603 loops=1)
   Filter: ((state)::text <> ALL ('{draft,confirmed,assigned}'::text[]))
   Rows Removed by Filter: 472
 Planning Time: 0.315 ms
 Execution Time: 567.798 ms
```

kdwyer (approved):

PR #846, opened 2024-08-28
[ADD] udes_stock: Setting bypass_state_check to True for Support Users

armandcela:
Setting bypass_state_check to True for Support Users in order that they can unlink pickings when they are in cancel or draft states, if in progress will automatically cancel which will unreserve and after unlink

### DIFF
--- a/addons/udes_stock/tests/test_stock_picking.py
+++ b/addons/udes_stock/tests/test_stock_picking.py
@@ -1962,10 +1962,12 @@ class TestStockPickingPriorities(common.BaseUDES):
         pick.unlink()
         self.assertFalse(pick.exists())
 
-    def test_picking_cannot_be_deleted_outside_draft(self):
+    def test_picking_cannot_be_deleted_outside_draft_for_non_trusted_users(self):
         """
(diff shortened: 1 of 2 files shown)


addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -2005,10 +2007,30 @@ def test_picking_cannot_be_deleted_outside_draft(self):
...
+        pick.action_cancel()
+        self.assertEqual(pick.state, "cancel")
+        pick.unlink()
+        self.assertFalse(pick.exists())
+
     def test_picking_can_be_deleted_with_bypass_state_check(self):
         """
         Test that a picking can be deleted with context variable
         """
+        # Making sure is not a support user
```
PeterAlabaster, line 2032: nit, we interchange `support` and `trusted`. maybe we should just stick to `trusted` as that's the group name

PeterAlabaster (approved):

adamrpatrick (approved):

addons/udes_stock/tests/test_stock_picking.py
armandcela: Pushed the suggested change, thanks

PR #847, opened 2024-09-04
[FIX] udes_stock: Removing next and previous pickings links from move…

armandcela:
…s to be unlinked

SE-1416


addons/udes_stock/models/stock_move.py
```diff
@@ -386,6 +386,20 @@ def _create_split_move(self, move_lines, remaining_qty, new_move_qty, **kwargs):
...
 
+    def _clean_merged(self):
+        """
+        Remove the links of next pickings and previous pickings of the moves that will be
+        cancelled and unlinked after to not propagate cancellation on the next pickings.
+
+        Unlinking the moves of next and previous pickings if method is triggered by refactoring.
+        """
+        super()._clean_merged()
+        if self._context.get("refactoring"):
```
ws-mps, line 397: Considering this is a more generic module, we could rename this context variable to something like: `clean_merged_remove_related_moves`

PeterAlabaster (approved):

addons/udes_stock/models/stock_move.py
armandcela: Pushed the change

ws-mps (approved):

PR #850, opened 2024-09-10
[IMP] udes_stock: Add validation to ensure Delivery Control is comple…

JayshreeMahajan:
…ted before setting picking to draft

- Added a check in `button_draft` to raise an error if the linked delivery control is not completed.

SE-1270

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -465,8 +465,17 @@ class StockPicking(models.Model):
         when the button is clicked.
         Normally should not be any move lines, adding extra safety just in case in an
         unknown scenario there are move lines.
+        Additionally, Ensures that the Delivery Control process is completed before setting the picking to draft. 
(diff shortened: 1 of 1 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -465,8 +465,17 @@ def button_draft(self):
...
+        Additionally, Ensures that the Delivery Control process is completed before setting the picking to draft. 
+        This is triggered if there is a linked delivery control and its state is not "done".
         """
         self.ensure_one()
+
+        if (
+            self.u_delivery_control_picking_id
+            and self.u_delivery_control_picking_id.state != "done"
+        ):
+            raise UserError("Complete the delivery control process first.")
```
rajeeshknambiar, line 477: Better to provide the delivery control name also in the error message, that would be helpful to users.

PeterAlabaster (approved):

rajeeshknambiar (approved):

PR #853, opened 2024-09-24
[ADD] udes_stock: Transfer create procurement group on stock move 

armandcela:
SE-1442

### DIFF
--- a/addons/udes_stock/models/procurement.py
+++ b/addons/udes_stock/models/procurement.py
@@ -18,7 +18,7 @@ class ProcurementGroup(models.Model):
             domain += [('id', 'in', orderpoints)]
         return domain
 
-        
+
 class StockWarehouseOrderPoint(models.Model):
     _inherit = 'stock.warehouse.orderpoint'
 
--- a/addons/udes_stock/models/stock_move.py
+++ b/addons/udes_stock/models/stock_move.py
@@ -545,3 +545,25 @@ class StockMove(models.Model):
         """
         super()._action_assign()
         return self
+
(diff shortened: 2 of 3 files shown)


addons/udes_stock/models/stock_move.py
```diff
@@ -545,3 +545,29 @@ def _action_assign(self):
...
         super()._action_assign()
         return self
+
+    def _action_confirm(self, merge=True, merge_into=False):
+        """
+        Assign procurement group for moves where picking type create procurement groups is enabled
+        """
+        Procurement = self.env["procurement.group"]
+
+        for orderpoint_moves in self.filtered(
```
ws-mps, line 555: We can just loop over the moves that have `mv.picking_type_id.u_create_procurement_group and not mv.group_id` and then have an if to decide the name, then we only need 
```

Comment (armandcela) [reply addons/udes_stock/models/stock_move.py]: Pushed suggested changes

File: addons/udes_stock/models/procurement.py [hunk 2]
```diff
@@ -18,6 +18,17 @@ def _get_orderpoint_domain(self, company_id=False):
             domain += [('id', 'in', orderpoints)]
         return domain
 
+    @api.model
+    def _get_or_create_procurement_group(self, procurement_name):
```
ws-mps, line 22: Can we use `mixin.stock.model` and then we will get get or create from there?

armandcela: Pushed the suggested change

ws-mps (approved):

rajeeshknambiar (approved):

PR #854, opened 2024-10-02
SE-1435: Restrict File Uploads to Prevent Malicious File Uploads

armandcela:


adamrpatrick (approved):

kdwyer (approved):

addons/udes_security/models/ir_attachment.py
```diff
@@ -4,19 +4,22 @@
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError
 from PIL import Image
+import base64
+import mimetypes
+from odoo.tools.mimetypes import guess_mimetype # Uses python-magic to guess mimetype
```
PeterAlabaster, line 9: does this require any additional packages in requirements.txt?

PeterAlabaster (approved):

addons/udes_security/models/ir_attachment.py
armandcela: hmmm not sure but i am not using an extra library, in core method is trying to import with try catch and is a comment that might be 2 different packages. 

```
try:
    import magic
except ImportError:
    magic = None
else:
    # There are 2 python libs named 'magic' with incompatible api.
    # magic from pypi https://pypi.python.org/pypi/python-magic/
    if hasattr(magic, 'from_buffer'):
        _guesser = functools.partial(magic.from_buffer, mime=True)
    # magic from file(1) https://packages.debian.org/squeeze/python-magic
    elif hasattr(magic, 'open'):
        ms = magic.open(magic.MAGIC_MIME_TYPE)
        ms.load()
        _guesser = ms.buffer

    def guess_mimetype(bin_data, default=None):
        mimetype = _guesser(bin_data)
        # upgrade incorrect mimetype to official one, fixed upstream
        # https://github.com/file/file/commit/1a08bb5c235700ba623ffa6f3c95938fe295b262
        if mimetype == 'image/svg':
            return 'image/svg+xml'
        return mimetype
```

PeterAlabaster: OK, just wanted to make sure we didn't need to add any additional dependencies.

PR #858, opened 2024-10-09
[IMP] udes_stock: Refactor message body template to correctly replace…

JayshreeMahajan:
… quantity lines

- Applied dynamic rounding for quantities using `u_decimal_places`.

SE-1487

### DIFF
--- a/addons/udes_stock/views/stock_template.xml
+++ b/addons/udes_stock/views/stock_template.xml
@@ -6,5 +6,23 @@
                 <strong>The quantity has been updated.</strong>
             </xpath>
         </template>
+        <template id="message_body" inherit_id="stock.message_body">
+            <xpath expr="//li[contains(text(), 'Quantity :')]" position="replace">
+                <t t-if="move.state != 'done'">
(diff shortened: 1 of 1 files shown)


addons/udes_stock/views/stock_template.xml
```diff
@@ -6,5 +6,23 @@
...
             </xpath>
         </template>
+        <template id="message_body" inherit_id="stock.message_body">
+            <xpath expr="//li[contains(text(), 'Quantity :')]" position="replace">
+                <t t-if="move.state != 'done'">
+                    <t t-if="'product_uom_qty' in vals">
+                        <li>
+                            Quantity : 
+                            <t t-esc="round(move.product_uom_qty, move.product_uom.u_decimal_places)"/> -&gt; 
+                            <t t-esc="round(vals.get('product_uom_qty'), move.product_uom.u_decimal_places)"/>
```
armandcela, line 16: I think changes should be in udes_uom module because there is added the u_decimal_places field

PR #863, opened 2024-11-13
[IMP] udes_security: Process URL-based images, convert to binary, and…

JayshreeMahajan:
… remove EXIF data

SE-1555

### DIFF
--- a/addons/udes_security/models/ir_attachment.py
+++ b/addons/udes_security/models/ir_attachment.py
@@ -1,6 +1,7 @@
 import html
 import logging
 import io
+import requests
 from odoo import api, fields, models, _
 from odoo.exceptions import UserError
 from PIL import Image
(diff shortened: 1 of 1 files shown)


adamrpatrick (dismissed):

addons/udes_security/models/ir_attachment.py
```diff
@@ -127,6 +136,19 @@ def _remove_exif_data(self, datas):
         image.save(output, format=image.format)
         return base64.b64encode(output.getvalue())
 
+    def _process_url_image(self, url):
+        """Fetch an image from the URL and convert it to binary"""
```
kdwyer, line 140: At this point, is URL guaranteed to be safe (so a UDES URL I suppose)?  Or does it matter?

PeterAlabaster: The tests appear to be happy if we add a check like this:
```python
if not vals["url"].startswith("/"):
```
to avoid checking relative urls.

Still need to look into the checking of type though, as I'm not sure if the _remove_exif_data function will get hit, is "mimetype" going to be set to one of the allowed mimetypes if it's a URL?

kdwyer: I spun up a unit test locally and it seems to go through the mimetype checks.

PeterAlabaster: I'm happy with the latest change

adamrpatrick (approved):

addons/udes_security/models/ir_attachment.py
```diff
@@ -136,6 +165,22 @@ def write(self, vals):
...
             del vals["active"]
+
+        if vals.get("url", False):
+            # Restrict URLs to relative paths (starting with "/") or HTTPS (starting with "https://")
+            # This ensures only safe URL types are allowed, preventing schemes like "file://" or "sftp://"
+            # from bypassing security checks.
+            if not (vals["url"].startswith("/") or vals["url"].startswith("https://")):
+                # Raise an error if URL is not a valid type
+                raise UserError("Invalid URL: Only secure HTTPS links or relative URLs are allowed. Please update the URL format and try again.")
+            if "mimetype" in vals and vals["mimetype"] in allowed_image_mimetypes:
```
PeterAlabaster, line 176: If you create a url attachment and set it to something like `https://google.com` then write it and change it to `https://some_site.com/image.png` then this code doesn't get run. The mimetype is left as `application/octet-stream` on `self` and this code doesn't get run, thus the image is not sanitised.

PeterAlabaster: Addressed in c5b227f9c550cb52c1e88946fbac57a872fe716d

PR #864, opened 2024-11-14
SE-1515

ws-mps:


PeterAlabaster (approved):

addons/udes_stock_routing/models/stock_rule.py
```diff
@@ -36,7 +36,9 @@ def get_path_from_location(self, location):
             ]
         )
         if push_steps:
-            return push_steps.sorted(key=lambda p: p.location_src_id.parent_path.strip("/").split("/")[-1], reverse=True)[0]
+            to_order = push_steps.filtered(lambda p: p.location_src_id.parent_path)
+            ordered_steps =  to_order.sorted(key=lambda p: p.location_src_id.parent_path.strip("/").split("/")[-1], reverse=True)
+            return ordered_steps[0] if ordered_steps else (push_steps - to_order)[0]
```
armandcela, line 41: Little confusing to me, maybe i am wring

If not ordered_steps is there any possibility to_order is True?
to_order maybe needs to be smth plural as might be plural isnt it?

addons/udes_stock_routing/models/stock_rule.py
```diff
@@ -36,7 +36,9 @@ def get_path_from_location(self, location):
             ]
         )
         if push_steps:
-            return push_steps.sorted(key=lambda p: p.location_src_id.parent_path.strip("/").split("/")[-1], reverse=True)[0]
+            to_order = push_steps.filtered(lambda p: p.location_src_id.parent_path)
```
armandcela, line 39: Cant we search where parent_path of the location_src_id has value in the search, rather then filtering after searched

addons/udes_stock_routing/models/stock_rule.py
ws-mps: You are correct there is no need to substract, will amend.

PR #865, opened 2024-11-20
SE-1554 File types may be associated with multiple mimetypes

kdwyer:
This series of commits enables associating multiple mimetypes with an allowed file type.  The commits are in roughly three groups:

- preparatory changes: refactoring, linting etc
- the main change: adding mimetypes and their associations with allowed file types
- refactoring: removing duplication and some code moves


adamrpatrick (approved):

addons/udes_security/models/allowed_file_type.py
```diff
@@ -2,6 +2,27 @@
 from odoo.exceptions import UserError
 
 
+class MimeType(models.Model):
+    """Store mime types related to allowed file types."""
+
+    _name = "udes.mimetypes"
+    _description = "Mimetypes related to allowed file types."
+
+    file_type_ids = fields.Many2many(
```
armandcela, line 11: Is needed this many2many field?

addons/udes_security/models/allowed_file_type.py
```diff
@@ -21,6 +42,9 @@ class AllowedFileType(models.Model):
     )
     description = fields.Char("Description")
     active = fields.Boolean("Active?", default=True)
+    mimetype_ids = fields.Many2many(
```
armandcela, line 45: Many2many fields generally are defined where will be a possibility that for 2 different filetypes will have common mimetypes. Is it a scenario where same mimetype can be in 2 different filetypes, if not i would suggest one2many field not many2many.

kdwyer: In our allowed types we have `tif` and `tiff`, which both have a mime type of `image/tiff` so I think many to many is the best way to handle it. There could be others I'd imagine: for example some systems might use `text/plain` for js and css files.

armandcela: Yeah makes sense thats why asked do we have common mimetypes for different file types?

armandcela (approved):

addons/udes_security/models/allowed_file_type.py
kdwyer: We don't actually use it, so it can be removed (tbh I wasn't sure whether Odoo required the declaration in both models or not).

armandcela: No its not required, i think only in one the one which you are gonna use should be fine.

kdwyer: Removed in 4e274cbdd0c97c8af2754edfbb750d9574e4b087

PR #868, opened 2024-11-28
[FIX] Handle Single row CSVs

PeterAlabaster:
which have a misdetected mimetype of text/plain


PeterAlabaster:
Armand Cela
Should be quite fresh for Kevin as he recently added the mimetypes, but from what i remember when cannot find the mimetype from content it finds as text/plain. So potentially might be a security issue if a malicious file has .csv extension

as it might be that mimetype is computed as text/plain

Kevin Dwyer, 11:20
It might be safer to add it in one of the test base classes, rather than to production data?

Peter Alabaster
Maybe we should create an issue to look into this then

PR #871, opened 2025-01-07
[ADD] udes_stock_routing: Two stage outbound hook

PeterAlabaster:
SE-1701

### DIFF
--- a/addons/udes_stock_routing/models/stock_picking.py
+++ b/addons/udes_stock_routing/models/stock_picking.py
@@ -1,0 +1,159 @@
+from odoo import models, fields
+from collections import defaultdict, namedtuple
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    u_from_two_stage_split = fields.Boolean(
+        "Two stage split has been initiated",
+        copy=False,
+        default=False,
+        help="Used to track if a picking is a result of a two stage split, "
+        "to prevent it from being split multiple times if stock has been unreserved and re-reserved, "
+        "as the split occurs on reservation.",
+    )
+
+    def should_two_stage_initiate(self):
+        """Determines if the picking in self requires two stages."""
+        self.ensure_one()
+        Location = self.env["stock.location"]
+        # When stock gets reserved (and has not been split already)
+        if not self.u_from_two_stage_split and self.state == "assigned":
+            # We trigger the split to occur if any of the stock has been reserved
+            # against any locations requiring two stage. We also check their parents
+            # to avoid needing to configure two stage requirements on every individual location.
+            # Individual parent configurations are retrieved later in _get_two_stage_configuration_move_line_mapping()
+            from_locations = self.move_line_ids.location_id
+            two_stage_locations_incl_parents = Location.search_count(
+                [
+                    ("id", "parent_of", from_locations.ids),
+                    ("u_requires_two_stage_when_stock_reserved", "=", True),
+                ]
+            )
+            return bool(two_stage_locations_incl_parents)
+        return False
+
+    def _get_two_stage_configuration_move_line_mapping(self):
+        """
+        Retrieve a mapping of configuration (intermediate location, intermediate location dest, picking type)
+        and the respective moves for each unique configuration which require two stages.
+        """
+        StockMoveLine = self.env["stock.move.line"]
+        Location = self.env["stock.location"]
+
+        configuration_move_line_mapping = defaultdict(StockMoveLine.browse)
+        for move_line in self.move_line_ids:
+            two_stage_locations = Location.search(
+                [
+                    ("id", "parent_of", move_line.location_id.id),
+                    ("u_requires_two_stage_when_stock_reserved", "=", True),
+                ]
+            )
+            config_location = Location.browse()
+            if not two_stage_locations:
+                # This line isn't reserved for a two stage configured location, it will be left on the original picking
+                # while the move lines with two stage configurations get ripped out into new two stage pickings.
+                continue
+            elif len(two_stage_locations) > 1:
+                # We need to determine the location deepest in the tree (itself, or closest parent) to retreieve the configuration,
+                # as it is possible to configure a child location with its own two stage configuration override.
+                # The easiest way to do this is just get the location with the most amount of parent_path nodes.
+                config_location = two_stage_locations.sorted(
+                    lambda l: -len(l.parent_path.split("/"))
+                )[0]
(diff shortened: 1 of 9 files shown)


addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,177 @@
+from odoo import models, fields
+from collections import defaultdict, namedtuple
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    u_from_two_stage_split = fields.Boolean(
```
PeterAlabaster, line 8: should this live on picking extras?

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,177 @@
...
+            # Confirm the stage 2 pick.
+            stage_2_pick.action_confirm()
+
+        # Re-assign the pick if there are any active moves left.
+        # This pick doesn't need u_from_two_stage_split flagging as it only contained reserved
+        # stock from 'normal' locations, so another two stage split won't happen when it gets re-reserved.
+        if self.move_lines.filtered(lambda ml: ml.state != "cancel"):
+            self.action_assign()
+
+        # TODO: Consider picks to intermediate location may trigger the putaway push on drop rule, we don't want that to happen
```
PeterAlabaster, line 176: Needs looking at tomorrow.

addons/udes_stock_routing/tests/test_two_stage_outbound.py
```diff
@@ -0,0 +1,513 @@
...
+        self.assertEqual(normal_pick_banana_line.location_dest_id, self.check_location)
+        self.assertEqual(normal_pick.u_next_picking_ids, stage_2_pick.u_next_picking_ids)
+
+        # Assert that the picks can be completed in succession
+        self.complete_picking(stage_1_pick)
+        self.assertEqual(stage_2_pick.state, "assigned")
+        self.assertEqual(check_pick.state, "waiting")
+        self.complete_picking(stage_2_pick)
+        self.assertEqual(len(check_pick.move_line_ids), 1)
+        # NOTE palabaster: The check pick is assigned at this point. This seems kind of strange
```
PeterAlabaster, line 278: thoughts?

addons/udes_stock_routing/models/stock_picking.py
PeterAlabaster: push on drop rule is not triggered as it applies to the inbound route. so no additional check is required, but in future we _could_ come across a scenario which requires it. We can deal with it as the need arises.

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,174 @@
...
+                stage_1_move = move.copy(
+                    default={
+                        "location_dest_id": configuration.intermediate_location_id,
+                        "picking_id": stage_1_pick.id,
+                        "picking_type_id": configuration.operation_type_id,
+                        # preserve move_orig_ids, incase this move is not first-in-chain, as the original move will be removed.
+                        "move_orig_ids": [[6, 0, move.move_orig_ids.ids]],
+                    }
+                )
+                _stage_2_move = move.copy(
```
PeterAlabaster, line 147: I will change this to modify the original move with the updated links / locations instead of copying it, to avoid needing to cancel the original moves at the end which results in additional cancellations further on in the pick chain. However, the change introduces more breakages in unit tests so I will fix those when I get back

adamrpatrick (approved):

addons/udes_stock_routing/models/stock_picking.py
adamrpatrick: Happy to leave it here

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,183 @@
...
+            # but it is additional complexity outside the initial scope of this feature.
+            #
+            # if 'partially_available' in self.move_lines.mapped("state"):
+            #     return False
+            #
+            # We trigger the split to occur if any of the stock has been reserved
+            # against any locations requiring two stage. We also check their parents
+            # to avoid needing to configure two stage requirements on every individual location.
+            # Individual parent configurations are retrieved later in _get_two_stage_configuration_move_mapping()
+            from_locations = self.move_line_ids.mapped("location_id")
```
armandcela, line 46:
```suggestion
            from_locations = self.move_line_ids.location_id
```

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,183 @@
...
+            two_stage_locations_incl_parents = Location.search(
```
armandcela, line 47: Maybe search_count as not used the recordset anywhere

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -0,0 +1,183 @@
...
+        configuration_move_mapping = defaultdict(StockMove.browse)
+        for move_line in self.move_line_ids:
+            two_stage_locations = Location.search(
+                [
+                    ("id", "parent_of", move_line.location_id.id),
+                    ("u_requires_two_stage_when_stock_reserved", "=", True),
+                ]
+            )
+            config_location = Location.browse()
+            if len(two_stage_locations) == 0:
```
armandcela, line 73:
```suggestion
            if not two_stage_locations:
```

armandcela (approved):

addons/udes_stock_routing/models/stock_picking.py
```diff
@@ -104,37 +93,64 @@ def _get_two_stage_configuration_move_mapping(self):
...
         original_priority = self.priority
         original_origin = self.origin
-        configuration_move_mapping = self._get_two_stage_configuration_move_mapping()
-        # Unreserve stock on the current picking before proceeding.
-        self.do_unreserve()
-        for configuration, moves in configuration_move_mapping.items():
-            stage_1_pick = Picking.create(
+
+        configuration_move_line_mapping = self._get_two_stage_configuration_move_line_mapping()
+        # # Unreserve stock on the current picking before proceeding.
```
PeterAlabaster, line 126: will remove this and the below line, not needed.

PeterAlabaster: https://github.com/unipartdigital/udes-open/pull/871/commits/840052373340763e332992d572f21826c5eafcd3

PR #876, opened 2025-01-27
SE-1721

PeterAlabaster:
SE-1721


addons/udes_stock_routing/models/procurement_group.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models, api
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def _get_rule_domain(self, location, values):
+        """Extend rule domain search to exclude rules which run on stock assignment"""
+        res = super()._get_rule_domain(location, values)
+        res.append(("u_run_on_assign", "=", False))
```
armandcela, line 10: I think is safer like this
```suggestion
        res = expression.AND([res, [("u_run_on_assign", "=", False)]])
```

PeterAlabaster: bece12538bfc1af897a0c3b02329ab72292123e8

addons/udes_stock/models/stock_picking.py
```diff
@@ -166,11 +167,15 @@ def get_next_picking_name(self, vals, picking_type=None):
 
         Backorders have a naming pattern of adding `-001` to it,
         so they are more visible to users.
+
+        Do not customise name generation when the picking type differs.
         """
         if vals.get("backorder_id"):
```
armandcela, line 173: Question. Is it on the vals picking_type_id of the new picking> If yes maybe smth like this would work i think

```
if vals.get("backorder_id") and vals.get("picking_type_id") == picking_type.id:
    ir_sequence = picking_type.sequence_id
    # Specify sequence as it is picking type specific
    return get_next_name(self, "stock.picking", sequence=ir_sequence)
```

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -105,15 +105,147 @@ def _get_new_picking_values(self):
...
+            "origin": self.picking_id.origin,
+        }
+        # We only need to split the moves to a backorder if the picking type
+        # of the rule differs from the assigned moves picking type. If it
+        # is the same picking type, any post assign behaviour has already run, and
+        # the move lines and moves already exist on the picking of the correct type.
+        if assigned_moves.picking_type_id != rule.picking_type_id:
+            # TODO palabaster: This isn't ideal. Could there be cases where this is wrong?
+            # Use any picking from here (there could be multiple) to use as the basis for the backorder.
+            split_pick = assigned_moves.picking_id[0].split_move_lines_to_backorder(
```
armandcela, line 132: I think moves should be of same picking isnt it?
maybe above smth like 
assigned_moves.picking_id.ensure_one()
```suggestion
            split_pick = assigned_moves.picking_id.split_move_lines_to_backorder(
```

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -105,15 +105,147 @@ def _get_new_picking_values(self):
...
+        StockMoveObj = self.env["stock.move"]
+        Rule = self.env["stock.rule"]
+
+        res = StockMoveObj.browse()
+        applicable_rules = Rule.browse()
+
+        for move in self:
+            move_qty = move.product_uom_qty
+
+            applicable_rules = self.get_split_pick_applicable_rules()
```
armandcela, line 166: I think this can stand outside loop
and after directly 
```
if not applicable_rules:
     # If we didn't check rules and call super there, we need to call it as normal.
     res = super()._action_assign()
```

addons/udes_stock_routing/models/stock_quant.py
```diff
@@ -0,0 +1,49 @@
...
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    def _gather(self, product_id, location_id, **kwargs):
+        """
+        Extend _gather to filter by split pick rules, which were passed in via context.
+        """
+        StockQuant = self.env["stock.quant"]
+
+        split_pick_rule = self.env.context.get("split_pick_rule", None)
```
armandcela, line 14: Same thing in fact, can resolve this just if you like more
```suggestion
        split_pick_rule = self._context.get("split_pick_rule", None)
```

addons/udes_stock_routing/models/stock_move.py
PeterAlabaster: no, refactoring can split moves to multiple pickings on assignment

PeterAlabaster: i have changed code in _action_assign now so assigned_moves won't ever include multiple pickings, as that is the source of some other pains, so this will be removed.

addons/udes_stock/models/stock_picking.py
PeterAlabaster: we have to check the picking type of the backorder, not the picking type from self in vals. it doesn't fix the issue if we do it this way, unfortunately

PeterAlabaster:
pushed a big commit to fix how we handle multiple moves for different products, to ensure they don't get split into separate picks. was a bit of a rework but tests are passing (including the new one) and i have re-smoke tested and that is working the same also.

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -105,15 +106,163 @@ def _get_new_picking_values(self):
...
+                            new_moves.write({"picking_id": picking.id})
+                            moves_to_split_by_rule[rule] |= new_moves
+
+        # Now we have split the lines and built the information on which rule each move is related to
+        # we can split those into backorders. Doing it this way prevents us from having picks with multiple
+        # products (moves which cannot be _merged) backorder to several separate pickings.
+        for rule, assigned_moves in moves_to_split_by_rule.items():
+            res |= assigned_moves._split_pick_for_rule(assigned_moves, rule)
+
+        if not applicable_rules:
```
armandcela, line 197: Nit: Please move this above smth like line 166 after finding applicable_rules, not to run unneccessary code if there arent applicable rules

armandcela (approved):

adamrpatrick (approved):

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -105,15 +106,165 @@ def _get_new_picking_values(self):
...
+        Rule = self.env["stock.rule"]
+
+        res = StockMoveObj.browse()
+        applicable_rules = Rule.browse()
+        # Allows us to build data on which moves have been split for which rules
+        # which we can then split to backorder pickings after looping over all rules.
+        moves_to_split_by_rule = collections.defaultdict(StockMoveObj.browse)
+        applicable_rules = self.get_split_pick_applicable_rules()
+
+        if not applicable_rules:
```
ws-mps, line 167: We have moved this up but we still executing all the code, so we either add an else and indent or we call a function to do all the logic from lines 172 til 200.

armandcela: I didnt notice, but i think should be fine just to return if not applicable_rules?

PeterAlabaster: nope, we need to run the two stage hook afterwards

PeterAlabaster: it is hopefully a little easier to read now anyway.

PR #878, opened 2025-02-04
Se 1750

PeterAlabaster:


addons/udes_stock/data/product_categories.xml
```diff
@@ -41,5 +41,12 @@
       <field name="parent_id" ref="product.product_category_all"/>
     </record>
 
+    <!-- Construct "Pallet Categorisation" category -->
+    <!-- Subcategories of this will enforce pallets to be of the same category (SE-1741) -->
```
PeterAlabaster, line 45: nit: i'm not sure where this functionality will live. perhaps this comment should be removed.

armandcela: Yeah i think can be removed that comment

armandcela (approved):

adamrpatrick (approved):

PR #880, opened 2025-02-05
[ADD] SE-1746 udes_stock: Container Type Model and View

PeterAlabaster:
SE-1746
Tree View

Form View

### DIFF
--- a/addons/udes_stock_packaging/__manifest__.py
+++ b/addons/udes_stock_packaging/__manifest__.py
@@ -6,8 +6,10 @@
     "website": "http://github/unipartdigital/udes-open",
     "category": "UDES",
     "version": "0.1",
-    "depends": ["uom", "stock", "package_hierarchy", "udes_stock"],
+    "depends": ["uom", "stock", "package_hierarchy", "udes_stock", "udes_get_info"],
     "data": [
+        "security/ir.model.access.csv",
+        "views/container_type_views.xml",
         "views/stock_picking_type.xml",
         "views/stock_warehouse.xml",
     ],
--- a/addons/udes_stock_packaging/models/__init__.py
+++ b/addons/udes_stock_packaging/models/__init__.py
@@ -1,5 +1,5 @@
 """Models for the UDES Stock Packaging module."""
-
+from . import container_type
 from . import stock_quant_package  # noqa: F401
 from . import stock_warehouse  # noqa: F401
 from . import stock_move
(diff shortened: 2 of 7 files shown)


addons/udes_stock/models/container_type.py
```diff
@@ -0,0 +1,59 @@
+from odoo import models, fields, api, _
+from odoo.exceptions import ValidationError
+
+
+class ContainerType(models.Model):
+    _name = "container.type"
+    _description = "Container Type"
+    _inherit = "mail.thread"
```
armandcela, line 8: Consider inheriting mixin.stock.model

addons/udes_stock/models/container_type.py
```diff
@@ -0,0 +1,59 @@
...
+            "name_uniq",
+            "unique (name)",
+            "A container type with that name already exists.",
+        ),
+    ]
+
+    @api.constrains("length", "width", "height", "weight")
+    def _check_dimensions_and_weight(self):
+        for record in self:
+            if record.length <= 0 or record.width <= 0 or record.height <= 0 or record.weight < 0:
```
armandcela, line 53: NIT: To many or-s, maybe can be improved a little the if control, i would use if any in self or filter the records from self with values less than 0 maybe

armandcela (approved):

addons/udes_stock/security/ir.model.access.csv
```diff
@@ -2,3 +2,4 @@ id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_stock_location_category_manager,stock.location.category.manager,model_stock_location_category,base.group_system,1,1,1,1
 access_stock_location_category_user,stock.location.category.user,model_stock_location_category,base.group_user,1,0,0,0
 access_stock_mark_as_todo_wizard,stock_mark_as_todo_wizard,model_stock_picking_mark_todo,udes_security.group_trusted_user,1,1,1,1
+access_container_type,container.type,model_container_type,base.group_user,1,1,1,1
```
adamrpatrick, line 5: Check with POs, but I think they might want these to be editable by Warehouse Manager and above.

ws-mps:
Wondering if this should go to udes_stock_packaging instead of udes_stock.

addons/udes_stock/security/ir.model.access.csv
PeterAlabaster: Checked:
> Hi. we will leave it as it is for now. We will do a permissions story later

adamrpatrick (approved):

PR #881, opened 2025-02-06
SE-1743: Extend result package handling

kdwyer:
Signal when an existing package should be discarded using an empty recordset. This will preserve the existing behaviour for false-y values.

### DIFF
--- a/addons/udes_stock_packaging/models/stock_quant_package.py
+++ b/addons/udes_stock_packaging/models/stock_quant_package.py
@@ -217,5 +217,10 @@ class StockQuantPackage(models.Model):
             # when result storage format is products
             if result_package_name:
                 raise ValidationError(_("Invalid parameters for products target storage format."))
+            # We never want to have a package set here. Return an empty
+            # recordset to signal to StockMoveLine.prepare that the package
+            # must be discarded.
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -315,10 +315,11 @@ def prepare(
         Example: [{"barcode": "PROD01", "uom_qty":1}]
 
         """
+        Package = self.env["stock.quant.package"]
         Product = self.env["product.product"]
 
         vals = {}
-        if result_package:
+        if isinstance(result_package, type(Package)):
```
PeterAlabaster, line 322: q: what else can `result_package` be?

kdwyer: `None`

ws-mps: Maybe a small comment to explain why we doing this: be able to remove (set false) that field

ws-mps: Similar to the commit message :)

PeterAlabaster (approved):

adamrpatrick (approved):

PR #883, opened 2025-02-17
SE-1704

adamrpatrick:


addons/udes_security/models/ir_attachment.py
```diff
@@ -60,6 +61,39 @@ def _check_contents(self, vals):
...
+
+        Raises:
+            ValidationError: If the attachment data cannot be parsed as a CSV.
+        """
+        try:
+            data_str = base64.b64decode(attachment_datas).decode('utf-8')
+            csv.Sniffer().sniff(data_str, delimiters=",|")
+            csv.Sniffer().has_header(data_str)
+            reader = csv.reader(data_str.splitlines())
+            for row in reader:
```
kdwyer, line 86: [This](https://stackoverflow.com/a/36763172/5320906) might be slightly faster, if we are going to read the entire file.

addons/udes_security/models/ir_attachment.py
```diff
@@ -60,6 +61,39 @@ def _check_contents(self, vals):
...
+                pass
+            if reader.line_num > 1:
+                # File is most likely a genuine CSV so allow it to be attached
+                _logger.info(f"User {self.env.uid} bypassing CSV upload")
+                return True
+        except csv.Error as e:
+            _logger.error(f"CSV validation failed: {e}")
+        raise ValidationError(_("Invalid CSV."))
```
kdwyer, line 94: Does the user see this error?  If so, is it possible (or practical) to give them any more information about the problem?

addons/udes_security/models/ir_attachment.py
```diff
@@ -60,6 +61,39 @@ def _check_contents(self, vals):
...
+
+        Returns:
+            bool: True if the attachment data is a valid multi-row CSV, False otherwise.
+
+        Raises:
+            ValidationError: If the attachment data cannot be parsed as a CSV.
+        """
+        try:
+            data_str = base64.b64decode(attachment_datas).decode('utf-8')
+            csv.Sniffer().sniff(data_str, delimiters=",|")
```
kdwyer, line 83: Should we be doing something with these sniffles?  For example using the dialect returned by `.sniff()` in the call to `csv.reader`? Something like

```python
sniffer = csv.Sniffer()
dialect = sniffer.sniff(...)
if not sniffer.has_header(...): return <bad outcome>
reader = csv.reader(data.splitlines(), dialect=dialect)
```

PeterAlabaster (approved):

ws-mps (approved):

addons/udes_security/models/ir_attachment.py
adamrpatrick: Updated

addons/udes_security/models/ir_attachment.py
adamrpatrick: We're intending to be vague here so it's not so easy to work around, security-wise. Aside from the fact that this is open source...

addons/udes_security/models/ir_attachment.py
adamrpatrick: Updated.

kdwyer (approved):

PR #885, opened 2025-02-18
[ADD] udes_sale_stock: add new udes.courier model and access control

JayshreeMahajan:
- Created the `udes.courier` model with fields for courier name.
- Added access rights for managing courier records.

SE-1753
Signed-off-by: jayshree <jayshree.mahajan@unipart.com>


armandcela:
Hmm i think we have a udes.courier model in udes_packing module, Please check if we can reuse the same model, or we need another model

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -59,6 +59,9 @@ class SaleOrder(models.Model):
         readonly=True, related="warehouse_id.u_allow_manual_sale_order_line_cancellation",
     )
 
+    courier_type = fields.Many2one("udes.courier", string="Courier")
```
armandcela, line 62: Many2one fields should be with ID , and if existing model should be called with prefix u_
u_courier_id example

JayshreeMahajan: addressed

addons/udes_sale_stock/models/courier.py
```diff
@@ -0,0 +1,7 @@
+from odoo import models, fields
+
+class UdesCourier(models.Model):
+    _name = "udes.courier"
+    _description = "Courier"
```
adamrpatrick, line 5: Let's keep this separate to our existing implementations of `udes.courier` and call this `udes.carrier`

addons/udes_sale_stock/models/sale_order.py
adamrpatrick: Rename to `u_carrier_id`

addons/udes_sale_stock/models/courier.py
```diff
@@ -0,0 +1,7 @@
...
+
+    name = fields.Char(string="Courier Name")
```
PeterAlabaster, line 7: Q: Do we want a view or menuitem to manage these records?

addons/udes_sale_stock/models/courier.py
adamrpatrick: So change all instances of "courier" to "carrier" please

addons/udes_sale_stock/models/__init__.py
```diff
@@ -8,3 +8,4 @@
 from . import stock_warehouse
 from . import stock_picking
 from . import stock_picking_type
+from . import courier
```
PeterAlabaster, line 11: nit: rename this and file to `udes_courier` and `udes_courier.py`

adamrpatrick: Good point, `udes.carrier` and `udes_carrier.py` please

PeterAlabaster: (after seeing adams comment, `udes_carrier` and `udes_carrier.py` :smile:

armandcela:
And as we are changing the model name please change the commit message too to match with new model name

addons/udes_sale_stock/models/courier.py
armandcela: Good question

addons/udes_sale_stock/models/sale_order.py
```diff
@@ -59,6 +59,9 @@ class SaleOrder(models.Model):
         readonly=True, related="warehouse_id.u_allow_manual_sale_order_line_cancellation",
     )
 
+    u_carrier_id = fields.Many2one("udes.carrier", string="Carrier")
```
armandcela, line 62: Do we need to show this field on sale order record?

addons/udes_sale_stock/models/courier.py
PeterAlabaster: Resolving, menuitems and views will be done later

adamrpatrick (approved):

armandcela (approved):

PR #886, opened 2025-02-18
Se 1752 2 Refactor by weight

PeterAlabaster:
Original MR: https://github.com/unipartdigital/udes-open/pull/884
Reworked to be based on moves instead of move lines, and support post confirm also.

### DIFF
--- a/addons/udes_stock_refactoring/models/stock_move.py
+++ b/addons/udes_stock_refactoring/models/stock_move.py
@@ -1,5 +1,6 @@
-from odoo import api, models, fields, _
+from odoo import api, models, fields, _, tools
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools.float_utils import float_compare, float_is_zero
 from ..registry.refactor import REFACTOR_REGISTRY
 import logging
 
@@ -208,7 +209,6 @@ class StockMove(models.Model):
         pickings = self.picking_id
 
         for key, move_group in groups:
-
             if len(move_group.location_id) > 1 or len(move_group.location_dest_id) > 1:
                 raise UserError(
                     _(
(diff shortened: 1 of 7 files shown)


addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+
+        all_moves = StockMove.browse()
+
+        for picking, moves in self.groupby("picking_id"):
+            moves_to_refactor = []
+            grouped_moves = moves._split_and_group_by_weight(maximum_weight)
+            max_range = len(grouped_moves)
+            # Build iterable of grouped moves to refactor, which can be passed into refactor_by_move_groups
+            for idx, moves_to_keep in enumerate(grouped_moves[:max_range]):
+                group_name = f"{picking.name}-{str(idx).zfill(3)}"
```
ws-mps, line 382: f"{picking.name}-{idx:0>3}" will do the same

PeterAlabaster: c14a89540e0a0e1f890c19d7b79515b54fa22b1b

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+                )
+                if rounded_uom_amount != extra_qty:
+                    raise ValidationError(
+                        _(
+                            "Refactor by weight attempted to split a move in a way which the uom (%s) of the product does not support. "
+                            "Wanted to split off %s, the uom would only support splitting off %s"
+                        )
+                        % (move.product_id.uom_id.name, extra_qty, rounded_uom_amount)
+                    )
+                # TODO palabaster: This is post confirm action, so why tf do I need to do this to avoid the UserError? :(
```
ws-mps, line 476: Do we want to resolve this?
What happens if it is triggered on action assign?

PeterAlabaster: Oops, i forgot about this one! I'll look into it now

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+
+        for picking, moves in self.groupby("picking_id"):
+            moves_to_refactor = []
+            grouped_moves = moves._split_and_group_by_weight(maximum_weight)
+            max_range = len(grouped_moves)
+            # Build iterable of grouped moves to refactor, which can be passed into refactor_by_move_groups.
+            # Give each set of moves a unique group name so the pickings are kept separate.
+            for idx, moves_to_keep in enumerate(grouped_moves[:max_range]):
+                group_name = f"{picking.name}-{idx:0>3}"
+                moves_to_refactor.insert(0, (group_name, moves_to_keep))
```
armandcela, line 383: Question: Why do we insert at index 0 moves_to_keep?

PeterAlabaster: No reason, can probably just append as order doesn't matter

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+        Split moves into groups of up to a maximum weight
+        :param maximum_weight: float weight to split and group moves
+        :returns: list of grouped moves
+        """
+        grouped_moves = []
+
+        remaining_moves = self
+
+        # See if any mls are equal to the maximum and add them as individual groups
+        exact_moves = self.filtered(
```
armandcela, line 400: Question: Do we expect to have moves exactly with maximum_height frequently, so that you are trying to handle in a separate filter?

PeterAlabaster: it's just copying how refactor by weight works tbh, and it seemed like a sane check to remove exact matches first before continuing with partial matches. I could probably change this to >= maximum_weight and handle over weight items (as they need to be in their own pick) here trather than inside `_moves_for_weight`

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+            )
+            greater_equal_moves = sorted_moves.filtered(
+                lambda m: float_compare(
+                    (m.product_id.weight * m.product_uom_qty),
+                    weight,
+                    precision_rounding=m.product_uom.rounding,
+                )
+                >= 0
+            )
+            # last one will be at least equal
```
armandcela, line 446: Dont understand this comment, why should be last one at least equal ?

addons/udes_stock_refactoring/models/stock_move.py
PeterAlabaster: 519a5f3a0d8900a9d687b5cb27f590c1169b3b5b

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -359,6 +359,132 @@ def _refactor_action_by_maximum_quantity(self, maximum_qty):
...
+                        _(
+                            "Refactor by weight attempted to split a move in a way which the uom (%s) of the product does not support. "
+                            "Wanted to split off %s, the uom would only support splitting off %s"
+                        )
+                        % (move.product_id.uom_id.name, extra_qty, rounded_uom_amount)
+                    )
+                # TODO palabaster: This is post confirm action, so why tf do I need to do this to avoid the UserError? :(
+                # Use this instead maybe? StockMove._create_split_move()
+                move.state = "confirmed"
+                new_move_vals = move._split(qty=extra_qty)
```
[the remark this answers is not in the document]
armandcela: Sorry didnt spot the while loop

addons/udes_stock_refactoring/models/stock_move.py
PeterAlabaster: I will re-word this

It was a copypasta from refactor by quantity.

All moves in `greater_equal_moves` will be at least equal (due to the previous filter) and the comment doesn't really give us any additional insight..

49f1f92fac55119ca6a74c7b18ee6ce583e988ce

addons/udes_stock_refactoring/models/stock_move.py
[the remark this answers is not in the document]
PeterAlabaster: 6c255601807e158a6ebfe98aa99846d98b2a9946

addons/udes_stock_refactoring/models/stock_move.py
PeterAlabaster: >  What happens if it is triggered on action assign?
I've smoke tested both when Pick is configured to be post assign, and post confirm, and both seem to work as expected

armandcela (approved):

PR #889, opened 2025-02-19
Se 1824 Constrain fields on various models to be positive

PeterAlabaster:


addons/udes_sale/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    NON_NEGATIVE_FIELDS = [
+        "price_unit",
+        "product_uom_qty",
+    ]
+
+    @api.constrains(*NON_NEGATIVE_FIELDS)
+    def _constrain_non_negative_values(self):
```
PeterAlabaster, line 14: self nit; duplicated code in 3 files

armandcela: Can we add the constrain in models ? 
and NON_NEGATIVE_FIELDS = []
In all classes the specific NON_NEGATIVE_FIELDS ?

armandcela (approved):

addons/udes_sale/models/sale_order.py
ws-mps: Good idea, we probably can do something generic. Let's create an issue to properly design this, since it could be a property on the field too.

PeterAlabaster: i think we still need to define the `@api.constrains(...)` on every model, but we could at least split out the check into a basemodel function

ws-mps (approved):

addons/udes_sale/models/sale_order.py
PeterAlabaster: https://unipart.atlassian.net/browse/SE-1824 raised

PR #890, opened 2025-02-20
Se 1751

armandcela:


addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -475,17 +475,22 @@ def _get_task_grouping_criteria(self):
 
         return lambda ml: tuple(chain(*[part(ml) for part in parts]))
 
-    def get_available_move_lines(self):
+    def get_available_move_lines(self, not_available_move_line_ids=None):
```
PeterAlabaster, line 478: nit: rename `excluded_move_line_ids`

addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -71,13 +71,13 @@ class StockPickingBatch(models.Model):
         required=False,
         help="Name of the batch from which this batch was derived",
     )
-    u_picking_zone_id = fields.Many2one(
```
PeterAlabaster, line 74: Q: Have we made sure we have changed all instances of `u_picking_zone_id` to the new field format? Looks like there are some still in `addons/udes_stock_refactoring/tests/test_refactoring.py`, may be others too. I might be wrong as i'm looking at my local code rather than SE-1770 though

addons/udes_stock_picking_batch/models/stock_picking_batch.py
armandcela: Will rename in all places used

PeterAlabaster: I don't know if it makes sense as a change everywhere, haven't checked them all, but yeah if you could just tweak to reflect what it's purpose is that would be great, thanks! :smile:

addons/udes_stock_picking_batch/models/stock_picking_batch.py
armandcela: The field is in 2 models, in stock.picking.batch and stock.location. Renamed in stock.picking.batch model, i think i checked the remaining ones are in location. Hopefully havent forgot any

addons/udes_stock_picking_batch/models/stock_picking_batch.py
armandcela: excluded sounds better to me, didnt think for that at the beginning :(

addons/udes_stock_picking_batch/models/stock_picking_batch.py
PeterAlabaster: Cool, just wanted to double check! Thanks

ws-mps (approved):

PR #893, opened 2025-02-25
[TEST] udes_stock: Reset received location storage format

adamrpatrick:
Ensures that the correct location target storage format is used for receive locations.


armandcela (approved):

addons/udes_stock/tests/common.py
```diff
@@ -172,6 +172,8 @@ def _setup_locations(cls):
         cls.stock_location.write({"name":"TEST_STOCK"})
 
         cls.received_location = cls.stock_location.copy({"name": "TEST_INPUT"})
+        # Reset the location storage format so we default to picking type
+        cls.received_location.u_location_storage_format = False
```
ws-mps, line 176: We can set `u_location_storage_format` in the previous copy.

PeterAlabaster (approved):

PR #897, opened 2025-03-04
[ADD] udes_sale_stock, udes_stock_cron: Full sale reservation

armandcela:
SE-1844

### DIFF
--- a/addons/udes_sale_stock/models/stock_picking.py
+++ b/addons/udes_sale_stock/models/stock_picking.py
@@ -21,3 +21,15 @@ class StockPicking(models.Model):
             return f"Sale Order(s) {''.join(so_names)} have incomplete pickings in previous stages."
         else:
             return False
+
+    def add_more_pickings_to_reserve(self, picking_type):
+        """
+        Extend pickings by inheriting with super, to add more pickings depending on picking type configurations
+        """
+        pickings = super().add_more_pickings_to_reserve(picking_type)
+        if picking_type.u_full_sale_reservation:
+            # Adding more pickings and not overriding to make sure that if pickings are not linked with sale orders.
+            pickings |= pickings.move_lines.sale_line_id.order_id.picking_ids.filtered(
+                lambda p: p.picking_type_id.u_full_sale_reservation and p.state == "confirmed"
+            )
+        return pickings
--- a/addons/udes_sale_stock/models/stock_picking_type.py
+++ b/addons/udes_sale_stock/models/stock_picking_type.py
@@ -12,3 +12,8 @@ class PickingType(models.Model):
 
     # Adding suggest by customer on suggest locations policy options
     u_suggest_locations_policy = fields.Selection(selection_add=[SUGGEST_BY_CUSTOMER])
+    u_full_sale_reservation = fields.Boolean(
+        string="Full Sale Reservation",
+        default=False,
+        help="Flag to indicate that all picks which are created from same sale order will be reserved at same time.",
+    )
--- a/addons/udes_sale_stock/tests/common.py
+++ b/addons/udes_sale_stock/tests/common.py
@@ -1,0 +1,35 @@
+from odoo.addons.udes_stock.tests.common import BaseUDES
+
+
+class BaseSaleUDES(BaseUDES):
+
+    """
+    Creating a BaseSaleUDES testing class on top of BaseUDES which we can reuse when we need to have BaseUDES with
+    some helper methods related with sale order.
+    """
+
+    @classmethod
+    def create_sale_line(cls, sale, product, qty, **kwargs):
+        """Create a sale order line and attach it a sale order"""
+        create_values = {
+            "name": product.name,
+            "order_id": sale.id,
+            "product_id": product.id,
+            "product_uom_qty": qty,
+            "product_uom": product.uom_id.id,
+            "price_unit": product.list_price,
+        }
+        create_values.update(kwargs)
+        return cls.env["sale.order.line"].create(create_values)
+
+    @classmethod
+    def create_sale(cls, customer, **kwargs):
+        """Create a sale order"""
+        create_values = {
+            "partner_id": customer.id,
+            "partner_invoice_id": customer.id,
+            "partner_shipping_id": customer.id,
+            "pricelist_id": cls.env.ref("product.list0").id,
+        }
+        create_values.update(kwargs)
+        return cls.env["sale.order"].create(create_values)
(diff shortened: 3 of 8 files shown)


addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -30,54 +30,55 @@ def _find_unreservable_moves(self, batch_size=1000, return_fast=False):
 
         # Create empty record sets for moves.
         unreservable_moves = Move.browse()
+        for picking_type, pickings in self.groupby(lambda p: p.picking_type_id):
+
+            location = picking_type.default_location_src_id
+            stock_for_products = defaultdict(int)
```
ws-mps, line 36: Should this go outside? If two picking types have the same source location, we could potentially count the same stock as available twice.

addons/udes_sale_stock/models/stock_picking.py
```diff
@@ -21,3 +21,15 @@ def warn_picking_any_previous_pickings_not_complete_sale(self):
...
+
+    def add_more_pickings_to_reserve(self, picking_type):
+        """
+        Extend pickings by inheriting with super, to add more pickings depending on picking type configurations
+        """
+        pickings = super().add_more_pickings_to_reserve(picking_type)
+        if picking_type.u_full_sale_reservation:
+            # Adding more pickings and not overriding to make sure that if pickings are not linked with sale orders.
+            pickings |= pickings.move_lines.sale_line_id.order_id.picking_ids.filtered(
+                lambda p: p.picking_type_id.u_full_sale_reservation
```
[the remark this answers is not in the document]
armandcela: Should not be any as far as they are going to be full reserved, so there will be all unreserved or none reserved

addons/udes_sale_stock/models/stock_picking_type.py
```diff
@@ -12,3 +12,8 @@ class PickingType(models.Model):
 
     # Adding suggest by customer on suggest locations policy options
     u_suggest_locations_policy = fields.Selection(selection_add=[SUGGEST_BY_CUSTOMER])
+    u_full_sale_reservation = fields.Boolean(
+        string="Full Sale Reservation",
+        default=False,
+        help="Flag to indicate that all picks which are created from same sale order will be reserved at same time .",
```
PeterAlabaster, line 18: nit:
```suggestion
        help="Flag to indicate that all picks which are created from same sale order will be reserved at same time.",
```

addons/udes_sale_stock/tests/common.py
```diff
@@ -0,0 +1,38 @@
+from odoo.addons.udes_stock.tests.common import BaseUDES
+
+
+class BaseSaleUDES(BaseUDES):
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Creating a BaseSaleUDES testing class on top of BaseUDES which we can reuse when we need to have BaseUDES with
```
PeterAlabaster, line 9: this docstring can live on the class and the `setUpClass(cls):` function can be removed

addons/udes_sale_stock/tests/test_full_sale_reservation.py
```diff
@@ -0,0 +1,194 @@
...
+                "usage": "internal",
+                "location_id": picking_zone2.id,
+            }
+        )
+        cls.picking_type_pick.default_location_src_id = picking_zone
+        cls.picking_type_pick2 = cls.picking_type_pick.copy({"default_location_src_id": picking_zone2.id})
+
+        products_info = [{"product": cls.apple, "uom_qty": 10.0}]
+
+        cls.picking_type_pick.u_num_reservable_pickings = -1
```
PeterAlabaster, line 55: nit: if picking_type_pick2 (the copy) has all this config the same, we could .copy() after setting it on picking_type_pick to avoid needing to set it all twice

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -265,9 +268,47 @@ def _process_pickings(self):
         if unsatisfied:
             # Unreserve if the picking type cannot handle partials or it
             # can but there is nothing allocated (no stock.move.lines)
-            if not self.picking_type_id.u_handle_partials or not mls:
+            # If there are several picking types, they should have all same handle_partials,
+            # can be improved to handle it better if necessary.
+            handle_partial = self.picking_type_id[0].u_handle_partials
```
PeterAlabaster, line 273: nit, should pobably enforce this, not sure if there is a better way
```python
handle_partial = set(picking_types.mapped("u_handle_partials"))
if len(handle_partial) > 1:
    raise ...
if handle_partial = {False} or not mls:
```

armandcela: We had a discussion about handle partials that should be all same, but i think you are right better safe than sorry

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -265,9 +268,47 @@ def _process_pickings(self):
...
+
+        Args:
+            picking_type (recordset): Stock picking type recordset.
+        Returns:
+            pickings (recordset): Picking to reserve recordset
+        """
+        pickings = self
+        if self.batch_id:
+
+            if self.u_reserve_batches:
```
PeterAlabaster, line 294: nit: we should combine these branches unless we plan on changing behaviour of one over the other in future
```
if self.u_reserve_batches or picking_type.u_reserve_batches:
    pickings = pickings.batch_id.picking_ids
```

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -265,9 +268,47 @@ def _process_pickings(self):
...
+        """
+        Checking different configuration of the picking type, and add processed pickings to be skipped when reserving.
+        Method can be override on specific implementations to extend pickings that are skipped.
+
+        Args:
+            processed_pickings  (recordset): Stock picking recordset.
+        Returns:
+            processed_pickings  (recordset): Stock picking recordset.
+        """
+        if self.batch_id and self.batch_id.state == "draft":
```
PeterAlabaster, line 312: nit: if `self.batch_id` is emptyset, then `self.batch_id.state` will equal False. I don't think we need check for `if self.batch_id and`

PeterAlabaster (approved):

addons/udes_stock_cron/models/stock_picking.py
```diff
@@ -30,54 +30,55 @@ def _find_unreservable_moves(self, batch_size=1000, return_fast=False):
...
-                        and not picking.picking_type_id.u_handle_partials
-                    ):
-                        unreservable_moves |= move
-                        if return_fast:
-                            return unreservable_moves
-                    stock_for_products[product] -= qty_ordered
+        for picking_type, pickings in self.groupby(lambda p: p.picking_type_id):
+
+            location = picking_type.default_location_src_id
+            skip_states = ("assigned", "done", "cancel")
```
ws-mps, line 37: This can go outside too.

ws-mps (approved):

PR #904, opened 2025-03-18
[ADD] udes_get_info:2 optional parameters in _get_info method

armandcela:
SE-1687

### DIFF
--- a/addons/udes_get_info/models/models.py
+++ b/addons/udes_get_info/models/models.py
@@ -40,7 +40,7 @@ def get_info(self, level=1, info_fields=frozenset(), extra_fields=frozenset(), *
 
 
 @add_if_not_exists(models.BaseModel)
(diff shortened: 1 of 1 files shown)


addons/udes_get_info/models/models.py
```diff
@@ -40,7 +40,7 @@ def get_info(self, level=1, info_fields=frozenset(), extra_fields=frozenset(), *
 
 
 @add_if_not_exists(models.BaseModel)
-def _get_info(self, level, info_fields, extra_fields):
+def _get_info(self, level, info_fields=frozenset(), extra_fields=frozenset(), **kwargs):
```
kdwyer, line 43: Why do we need this change?  We shouldn't be calling `_get_info` directly, and `get_info` always passes these arguments.

armandcela:
Closed MR as per discussion with Kevin not needed

PR #910, opened 2025-03-28
[ADD] udes_stock_packaging: Check package contains single item method

armandcela:
SE-1889

### DIFF
--- a/addons/udes_stock_packaging/models/stock_quant_package.py
+++ b/addons/udes_stock_packaging/models/stock_quant_package.py
@@ -122,6 +122,25 @@ class StockQuantPackage(models.Model):
             res[quant.product_id] += quant.quantity
         return res
 
+    def check_package_contains_single_item(self, raise_error=True):
+        """
+        Check if package has one item only.
+        Raise Error if raise_error boolean optional parameter is True.
+        """
+        self.ensure_one()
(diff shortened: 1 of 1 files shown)


PeterAlabaster (approved):

addons/udes_stock_packaging/models/stock_quant_package.py
```diff
@@ -122,6 +122,25 @@ def _get_all_products_quantities(self):
...
         return res
 
+    def check_package_contains_single_item(self, raise_error=True):
+        """
+        Check if package has one item only.
+        Raise Error if raise_error boolean optional parameter is True.
+        """
+        self.ensure_one()
+        product_qtys = self._get_product_quantities()
+        if len(product_qtys) != 1 or sum(product_qtys.values()) != 1:
```
kdwyer, line 132: Do we need to be concerned with float qtys?

armandcela: This was already in Jayshree's story, i just wrapped into a method, changed the method name and in both places called the new method. 
Not sure if we will have qty with floats like 0.5 or 1.3 etc, i think in all cases we should use box, didnt want to possible change behaviour on smth that is already merged

kdwyer (approved):

PR #913, opened 2025-04-01
[IMP] udes_stock_packaging: Make result_package_name optional

ws-mps:
If it is already set in the result, no need to create new seq.

### DIFF
--- a/addons/udes_stock_packaging/models/stock_picking.py
+++ b/addons/udes_stock_packaging/models/stock_picking.py
@@ -122,7 +122,8 @@ class StockPicking(models.Model):
             f"**Weight:** {float_repr(float_round(weight_kg + container.weight, 2), 2)}kg",
             f"**Dimensions:** Length {container.length}mm x Width {container.width}mm x Height {container.height}mm",
         ]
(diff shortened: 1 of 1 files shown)


kdwyer (approved):

addons/udes_stock_packaging/models/stock_picking.py
```diff
@@ -122,7 +122,8 @@ def get_pallet_move_details_extra_info(self, result, package, container_id=None)
             f"**Weight:** {float_repr(float_round(weight_kg + container.weight, 2), 2)}kg",
             f"**Dimensions:** Length {container.length}mm x Width {container.width}mm x Height {container.height}mm",
         ]
-        result_package_name = IrSequence.next_by_code("stock.quant.package.parcel")
-        # Add the resulting package name to context, so it can be marked on the picked lines later.
-        res["result_package_name"] = result_package_name
+        if "result_package_name" not in result:
```
PeterAlabaster, line 125: nit, result gets updated by the super call to variable `res`. we should be checking against that incase any other hooks on `get_pallet_move_details_extra_info` update it
```suggestion
        if "result_package_name" not in res:
```

PeterAlabaster (approved):

PR #915, opened 2025-04-07
[ADD] udes_stock: Backorder when multi users is enabled

armandcela:
SE-1908

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -508,11 +508,31 @@ class StockPicking(models.Model):
         to create backorders when pickings are validated regardless of the user's generic
         permissions.
         """
+        multi_users_enabled = self.picking_type_id.u_multi_users_enabled
         # If configuration is enabled, copy the batch to the backorder too.
         batch_id = False
         if self.picking_type_id.u_preserve_backorder_batch:
             # If the picking has no batch, batch_id will be False.
             batch_id = self.batch_id.id
+        # If multi users config is enabled, moves to keep will stay in the existing picking and others will be
+        # transferred in the new pick.
+        if multi_users_enabled:
+            # Finding all moves of existing picking and moves which might be created during split moves when moves have
+            # been partially done.
+            all_moves = self.move_lines | moves
+            # Instead of moving moves to the backorder, when multi users is enabled we want to keep same picking as
+            # another user might be working on same time. That's why we let the moves that needs picking in the original
+            # picking and moving all the others to the backorder picking.
+            moves = all_moves - moves
+            # It might be that moves has been partially done, and when creating the backorder it splits the moves by
+            # creating new moves which are added to the existing picking, so that other users continue picking with
+            # what is left.
+            new_moves = all_moves - self.move_lines
+            if new_moves:
+                self.write({
+                    "move_lines": [(4, move.id) for move in new_moves],
+                    "move_line_ids": [(4, move_line.id) for move_line in new_moves.move_line_ids]
+                })
         vals = {
             "name": "/",
             "move_lines": [(6, 0, moves.ids)],
@@ -919,6 +939,7 @@ class StockPicking(models.Model):
         Picking = self.env["stock.picking"]
 
         new_picking = Picking.browse()
+        picking_to_validate = self
         requires_backorder = False
         mls_to_keep = None
         with self.statistics() as stats:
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_picking.py
```diff
@@ -508,11 +508,25 @@ def _create_backorder_picking(self, moves, **kwargs):
...
         # If configuration is enabled, copy the batch to the backorder too.
         batch_id = False
         if self.picking_type_id.u_preserve_backorder_batch:
             # If the picking has no batch, batch_id will be False.
             batch_id = self.batch_id.id
+        # If multi users config is enabled, moves to keep will stay in the existing picking and others will be
+        # transferred in the new pick.
+        if multi_users_enabled:
+            all_moves = self.move_lines | moves
+            moves = all_moves - moves
```
kdwyer, line 521: Do we need this line to remove any duplicates?  A comment might be helpful.

addons/udes_stock/models/stock_picking.py
```diff
@@ -508,11 +508,25 @@ def _create_backorder_picking(self, moves, **kwargs):
         to create backorders when pickings are validated regardless of the user's generic
         permissions.
         """
+        multi_users_enabled = self.picking_type_id.u_multi_users_enabled
         # If configuration is enabled, copy the batch to the backorder too.
         batch_id = False
         if self.picking_type_id.u_preserve_backorder_batch:
             # If the picking has no batch, batch_id will be False.
             batch_id = self.batch_id.id
+        # If multi users config is enabled, moves to keep will stay in the existing picking and others will be
```
PeterAlabaster, line 517: Q: Why? Could we document the reason why we do this?

addons/udes_stock/models/stock_picking.py
```diff
@@ -952,11 +967,16 @@ def validate_picking(self, create_backorder=False, force_validate=False):
                 # NOTE: Added sudo() - Uses sudo() here as a user might not have the full access rights to stock.picking
                 # but still needs more access rights for the flow
                 new_picking = self.sudo().backorder_move_lines(mls_to_keep=mls_to_keep)
+                # Picking to validate and backorder picking as when the config is enabled, moves to keep will stay in
```
PeterAlabaster, line 970: Not sure I understand the first part of this comment:
>  Picking to validate and backorder picking as when the config is enabled, moves to keep will stay in the existing picking

addons/udes_stock/models/stock_picking.py
armandcela: Added some other comments on below lines which hopefully explains what is done

addons/udes_stock/models/stock_picking.py
armandcela: Not sure I understood the comment, added some more comments and maybe detailed more which hopefully can make more sense :(

PeterAlabaster: I think kevin is asking why do we need this line?

Could we not just say `moves = self.move_lines`.

I think if there is crossover e.g
```python
self.moves_lines = stock.move(1,2,3)
moves = stock.move(3,4,5)
```
Then moves should become `stock.move(1,2)`, which is why we need this extra operation?

PeterAlabaster: i wonder if we could just do
```python
new_moves = moves
moves -= self.move_lines
```

PeterAlabaster: i wonder if this whole block would be better off like this;
```python
        # When this config is enabled, we need to allow other users to continue picking,
        # so the moves to backorder need to be swapped from the unfinished moves (moves)
        # to everything else.
        if multi_users_enabled:
            new_moves = moves
            moves = self.move_lines - new_moves
            # What we were originally going to backorder should be instead attached back to the picking in self
            if new_moves:
                self.write({
                    "move_lines": [(4, move.id) for move in new_moves],
                    "move_line_ids": [(4, move_line.id) for move_line in new_moves.move_line_ids]
                })
```

PeterAlabaster: Ah no my previous comment on how it works is wrong:
```python
self.move_lines = stock.move(1,2,3)
moves = stock.move(3,4,5)

# Current
all_moves = self.move_lines | moves # stock.move(1,2,3,4,5)
moves = all_moves - moves # stock.move(1,2)
new_moves = all_moves - self.move_lines # stock.move(4,5)

# Alt (suggested)
new_moves = moves # stock.move(3,4,5)
moves = self.move_lines - new_moves # stock.move(1,2)
```
So we probably need to keep it the way it is. We should just explain that we want to remove crossover moves both from new_moves and moves then. I have edited my suggestion in my previous comment

armandcela: New moves are not moves, new moves that will be added in the above example are stock.move(4,5).
And moves to be moved into the backorder are stock.move(1,2) which are done moves.
stock.move(3) will stay where it was, so no change on that as needs to be done yet that why is in the moves variable.

PeterAlabaster: I can't read it the way it is at the minute - I find the comments make it harder to understand, but I'm not going to block it going forwards and have thumbed

PeterAlabaster (approved):

PR #916, opened 2025-04-10
Se 1831

armandcela:


addons/udes_stock/models/product_template.py
```diff
@@ -26,6 +26,26 @@ def _constrain_non_negative_values(self):
...
+            tracking_types_mapping = {
+                "none": "No Tracking",
+                "lot": "Lots",
+                "serial": "Serial Number"
+            }
+            # Clearing allowed tracking types which are not possible option of tracking types.
+            allowed_tracking_types = [
+                tracking for tracking in allowed_tracking_types_list if tracking in tracking_types_mapping.keys()
+            ]
+            if product_template.tracking not in allowed_tracking_types:
```
PeterAlabaster, line 44: just checking, this gets run when modifying `product.product` records too?

addons/udes_stock/models/product_template.py
```diff
@@ -72,6 +92,12 @@ def unlink(self):
         """Override superclass to prevent deletion."""
         raise ValidationError(_("Products may not be deleted. Please archive them instead."))
 
+    def write(self, values):
+        result = super().write(values)
+        if values.get("active"):
```
PeterAlabaster, line 97: Q: Does the constraint not get run on inactive products?

PeterAlabaster: Oh, it's not SQL constraint, so only gets run when the field changes. So is this to do with the fact we may have instances where the constraint is violated on an inactive product and we do not want them to be able to be re-activated? Maybe a better solution is to `@api.constrains("tracking", "active")` on the constraint itself

PeterAlabaster: > No because api.constrains is triggered on change of field tracking. In case i add active field in the api constrains, it wont allow to archive existing products, or need to check on the constrain the value of active

Are you sure? The user can't configure the warehouse to restrict those tracking options if there are products active with them, so the user will get the error on stock.warehouse, go to products (who _currently_ have a valid tracking, and then archive it. It is still valid). Now they can go and change the warehouse config as inactive products are ignored

ws-mps: You can put active in the list of fields to trigger the constrian too

kdwyer (approved):

PR #917, opened 2025-04-10
[FIX] udes_stock: Ensure product quantity on lot only appears on rows which are relevant 

dandihar:


PeterAlabaster (approved):

PeterAlabaster:
We should squash these 3 commits down to 1

PR #920, opened 2025-04-28
[IMP] udes_stock: New field u_total_weight added in the Picking model

JayshreeMahajan:
SE-1917

### DIFF
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -162,6 +162,13 @@ class StockPicking(models.Model):
         copy=True,
     )
 
+    u_total_weight = fields.Float(
+        string="Total Weight",
+        compute="_compute_total_weight",
+        store=False,
+        help="Total weight based on moves and product weight",
+    )
+
     def get_next_picking_name(self, vals, picking_type=None):
         """
         Override the method in core stock to customise the name generation.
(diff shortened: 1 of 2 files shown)


kdwyer (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -162,6 +162,13 @@ class StockPicking(models.Model):
         copy=True,
     )
 
+    u_total_weight = fields.Float(
```
armandcela, line 165: Nit: Black formatting

PeterAlabaster (approved):

addons/udes_stock/models/stock_picking.py
```diff
@@ -179,6 +186,14 @@ def get_next_picking_name(self, vals, picking_type=None):
                 return get_next_name(self, "stock.picking", sequence=ir_sequence)
         return super().get_next_picking_name(vals, picking_type=picking_type)
 
+    @api.depends('move_lines.product_id', 'move_lines.product_uom_qty', 'move_lines.product_id.weight')
+    def _compute_total_weight(self):
+        for picking in self:
+            total = 0.0
+            for move in picking.move_lines:
+                total += move.product_uom_qty * move.product_id.weight
+            picking.u_total_weight = total
```
armandcela, line 195: move_weights = [move.product_uom_qty*move.product_id.weight for move in picking.move_lines]
picking.u_total_weight = sum(move_weights)

addons/udes_stock/models/stock_picking.py
```diff
@@ -179,6 +186,14 @@ def get_next_picking_name(self, vals, picking_type=None):
                 return get_next_name(self, "stock.picking", sequence=ir_sequence)
         return super().get_next_picking_name(vals, picking_type=picking_type)
 
+    @api.depends('move_lines.product_id', 'move_lines.product_uom_qty', 'move_lines.product_id.weight')
```
armandcela, line 189: Don't think that the method gets triggered when product weight changes, it gets called every time on form load and when changing the other 2, but not when changing the weight, because is not a store field and weight is not a field visible in the form view as is inside product form.
```suggestion
    @api.depends('move_lines.product_id', 'move_lines.product_uom_qty')
```

armandcela (approved):

armandcela:
Commit message would be better to be shorter as it cuts and adds in the comment section

PR #922, opened 2025-04-29
[ADD] udes_stock_packaging: Adding package type

armandcela:
SE-1929

### DIFF
--- a/addons/udes_stock_packaging/data/package_types.xml
+++ b/addons/udes_stock_packaging/data/package_types.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0"?>
+<odoo>
+  <data>
+    <!--Pallet package type-->
+    <record id="sequence_pallet" model="ir.sequence">
+      <field name="name">Package Type Pallet</field>
+      <field name="prefix">UDES</field>
+      <field name="padding" eval="5"/>
+      <field name="code">stock.quant.package.pallet</field>
+    </record>
+
+    <record id="pallet_package_type" model="package.type">
+      <field name="name">Pallet</field>
+      <field name="sequence">1</field>
+      <field name="sequence_id" ref="sequence_pallet"/>
+      <field name="package_type_regex">^UDES[0-9]+$</field>
+    </record>
+
+    <!--Package package type-->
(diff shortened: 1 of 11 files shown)


addons/udes_stock_packaging/tests/test_stock_quant_package.py
```diff
@@ -172,6 +172,108 @@ def test_prevents_adding_packages_to_package_via_write(self):
...
+        cls.parcel_package_type = cls.env.ref("udes_stock_packaging.parcel_package_type")
+
+    def test_creates_pallet_with_valid_name(self):
+        """
+        We should be able to create a package of pallet type if we provide a valid name.
+        Pallet regex format restricts that name should start with UDES.
+        """
+        name = "UDES10001"
+
+
```
kdwyer, line 196: Nit: extra empty line

kdwyer (approved):

addons/udes_stock_packaging/data/package_types.xml
```diff
@@ -0,0 +1,80 @@
...
+      <field name="prefix">UDES</field>
+      <field name="padding" eval="5"/>
+      <field name="code">stock.quant.package.pallet</field>
+    </record>
+
+    <record id="pallet_package_type" model="package.type">
+      <field name="name">Pallet</field>
+      <field name="sequence">1</field>
+      <field name="sequence_id" ref="sequence_pallet"/>
+      <field name="package_type_regex">^UDES(?:[0-9])+$</field>
```
kdwyer, line 16: Nit: I'm not sure we need the non-capturing group in these regexes? `^UDES[0-9]+$` should be fine I think.

armandcela: I believe copied it from pallet regex on the warehouse which we have been using, but yeah i agree with you

addons/udes_stock_packaging/tests/test_stock_quant_package.py
armandcela: Pushed the change

dandihar (approved):

PR #923, opened 2025-04-30
[IMP] udes_stock_packaging: Box dims can now be 0

dandihar:
Specifically when use_product_dims is True

SE-1922

### DIFF
--- a/addons/udes_stock_packaging/models/container_type.py
+++ b/addons/udes_stock_packaging/models/container_type.py
@@ -59,14 +59,21 @@ class ContainerType(models.Model):
         ),
     ]
 
-    @api.constrains("length", "width", "height", "weight")
+    @api.constrains("length", "width", "height", "weight", "use_product_dims")
     def _check_dimensions_and_weight(self):
         """
         Ensure these fields do not contain negative quantities else
         it could impact calculations when the container types are used.
+
+        Exception is made for no box (use_product_dims = True)
         """
+
         non_negative_fields = ["length", "width", "height", "weight"]
-        violating_records = self.filtered(lambda ct: any([getattr(ct, f) <= 0 for f in non_negative_fields]))
+        violating_records = self.filtered(
+            lambda ct: not ct.use_product_dims and any(
+                getattr(ct, f) <= 0 for f in non_negative_fields
+                )
+            )
         if violating_records:
             raise ValidationError(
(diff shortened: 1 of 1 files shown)


addons/udes_stock_packaging/models/container_type.py
```diff
@@ -64,9 +64,16 @@ def _check_dimensions_and_weight(self):
...
         Ensure these fields do not contain negative quantities else
         it could impact calculations when the container types are used.
+
+        Exception is made for no box (use_product_dims = True)
         """
+
         non_negative_fields = ["length", "width", "height", "weight"]
-        violating_records = self.filtered(lambda ct: any([getattr(ct, f) <= 0 for f in non_negative_fields]))
+        violating_records = self.filtered(
+            lambda ct: not getattr(ct, "use_product_dims", False) and any(
```
armandcela, line 73: Why is needed getattr, the field use_product_dims is added in this module so i think not needed to use getattr
```suggestion
            lambda ct: not ct.use_product_dims and any(
```

armandcela:
Try to make the commit title smaller, and if needs more explanation write on new lines as if you seen it cuts automatically and doesnt look good.

armandcela:
Now that we are adding another fields on api.constraint we need to modify the api.depends by adding the new field.

addons/udes_stock_packaging/models/container_type.py
```diff
@@ -59,14 +59,21 @@ class ContainerType(models.Model):
...
         Ensure these fields do not contain negative quantities else
         it could impact calculations when the container types are used.
+
+        Exception is made for no box (use_product_dims = True)
         """
+
         non_negative_fields = ["length", "width", "height", "weight"]
-        violating_records = self.filtered(lambda ct: any([getattr(ct, f) <= 0 for f in non_negative_fields]))
+        violating_records = self.filtered(
+            lambda ct: not "use_product_dims" and any(
```
JayshreeMahajan, line 73: This condition doesn't really check the value of ct.use_product_dims. Instead, it's just checking the text "use_product_dims", which is always considered True because it's a non-empty string

```suggestion
            lambda ct: not "use_product_dims" and any(
             lambda ct: not ct.use_product_dims and any(
```

JayshreeMahajan (approved):

PR #925, opened 2025-05-08
[IMP] udes_sale_stock: Confirm sale order cron job now runs every 10 …

dandihar:
…minutes

SE-1927


addons/udes_sale_stock/data/ir_cron.xml
```diff
@@ -7,7 +7,7 @@
       <field name="name">Confirm Orders</field>
       <field name="active" eval="True"/>
       <field name="user_id" ref="base.user_root"/>
-      <field name="interval_number">15</field>
+      <field name="interval_number">10</field>
```
armandcela, line 10: Question: Is this a generic change or only for Konica?

armandcela (approved):

addons/udes_sale_stock/data/ir_cron.xml
dandihar: This change is generic

scott-redshaw (approved):

PR #928, opened 2025-05-13
[IMP] udes_stock: Adding updating move line functionality and related functionality to udes_stock

scott-redshaw:
SE-1920

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -173,6 +173,18 @@ class StockMoveLine(models.Model):
 
         return quants
 
+    def reset_move_line_data(self):
+        """Resets the move_line data for location destination, quantity done, done datetime and
+         done status to default values."""
+        self.u_picking_type_id.ensure_one()
+        default_location_dest = self.u_picking_type_id.default_location_dest_id
+        self.write({
+            "location_dest_id": default_location_dest,
+            "qty_done": 0.00,
+            "u_done_datetime": False,
+            "u_done_by_id": False
+        })
+
     def get_quantities_by_key(self, get_key=lambda ml: ml.product_id):
         """This function computes the different product quantities for the given move lines
         :kwargs:
--- a/addons/udes_stock/models/stock_picking.py
+++ b/addons/udes_stock/models/stock_picking.py
@@ -267,6 +267,22 @@ class StockPicking(models.Model):
         for picking in self:
             picking.u_num_pallets = len(picking.move_line_ids.result_package_id)
 
+    def reset_move_lines_from_pack(self):
+        """Resets move lines from location destination, quantity done, completed 
+         datetime and done status in the packing process to default or Null values
+         from self recordset
+        """
+        done_move_lines = self.move_line_ids.filtered(
+            lambda ml: ml.location_dest_id 
+            and ml.qty_done 
+            and ml.u_done_datetime
+            and ml.u_done_by_id
+        )
+        if done_move_lines:
+            done_move_lines.reset_move_line_data()
+        return True
+
+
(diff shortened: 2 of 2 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -173,6 +173,15 @@ def get_quants(self):
 
         return quants
 
+    def update_move_line(self, default_location_dest):
```
kdwyer, line 176: We need a better name here.  `reinitialise_move_line` tells us more about the update that is taking place, but it is still a bit generic.  Ideally we want something that uses the language of the business domain.  How about `discard_accumulated_work` or similar?

ws-mps: How about `reset_move_line_data`?

addons/udes_stock/models/stock_picking.py
```diff
@@ -267,6 +267,22 @@ def _compute_num_pallets(self):
...
+        """
+        picking = self.browse(picking_id)
+        done_move_lines = picking.move_line_ids.filtered(
+            lambda ml: ml.location_dest_id 
+            and ml.qty_done 
+            and ml.u_done_datetime
+            and ml.u_done_by_id
+        )
+        done_move_lines.update_move_line(picking.picking_type_id.default_location_dest_id.id)
+        return True
```
armandcela, line 283:
```suggestion
    def update_move_lines_from_pack(self):
        """ Updates move lines from location destination, quantity done, completed 
         datetime and done status in the packing process to default or Null values
         from self recordset
        """
        done_move_lines = self.move_line_ids.filtered(
            lambda ml: ml.location_dest_id 
            and ml.qty_done 
            and ml.u_done_datetime
            and ml.u_done_by_id
        )
        done_move_lines.update_move_line(self.picking_type_id.default_location_dest_id.id)
        return True
```

addons/udes_stock/models/stock_move_line.py
kdwyer: Let's go with that then

addons/udes_stock/models/stock_picking.py
```diff
@@ -267,6 +267,21 @@ def _compute_num_pallets(self):
...
+         datetime and done status in the packing process to default or Null values
+         from self recordset
+        """
+        done_move_lines = self.move_line_ids.filtered(
+            lambda ml: ml.location_dest_id 
+            and ml.qty_done 
+            and ml.u_done_datetime
+            and ml.u_done_by_id
+        )
+        done_move_lines.reset_move_line_data(self.picking_type_id.default_location_dest_id.id)
```
armandcela, line 281: Question:

Do we need to pass the default_location_dest_id in this method as parameter, i think is not needed.
In reset_move_line_data can say:
default_location_dest = self.u_picking_type_id.default_location_dest_id

addons/udes_stock/models/stock_move_line.py
```diff
@@ -173,9 +173,10 @@ def get_quants(self):
 
         return quants
 
-    def reset_move_line_data(self, default_location_dest):
+    def reset_move_line_data(self):
         """Resets the move_line data for location destination, quantity done, done datetime and
          done status to default values."""
+        default_location_dest = self.u_picking_type_id.default_location_dest_id
```
armandcela, line 180: This will raise error in case that move lines will be off different picking type. We might reuse this method in future and would be good to add a ensure_one for the picking type so we know that we need to use this method when resetting move lines of same picking type.

self.u_picking_type_id.ensure_one()

armandcela (approved):

kdwyer (approved):

adamrpatrick (approved):

PR #930, opened 2025-05-19
[ADD] udes_stock: UDES 14 Force upper case for internal ref and barcode

DarpanBhatt-UP:
SE-1930


addons/udes_stock/models/stock_warehouse.py
```diff
@@ -56,6 +56,21 @@ def _domain_u_damaged_location_id(self):
         help="Allowed tracking types on product level, values are comma separated and can use any combination of 3 "
              "possible options none, lot,serial."
     )
+    u_force_upper_case_config = fields.Text(
```
PeterAlabaster, line 59: suggestion: use a many2many table of `model_id` and `field_ids` instead of text, so we do not need to handle json parse errors

addons/udes_stock/models/common.py
```diff
@@ -83,3 +82,36 @@ def get_next_name(obj, code, sequence=None):
...
+        containing valid UDES fields and value as a key value pair
+        user -> Current UDES user
+        """
+    wh = user.get_user_warehouse()
+    # There will always be only one warehouse as per get_user_warehouse()
+    if wh.u_force_upper_case_field_ids:
+        #field names can be similar in different models filter out records based on which active_model
+        #validation request is coming from
+        for field in wh.u_force_upper_case_field_ids.filtered(lambda x: x.model_id.model == active_model):
+            if vals.get(field.name):
```
ws-mps, line 112: Can we put this in a variable? it is used 3 times.

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -56,21 +63,13 @@ def _domain_u_damaged_location_id(self):
...
-        Below example will force upper case on fields default_code and barcode of UDES models product.product and
-        stock.location. At the moment this validation is available on UDES models product.product 
-        and stock.location
-        {
-          "product.product":  "default_code,barcode",
-          "stock.location": "barcode",
-        }
-        """
-    )
+    u_force_upper_case_field_ids = fields.Many2many('ir.model.fields',
```
PeterAlabaster, line 66: nit: this doesn't look like black formatting, could we put first param on a newline so there's not loads of whitespace to the left? Thanks!

DarpanBhatt-UP: let me auto format

addons/udes_stock/models/common.py
DarpanBhatt-UP: Please review latest commit

PR #931, opened 2025-05-19
[ADD] udes_stock: Pick by multiple measures functionality 

armandcela:
When use multiple measures config is enabled.

SE-1983


PeterAlabaster (approved):

addons/udes_stock/models/stock_move_line.py
```diff
@@ -11,6 +11,13 @@
 
 _logger = logging.getLogger(__name__)
 
+measure_type_options = [
```
ws-mps, line 14: This usually is uppercase

ws-mps (approved):

PR #933, opened 2025-05-22
[IMP] udes_sale_stock: Refactor get_sale_order to retrieve all move l…

JayshreeMahajan:
…ines regardless of state

SE-1939

### DIFF
--- a/addons/udes_stock/models/product_product.py
+++ b/addons/udes_stock/models/product_product.py
@@ -91,3 +91,35 @@ class ProductProduct(models.Model):
         )
 
         return bool(stock_pickings) or self.get_quant_counts()
+
+    def convert_measure_type_quantity(self, quantity, measure_type):
+        """
+        Getting the quantity in eaches when user selects a measure type, and the quantity is for measure selected.
+        """
+        self.ensure_one()
+        measure_qty = 0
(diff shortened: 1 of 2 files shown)


dandihar (approved):

addons/udes_sale_stock/models/stock_quant_package.py
```diff
@@ -8,6 +8,6 @@ def get_sale_order(self):
         Getting sale order from a package.
         """
         self.ensure_one()
-        mls = self.get_move_lines_of_children([("state",  "=", "assigned")])
+        mls = self.get_move_lines_of_children()
```
scott-redshaw, line 11: As there are no parameters does this mean the move lines could be done at any stage of the process?

JayshreeMahajan: yes

scott-redshaw: Just thinking if this would cause any errors if function was ran when the system is in state ready or waiting?

JayshreeMahajan: I have added candition for done and assign state

scott-redshaw (approved):

PR #937, opened 2025-06-10
Se 2022

armandcela:


addons/udes_stock/models/product_barcode.py
```diff
@@ -0,0 +1,23 @@
+from odoo import fields, models, api, _
+from odoo.exceptions import ValidationError
+from . common import check_upper_case_validation
```
PeterAlabaster, line 3: nit: unused import

addons/udes_stock/models/product_template.py
```diff
@@ -78,6 +87,25 @@ def _constrain_tracking_in_allowed_tracking_values(self):
                         % tracking_types_mapping[product_template.tracking]
                     )
 
+    def _compute_multiple_barcodes_config(self):
+        """Computed field to show the configuration set on the warehouse for multiple barcodes."""
+        warehouse = self.env.ref("stock.warehouse0")
```
PeterAlabaster, line 92: nit: we should use `self.env.user.get_user_warehouse()`

addons/udes_stock/models/product_template.py
```diff
@@ -78,6 +87,25 @@ def _constrain_tracking_in_allowed_tracking_values(self):
...
+        for product in self:
+            product.u_multiple_barcodes = warehouse.u_product_multiple_barcodes
+
+    @api.depends("product_variant_ids", "product_variant_ids.u_barcode_ids")
+    def _compute_barcode_ids(self):
+        for p in self:
+            if len(p.product_variant_ids) == 1:
```
PeterAlabaster, line 99: Q: If there is more than 1 variant, do we set u_barcode_ids on the variant instead?

addons/udes_stock/models/stock_warehouse.py
```diff
@@ -6,6 +6,12 @@ class StockWarehouse(models.Model):
     _name = "stock.warehouse"
     _inherit = ["stock.warehouse", "mail.thread"]
 
+    u_product_multiple_barcodes = fields.Boolean(
+        string="Product Multiple Barcodes",
+        help="Flag to enable usage of multiple barcodes per same product.",
```
PeterAlabaster, line 11:
```suggestion
        help="Flag to enable usage of multiple barcodes on a single product.",
```

addons/udes_stock/models/product_product.py
```diff
@@ -134,3 +135,61 @@ def write(self, vals):
...
+        if warehouse.u_product_multiple_barcodes:
+            # Using in operator when searching for the product, as the first one might be a new barcode.
+            # We want to make sure product doesn't exist.
+            if not isinstance(barcode, list):
+                barcode = [barcode]
+            product_barcodes = ProductBarcode.search([("name", "in", barcode)], order="id")
+            if product_barcodes:
+                product = product_barcodes.product_id
+                # Barcodes should be of the same product or not on any product
+                product.ensure_one()
```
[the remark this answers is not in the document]
armandcela: Responded before seeing the custom validation error, ok agreed with validation error message

addons/udes_stock/models/product_product.py
```diff
@@ -134,3 +135,61 @@ def write(self, vals):
...
+        Get barcode of the product depending on configuration
+        """
+        self.ensure_one()
+        if self.u_multiple_barcodes:
+            barcode = self.u_barcode_ids and self.u_barcode_ids[0].name or None
+        else:
+            barcode = self.barcode
+        return barcode
+
+    def get_list_barcodes(self):
```
kdwyer, line 149: What do you think about `get_all_barcodes` for the name of this method?  I think it would make it clearer that we are getting all the barcodes for a product, rather than just some list of barcodes?

PeterAlabaster: +1

PeterAlabaster (approved):

addons/udes_stock/models/product_product.py
```diff
@@ -134,3 +135,61 @@ def write(self, vals):
...
+        self.ensure_one()
+        if self.u_multiple_barcodes:
+            list_barcodes = self.u_barcode_ids.mapped("name")
+        else:
+            # Always returning a list, even when only one barcode is in it.
+            list_barcodes = [self.barcode]
+        return list_barcodes
+
+    @api.model
+    def get_product_from_barcode(self, barcode):
```
kdwyer, line 162: How about `get_by_barcode`?  It's a method on the `Product` model, so it should be obvious that it will return a product?

PR #942, opened 2025-06-27
[IMP] udes_stock: Set done qty on move lines via a loop

DarpanBhatt-UP:
SE-2026

### DIFF
--- a/addons/udes_stock/tests/test_stock_move_line.py
+++ b/addons/udes_stock/tests/test_stock_move_line.py
@@ -1675,6 +1675,132 @@ class TestStockMoveLinePrepareAndMarkMoveLines(common.BaseUDES):
                 product_ids=product_ids, location=location, result_package=pallet_to_pick_onto
             )
 
+    def test_picking_with_zero_move_line_done_qty(self):
+        """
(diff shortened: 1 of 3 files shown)


addons/udes_stock/tests/test_stock_move_line.py
```diff
@@ -1675,6 +1675,132 @@ def test_user_cannot_swap_in_more_stock_then_on_the_system(self):
...
+        banana_barcode = self.banana.barcode
+        apple_ml = self.apple_move_line
+        banana_ml = self.banana_move_line
+        result_package = self.create_package()
+        location_dest = self.test_goodsout_location_01
+        scan_banana = [{"barcode": banana_barcode, "uom_qty": 5}]
+        scan_apple = [{"barcode": apple_barcode, "uom_qty": 0}]
+        prepare_apple = mls.prepare(product_ids=scan_apple)
+        self.assertEqual(len(picking.move_line_ids), 2)
+        self.assertEqual(apple_ml.product_uom_qty, 4)
```
scott-redshaw, line 1696: Where is this value of 4 for the quantity of apples coming from?

PR #944, opened 2025-07-01
[TEST] udes_sale_stock: Add client_order_ref param to helper

kdwyer:
SE-2034

SaleOrder.client_order_ref is required, so make it required for the create_sale test helper method. Likewise requested_date.

### DIFF
--- a/addons/udes_sale_stock/tests/common.py
+++ b/addons/udes_sale_stock/tests/common.py
@@ -2,7 +2,6 @@ from odoo.addons.udes_stock.tests.common import BaseUDES
 
 
 class BaseSaleUDES(BaseUDES):
-
     """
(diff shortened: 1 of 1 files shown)


PeterAlabaster (approved):

addons/udes_sale_stock/tests/common.py
```diff
@@ -23,13 +22,15 @@ def create_sale_line(cls, sale, product, qty, **kwargs):
         return cls.env["sale.order.line"].create(create_values)
 
     @classmethod
-    def create_sale(cls, customer, **kwargs):
+    def create_sale(cls, customer, client_order_ref, requested_date, **kwargs):
```
armandcela, line 25: Question: Is this method called always with client_order_ref and requested date on other classes which inherits from BaseSaleUDES?

PR #945, opened 2025-07-01
SE-2021

ws-mps:
* Add new field to stock.picking.batch to hold multiple users
* Fix _create_split_move to get the picking_id from the move if one is provided, this can happen when picking_id is in kwargs
* Only log continuation batch comments if a continuation batch is necessary


addons/udes_stock_picking_batch/models/stock_picking_batch.py
```diff
@@ -24,6 +24,26 @@ class StockPickingBatch(models.Model):
             "in_progress": [("readonly", False)],
         },
     )
+    u_user_ids = fields.Many2many(
```
armandcela, line 27: Should we consider having only one field u_user_ids now, and removing Responsible User user_id field, or is a big change?

addons/udes_stock/models/stock_move.py
```diff
@@ -377,7 +377,8 @@ def _create_split_move(self, move_lines, remaining_qty, new_move_qty, **kwargs):
         new_move = self.copy(default_values)
 
         # Update the moved move lines with the new move, and no picking id
```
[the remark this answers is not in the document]
ws-mps: Yes, comment can be removed now. Picking id is not in the default values but it can be in **kwargs

addons/udes_stock_picking_batch/models/stock_picking_batch.py
ws-mps: Very good point, but yes it is a change too big, it will go its own story.

armandcela (approved):

PR #949, opened 2025-07-09
[FIX] udes_stock: Allow total qty scan for lots

DarpanBhatt-UP:
SE-2090

### DIFF
--- a/addons/udes_stock/models/stock_move_line.py
+++ b/addons/udes_stock/models/stock_move_line.py
@@ -435,13 +435,33 @@ class StockMoveLine(models.Model):
                         qty_done, product, package, lot_name, location, picking=self.picking_id
                     )
                     prod_dict = {"qty_done": qty_done, "lot_name": lot_name_val}
-                    if lot_expiry_dates:
-                        # TODO: hidden dep to product_expiry
-                        prod_dict["expiration_date"] = lot_expiry_dates[idx]
+                    remaining_qty = qty_done
                     if lot_measure_qty:
-                        vals.update({"u_measure_qty": lot_measure_qty, "u_measure_type": measure_type})
+                        vals.update({"u_measure_type": measure_type})
                     prod_dict.update(vals)
-                    res[prod_mls] = prod_dict
+                    for move_line in prod_mls:
+                        move_line_dict = prod_dict.copy()
+                        if not remaining_qty:
+                            break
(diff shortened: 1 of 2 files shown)


addons/udes_stock/models/stock_move_line.py
```diff
@@ -455,10 +460,10 @@ def prepare(
...
                         remaining_qty -= move_line.product_uom_qty
                         prod_dict["qty_done"] = move_line.product_uom_qty
-                if measure_quantity:
-                    vals.update({"u_measure_qty": measure_quantity, "u_measure_type": measure_type})
-                prod_dict.update(vals)
-                res[prod_mls] = prod_dict
+                    res[move_line] = prod_dict
+                    if measure_quantity:
+                        vals.update({"u_measure_qty": measure_quantity, "u_measure_type": measure_type})
+                    prod_dict.update(vals)
```
armandcela, line 466: I have a similar change on my story in commit https://github.com/unipartdigital/udes-open/pull/948/commits/68ea3137c84024244236f4bdc26ce910c403b4a6

The difference is that in my commit i am copying prod_dict as when used product_dict, depending on the change after first line when was setting res[move_line] = prod_dict for the second move line was changing even the first line. But they are very similar

ws-mps: Let's go with your solution @armandcela since it is safer.

addons/udes_stock/models/stock_move_line.py
```diff
@@ -434,13 +434,18 @@ def prepare(
...
-                    res[prod_mls] = prod_dict
+                    remaining_qty = qty_done
+                    for move_line in prod_mls:
+                        if move_line.product_uom_qty <= qty_done and remaining_qty:
+                            remaining_qty -= move_line.product_uom_qty
+                            prod_dict["qty_done"] = move_line.product_uom_qty
+                        if lot_expiry_dates:
+                            # TODO: hidden dep to product_expiry
+                            prod_dict["expiration_date"] = lot_expiry_dates[idx]
+                        if lot_measure_qty:
```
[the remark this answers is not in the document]
DarpanBhatt-UP: In next commit it will look like this under loop

                       ```
 if lot_expiry_dates:
                            # TODO: hidden dep to product_expiry
                            move_line_dict["expiration_date"] = lot_expiry_dates[idx]
```

PR #956, opened 2025-07-21
[ADD] udes_stock_refactoring: Recomputing u_grouping_key every

armandcela:
Recomputing u_grouping_key every time method group_by_key is called.

SE-2119

### DIFF
--- a/addons/udes_stock_refactoring/models/stock_move_line.py
+++ b/addons/udes_stock_refactoring/models/stock_move_line.py
@@ -39,11 +39,10 @@ class StockMoveLine(models.Model):
             raise UserError(
                 _("Cannot group move lines when their picking type has no grouping key set.")
             )
-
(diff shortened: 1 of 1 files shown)


addons/udes_stock_refactoring/models/stock_move_line.py
```diff
@@ -39,11 +39,10 @@ def group_by_key(self):
             raise UserError(
                 _("Cannot group move lines when their picking type has no grouping key set.")
             )
-
+        # Computing u_grouping_key field with context compute_key = True every time that we call group_by_key method.
```
PeterAlabaster, line 42: nit: We should probably explain why this is needed e.g
```python
# Recompute the grouping_key every time as there are some edge cases where it will not recompute on the fly
```
or something, I don't know if that accurately describes the issue we were having though

PeterAlabaster (approved):

PR #958, opened 2025-07-23
[ADD] udes_sale_stock: Create a new status for sale orders

DarpanBhatt-UP:
SE-1974

### DIFF
--- a/addons/udes_sale_stock/models/sale_order.py
+++ b/addons/udes_sale_stock/models/sale_order.py
@@ -67,6 +67,53 @@ class SaleOrder(models.Model):
         help="Special instructions associated with the order.",
     )
     u_done_date = fields.Datetime(string="Done Date", readonly=True)
+    u_delivery_state = fields.Selection(
+        [
+            ("confirmed", "Confirmed"),
+            ("no_stock", "No Stock"),
+            ("allocated", "Allocated"),
+            ("picked", "Picked"),
+            ("packed", "Packed"),
+            ("done", "Done"),
+            ("cancelled", "Cancelled"),
+        ],
+        string="Delivery Status",
+        compute="_compute_delivery_state",
+        store=True,
+        help="""
+        "Confirmed: Pickings are created and are in waiting state"
+        "No Stock: One or more order line has no stock"
+        "Allocated: All pick pickings are reserved"
+        "Picked: All pick pickings are in done state"
+        "Packed: All pack pickings are in done state"
+        "Cancelled: Sale order is in cancel state"
+        "Done: All pickings are in done state"
+        """,
+    )
+
+    @api.depends("order_line.u_delivery_line_state")
+    def _compute_delivery_state(self):
+        """
+        Compute the orders delivery status based on the order lines u_delivery_line_state
+        """
+        for order in self:
+            state = False
+            line_states = order.order_line.mapped("u_delivery_line_state")
+            if order.state == "cancel":
+                state = "cancelled"
+            elif any(state == "no_stock" for state in line_states):
+                state = "no_stock"
+            elif any(state == "confirmed" for state in line_states):
+                state = "confirmed"
+            elif any(state == "allocated" for state in line_states):
+                state = "allocated"
+            elif any(state == "picked" for state in line_states):
+                state = "picked"
+            elif any(state == "packed" for state in line_states):
+                state = "packed"
+            elif any(state == "done" for state in line_states):
+                state = "done"
+            order.u_delivery_state = state
 
     @api.depends("order_line.move_ids.picking_id")
     def _compute_picking_ids_by_line(self):
--- a/addons/udes_sale_stock/models/sale_order_line.py
+++ b/addons/udes_sale_stock/models/sale_order_line.py
@@ -23,6 +23,84 @@ class SaleOrderLine(models.Model):
         index=True,
     )
     u_cancelled_qty = fields.Float(string="Cancelled Quantity", digits="Product Unit of Measure")
+    u_delivery_line_state = fields.Selection(
+        [
+            ("confirmed", "Confirmed"),
+            ("no_stock", "No Stock"),
+            ("allocated", "Allocated"),
+            ("picked", "Picked"),
+            ("packed", "Packed"),
+            ("done", "Done"),
+            ("cancelled", "Cancelled"),
+        ],
+        string="Delivery Status",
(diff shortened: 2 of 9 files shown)


addons/udes_sale_stock/models/sale_order.py
```diff
@@ -67,6 +67,46 @@ class SaleOrder(models.Model):
...
+    def _find_delivery_status(self):
+        """
+        Added as a part of SE-1974
+        u_delivery_state will be calculated based on order line field
+        u_delivery_line_state.
+        """
+        for order in self:
+            state = False
+            if order.order_line:
+                if any(x.u_delivery_line_state == "no_stock" for x in order.order_line):
```
PeterAlabaster, line 95: suggestion: pull out states into variable once, and compare against that, so we do less calls through ORM

```python
        for order in self:
            state = False
            if order.order_line:
                line_states = order.order_line.mapped("u_delivery_line_state")
                if any(state == "no_stock" for state in line_states):
                    state = "no_stock"
                elif any(state == "cancelled" for state in line_states):
                    state = "cancelled"
                elif any(state == "confirmed" for state in line_states):
                    state = "confirmed"
                elif any(state == "allocated" for state in line_states):
                    state = "allocated"
                elif any(state == "picked" for state in line_states):
                    state = "picked"
                elif any(state == "packed" for state in line_states):
                    state = "packed"
                elif any(state == "done" for state in line_states):
                    state = "done"
            order.u_delivery_state = state
```

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -22,6 +22,78 @@ class SaleOrderLine(models.Model):
...
+            if line.route_id:
+                route_id = line.route_id
+                pick_op_type = route_id.u_pick_operation_type_ids
+                pack_op_type = route_id.u_pack_operation_type_ids
+                if not pick_op_type and not pack_op_type:
+                    line.u_delivery_line_state = state
+                    continue
+                if line.order_id.state == "cancel":
+                    state = "cancelled"
+                    line.u_delivery_line_state = state
```
PeterAlabaster, line 70: nit: if we are doing `continue` we can probably skip assignment to `state` variable and just assign "cancelled" directly here, and similar change for other branches which end with `continue`.

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -22,6 +22,78 @@ class SaleOrderLine(models.Model):
...
+        compute = "_find_delivery_line_status",
+        store = True,
+
+    )
+
+    @api.depends("move_ids", "move_ids.state", "product_id", "order_id.state")
+    def _find_delivery_line_status(self):
+        """
+        Added as a part of SE-1974
+        - Sale order line will always have route_id field set.
```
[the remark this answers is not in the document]
DarpanBhatt-UP: this is an assumption. We have added configuration in stock.location.route class. My understanding is we will always have route_id field set in sales order lines. If we have a case when we do not set route_id i will need to change computation.

addons/udes_sale_stock/models/sale_order.py
PeterAlabaster: with the above code as well you can probably remove the `if order.order_line` branch entirely and it will still work because `sale.order.line().mapped("u_delivery_line_state")` will just produce `[]` so all the checks will be negative

PeterAlabaster (approved):

JayshreeMahajan (approved):

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -77,7 +77,7 @@ def _compute_delivery_line_state(self):
                 if line.order_id.state == "cancel":
                     line.u_delivery_line_state = "cancelled"
                     continue
-                if not line.product_id.qty_available:
+                if not line.product_id.qty_available and not line.qty_delivered:
```
PeterAlabaster, line 80: q: what if a line is partially delivered? i.e qty of the line is 5 and 2 are delivered, 3 are waiting stock?

DarpanBhatt-UP: Then probably matching ordered and delivered qty is required. Something like line.product_uom_qty != line.qty_delivered.

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -77,7 +77,7 @@ def _compute_delivery_line_state(self):
...
                     line.u_delivery_line_state = "no_stock"
                     continue
                 all_pickings = line.move_ids.filtered(lambda x: x.state not in ["cancel"]).mapped("picking_id")
```
ws-mps, line 83: Why no in cancel? instead of !=?
No need for mapped.

DarpanBhatt-UP: Yep let me improve must have thought initially we might need to check multiple status so used not in with list. I wanted to get stock.picking obj so used mapped.

PeterAlabaster: we can remove the `.mapped("picking_id")` and replace with just `.picking_id`

PeterAlabaster: e.g
```python
stock.move(1,2,3).picking_id
>> stock.picking(1,2)
```

DarpanBhatt-UP: cool let me make this change

addons/udes_sale_stock/models/sale_order_line.py
```diff
@@ -77,7 +77,7 @@ def _compute_delivery_line_state(self):
                 if line.order_id.state == "cancel":
```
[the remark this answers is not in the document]
ws-mps: I disagree, in that case to set u_delivery_status to cancel to you only have to check if the state of the sale order is cancelled.
I think it makes more sense to have more granularity on the state of the lines, being cancel if their flag is True for is_cancelled, then the sale u_delivery_status will be cancelled if ALL sale lines are in state cancel, which is the same as the sale order normal status is cancel.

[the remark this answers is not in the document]
ws-mps: Wasn't clear that I disagree about not computing the cancelled state for each line, but I do agree that we can check it by checking the state of the sale order instead.

PR #959, opened 2025-07-24
Se 2122

armandcela:


PeterAlabaster (approved):

addons/udes_sale_stock/models/stock_move.py
```diff
@@ -27,15 +27,17 @@ def _action_cancel(self):
...
-                self.filtered(lambda m: m.location_dest_id == location_customers)
-                .mapped("sale_line_id")
-                .filtered(lambda s: len(s.move_ids.filtered(not_cancelled_filter)) == 0)
-            )
-            if lines_to_cancel:
-                lines_to_cancel.action_cancel()
+        delivery_cancelled_moves = self.filtered(lambda m: m.location_dest_id == location_customers)
 
+        if not self.env.context.get("disable_sale_cancel", False) and delivery_cancelled_moves:
+            lines_to_cancel = delivery_cancelled_moves.sale_line_id.filtered(lambda s: len(s.move_ids.filtered(not_cancelled_filter)) == 0)
```
kdwyer, line 33: This is the only place that we use `not_cancelled_filter`, so perhaps we should replace it with its inverse to simplify the code?  It's always harder to think about `not`s.

PR #960, opened 2025-07-30
[FIX] udes_stock_cron: Do not reserve stock in draft batches

kdwyer:
SE-2137

The add_processed_pickings method adds all the pickings in a batch to the processed collection, and this messes up the comparison in reserve_stock_for_picking_type.

Don't process if the current picking is in the processed collection.


kdwyer:
Why is this still open?

PR #961, opened 2025-08-08
Se 2144

kdwyer:
1. Add a filtering method to stock quant to allow us to arbitrarily filter the results of `_gather`

2. Implement unreserving of stock for split picks.


addons/udes_stock_routing/models/stock_move.py
```diff
@@ -289,3 +289,75 @@ def _update_reserved_quantity(
...
+        res = super()._do_unreserve()
+        applicable_rules = Rule.search(
+            [
+                ("u_run_on_assign", "=", True),
+                ("picking_type_id", "in", self.picking_type_id.ids),
+            ]
+        )
+
+        # TODO What if no origin?
+        for origin, moves in self.groupby("origin"):
```
ws-mps, line 317: This goupby should go into a method so it can be overriden.

kdwyer: Done (fixed in a force-push, sorry)

PR #962, opened 2025-08-08
Se 2144 14.25.27.7

kdwyer:


addons/udes_stock_routing/models/stock_move.py
```diff
@@ -289,3 +289,77 @@ def _update_reserved_quantity(
...
+            [
+                ("u_run_on_assign", "=", True),
+                ("picking_type_id", "in", self.picking_type_id.ids),
+            ]
+        )
+        if not applicable_rules:
+            return res
+
+        # TODO What if no origin?
+        for origin, moves in self.groupby("origin"):
```
ws-mps, line 319: Same as other PR, this should be a method that can be overridden.

PR #963, opened 2025-08-12
Se 2144 14.0

kdwyer:


addons/udes_stock_routing/models/stock_move.py
```diff
@@ -289,3 +293,97 @@ def _update_reserved_quantity(
             owner_id=owner_id,
             strict=strict,
         )
+
+    def group_moves_to_unreserve(self):
```
armandcela, line 297: Can we pass origin as a parameter if we want to use in future to group by another field, if is not the case maybe changing the method name smth like group_by_origin_moves_to_unreserve

armandcela (approved):

addons/udes_stock_routing/models/stock_move.py
kdwyer: The idea is that this method is overridden as required; the grouping might be by a chain of relations, which `groupby` doesn't support directly

adamrpatrick (approved):

PR #969, opened 2025-08-29
SE-2170

ws-mps:


addons/udes_stock/views/stock_orderpoint_views.xml
```diff
@@ -21,13 +21,13 @@
 
     <!-- Hide Replenishment Menu -->
     <record id="stock.menu_reordering_rules_replenish" model="ir.ui.menu">
-        <field name="active">0</field>
+        <field name="active">1</field>
```
adamrpatrick, line 24: We don't use this menu anywhere else, is this the right "Replenishment" menu?

adamrpatrick (approved):

addons/udes_stock/wizard/udes_proactive_replens.py
```diff
@@ -0,0 +1,88 @@
...
+            extra = orderpoint.qty_forecast - orderpoint.product_min_qty
+            percent_above_min = extra / top * 100
+            if percent_above_min <= self.replen_percentage:
+                op_todo[orderpoint] = percent_above_min
+
+        counter = 0
+        for orderpoint, percent_above_min in sorted(op_todo.items(), key=lambda x: x[1], reverse=False):
+            if counter >= self.number_replen_moves:
+                break
+            quants = Quant.search(
```
kdwyer, line 55: Is it worth hoisting this query out of the loop, fetching quants for all the products at once and filtering inside the loop?

ws-mps: We could potentially have lots of order points, so I was trying to avoid to get all the quants for 1000 products, and only get the 20 or 30 we really need.

PR #970, opened 2025-09-03
[FIX] stock_picking_batch: Add domain on user_id to avoid boolean err…

JayshreeMahajan:
…or in search

SE-1488

### DIFF
--- a/addons/udes_stock_picking_batch/views/stock_picking_batch.xml
+++ b/addons/udes_stock_picking_batch/views/stock_picking_batch.xml
@@ -20,6 +20,11 @@
         <xpath expr="//button[@name='action_cancel']" position="attributes">
           <attribute name="states">ready,waiting,in_progress</attribute>
         </xpath>
+        <xpath expr="//field[@name='user_id']" position="replace">
+          <field name="user_id"
+            domain="[('active', '=', True), ('company_ids', 'in', [company_id])]"
+          />
+        </xpath>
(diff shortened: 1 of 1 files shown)


addons/udes_stock_picking_batch/views/stock_picking_batch.xml
```diff
@@ -20,6 +20,11 @@
         <xpath expr="//button[@name='action_cancel']" position="attributes">
           <attribute name="states">ready,waiting,in_progress</attribute>
         </xpath>
+        <xpath expr="//field[@name='user_id']" position="replace">
+          <field name="user_id"
+            domain="[('active', '=', True), ('company_ids', 'in', [company_id])]"
+          />
+        </xpath>
```
armandcela, line 27: Try avoiding replace where possible
```suggestion
        <xpath expr="//field[@name='user_id']" position="attributes">
          <attribute name="domain">[('active', '=', True), ('company_ids', 'in', [company_id])]</attribute>
        </xpath>
```

JayshreeMahajan: I’ll remember that.
Since we are adding restrictions in the wrapper method, I think these changes are not needed.

JayshreeMahajan: This error can occur in other places too, since other functions may pass a bool. Adding this check will fix all of them.

PR #971, opened 2025-09-05
Unreservation action fixes

kdwyer:
- Fix accidental overwriting of context when cancelling moves
- Do nothing with cancelled moves
- Do nothing with moves already in the original picking type
- instead of merging moves, call `_action_confirm` and delegate unreservation actions to the instance config.

### DIFF
--- a/addons/udes_stock_routing/tests/test_split_pick.py
+++ b/addons/udes_stock_routing/tests/test_split_pick.py
@@ -551,15 +551,24 @@ class SplitPickUnreservationTestCase(SplitPickBase):
     # These tests are based on those in TestSplitPick. However here we assert
     # the state after pickings have been unreserved.
 
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
+        # Create 200 apples on pallets in bulk and 100 loose in stock. With
+        # whole pallet reservation enabled, lines of 100 should result in bulk
+        # picks, and non-100 amounts "pick" picks, unless additional rules are
+        # applied.
+        cls.create_quant(
+            cls.apple.id, cls.bulk_location_01.id, 100, package_id=cls.create_package().id
+        )
+        cls.create_quant(
+            cls.apple.id, cls.bulk_location_01.id, 100, package_id=cls.create_package().id
+        )
+        cls.create_quant(cls.apple.id, cls.standard_location_01.id, 100)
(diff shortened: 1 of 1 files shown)


addons/udes_stock_routing/models/stock_move.py
```diff
@@ -337,55 +336,40 @@ def _do_unreserve(self):
                     continue
                 split_pick_type = rule.picking_type_id
                 original_picking_type = rule.u_run_on_assign_applicable_to
+                if split_pick_type == original_picking_type:
+                    _logger.debug(
+                        "Skipping because split and original picking types are the same: %r",
+                        split_pick_type | original_picking_type,
```
ws-mps, line 342: Do we need the | here if they are the same?

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -337,55 +336,40 @@ def _do_unreserve(self):
...
+                    )
+                    continue
                 split_pick_moves = moves.filtered(lambda p: p.picking_type_id == split_pick_type)
```
ws-mps, line 345: I'd move this line before the if and consider doing: split_pick_moves._action_confirm(merge=True) inside the if

ws-mps: Just thinking that the `split_pick_moves._action_confirm(merge=True)` would be better to put in a method so the behaviour can be overridden if necessary.

addons/udes_stock_routing/models/stock_move.py
```diff
@@ -337,55 +336,40 @@ def _do_unreserve(self):
...
-                        {
-                            "picking_type_id": original_picking_type.id,
-                            "location_id": original_picking_type.default_location_src_id,
-                        }
-                    )
-                    _logger.debug("Created new picking %r for split pick moves", target)
-                # Remove references to moves that have been merged away.
-                moves = moves.exists()
+                )
+                target.move_lines._action_confirm(merge=True)
```
ws-mps, line 373: Same as my suggestion above, put this in a method please.

addons/udes_stock_routing/models/stock_move.py
kdwyer: No fixed in ab1a7ef3c

PR #975, opened 2025-09-15
[IMP] udes_stock: Auto generating Reorder Rules

JayshreeMahajan:
SE-2145

### DIFF
--- a/addons/udes_stock/models/stock_warehouse_orderpoint.py
+++ b/addons/udes_stock/models/stock_warehouse_orderpoint.py
@@ -148,8 +148,8 @@ class Orderpoint(models.Model):
                 ("state", "not in", ["cancel", "done"]),
             ])
             if moves:
-                if all(m.state == "draft" for m in moves):
-                    moves.unlink()
+                if all(m.state in ["draft","waiting","confirmed"] for m in moves):
+                    moves._action_cancel()
                     old_rule.unlink()
                 else:
                     raise UserError(_(


addons/udes_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -67,3 +67,63 @@ def unlink(self):
         Product._compute_nbr_reordering_rules()
         return res
 
+    @api.model
+    def _create_or_update_replenishment_rules(self, product, location, qty):
```
kdwyer, line 71: Nit: This should be a public method (no leading underscore) as it is called from a different class.

addons/udes_stock/models/stock_picking.py
```diff
@@ -1046,6 +1050,17 @@ def validate_picking(self, create_backorder=False, force_validate=False):
             stats.count / stats.elapsed,
         )
 
+        if Warehouse.u_enabled_auto_replenishment and self.picking_type_id.id == picking_type_putaway.id:
```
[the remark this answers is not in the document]
JayshreeMahajan: addressed

addons/udes_stock/models/stock_picking.py
```diff
@@ -1046,7 +1047,57 @@ def validate_picking(self, create_backorder=False, force_validate=False):
...
+        if self.env.user.get_user_warehouse().u_enabled_auto_replenishment:
+           self._process_auto_replenishment()
+
         return new_picking
+    
+    def _get_picking_type_for_replenishment(self):
+        """
+        Determine the picking type to use for replenishment rules.
+
+        - Defined in **UDES Open**: defaults to *Internal Transfer* picking type.
```
PeterAlabaster, line 1059: These 2 lines can be removed

addons/udes_stock/models/stock_warehouse_orderpoint.py
```diff
@@ -67,3 +67,63 @@ def unlink(self):
...
+              - Not started → delete them.
+              - In progress → raise error.
+        """
+        StockMove = self.env["stock.move"]
+
+        # Check for existing rule on this location & product
+        existing_rule = self.search([
+            ("location_id", "=", location.id),
+            ("product_id", "=", product.id)
+        ], limit=1)
```
PeterAlabaster, line 90: Is it possible for users to manually create rules, so that there are in fact 2 which exist for the same product and location, and if so - did we want to 'tidy up' those here?

addons/udes_stock/models/stock_picking.py
```diff
@@ -1046,7 +1047,57 @@ def validate_picking(self, create_backorder=False, force_validate=False):
...
         )
 
+        if self.env.user.get_user_warehouse().u_enabled_auto_replenishment:
+           self._process_auto_replenishment()
+
         return new_picking
+    
+    def _get_picking_type_for_replenishment(self):
+        """
+        Determine the picking type to use for replenishment rules.
```
ws-mps, line 1057: Rephrase please: Determine the picking type to automatically create replenishment rules.

addons/udes_stock/models/stock_warehouse_orderpoint.py
JayshreeMahajan: not sure about this will confirm with POs.

PR #979, opened 2025-09-26
[IMP] udes_stock_refactoring: Improve refactoring logs

dandihar:
On refactor, logs now display the pickings, moves and move_lines SE-1483

### DIFF
--- a/addons/udes_stock_refactoring/models/stock_move.py
+++ b/addons/udes_stock_refactoring/models/stock_move.py
@@ -92,6 +92,7 @@ class StockMove(models.Model):
         if stage is not None and stage not in STOCK_REFACTOR_STAGES.values():
             raise UserError(_("Unknown stage for move refactor: %s") % (stage))
         moves = self.exists()
+        original_move_ids = moves.ids
 
         if self.env.context.get("disable_move_refactor"):
             return moves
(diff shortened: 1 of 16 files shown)


addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -143,6 +141,27 @@ def _action_refactor(self, stage=None, refactor_action=False):
...
+                    picking_refs = moves.mapped("picking_id.name")              
+                    
+                    source = self.env.context.get("refactor_source")
+                    if not source:
+                        stack = inspect.stack()
+                        caller = next((frame.function for frame in stack[1:6]), "unknown")
+                        source = f"{caller}"
+                    user = self.env.user.name
+
+                    _logger.info("Refactoring moves= [%s], move_lines= [%s] pickings= [%s] by %s, stage=%s, user=%s",
```
ws-mps, line 156: We are not logging the picking type, which is also useful.

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -143,6 +141,27 @@ def _action_refactor(self, stage=None, refactor_action=False):
                     stage_moves = stage_moves.with_context(remove_related_moves=True)
                     new_moves = refactor_class.do_refactor(stage_moves)
 
+                    moves = (stage_moves | new_moves)
```
ws-mps, line 144: It will be better to log them separated: like original_moves = XXX, new_moves = XXX

addons/udes_stock_refactoring/models/stock_move.py
```diff
@@ -143,6 +141,27 @@ def _action_refactor(self, stage=None, refactor_action=False):
...
+                    move_refs = moves.mapped("display_name")
+                    move_lines_refs = moves.mapped("move_line_ids.display_name")
```
ws-mps, line 146: Same as above.

kdwyer:
Can we have a unit test (or more) to capture the behaviour? You can use [assertLogs]([assertL](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs)).

PR #1015, opened 2026-05-26
SE-2619

ws-mps:


addons/udes_stock/models/stock_picking.py
```diff
@@ -954,6 +953,14 @@ def add_unexpected_parts(self, product_quantities, merge_on_confirm):
             new_moves._merge_moves()
         return new_mls
 
+    @api.depends('move_type', 'immediate_transfer', 'move_lines.state', 'move_lines.picking_id')
+    def _compute_state(self):
+        """Set date_done to cancelled pickings if not is not set."""
```
scott-redshaw: Reword docstring to something like this "Set date_done to cancelled pickings if not, it is not set" something like that instead?

scott-redshaw (commented):
This maybe a stupid question but how does these changes stop the EDI requests from merging? (as mentioned on the story)

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1107,6 +1107,32 @@ def test_create_multiple_pickings(self):
...
+        all_pickings = pick
+        next_pickings = pick.u_next_picking_ids
+        while next_pickings:
+            all_pickings |= next_pickings
+            next_pickings = next_pickings.u_next_picking_ids
+        all_pickings.picking_type_id.write({"u_propagate_cancel": True})
+        self.assertEqual(pick.state, "confirmed")
+        self.assertFalse(pick.date_done)
+        for ppp in (all_pickings - pick):
+            self.assertEqual(ppp.state, "waiting")
```
kdwyer:
```python
for ppp in all_pickings:
    with self.subTest(picking=ppp):
       self.assertEqual(ppp.state, "waiting")
```

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1107,6 +1107,32 @@ def test_create_multiple_pickings(self):
...
+        self.assertTrue(all(p.date_done == False for p in all_pickings))
+        pick.action_cancel()
+        for ppp in all_pickings:
```
kdwyer, line 1137: Another `subTest` opportunity

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1107,6 +1107,32 @@ def test_create_multiple_pickings(self):
...
+        next_pickings = pick.u_next_picking_ids
+        while next_pickings:
+            all_pickings |= next_pickings
+            next_pickings = next_pickings.u_next_picking_ids
+        all_pickings.picking_type_id.write({"u_propagate_cancel": True})
+        self.assertEqual(pick.state, "confirmed")
+        self.assertFalse(pick.date_done)
+        for ppp in (all_pickings - pick):
+            self.assertEqual(ppp.state, "waiting")
+        self.assertTrue(all(p.date_done == False for p in all_pickings))
```
kdwyer: I'm not very keen on these `assertTrue(all(...))` constructions, they aren't very informative if the test fails. Perhaps something like this instead?

```python
self.assertCountEqual(ppp.mapped('date_done'), [False] * len(ppp))
```

addons/udes_stock/tests/test_stock_picking.py
```diff
@@ -1107,6 +1107,32 @@ def test_create_multiple_pickings(self):
         # Check batch created
         self.assertTrue(picks.batch_id)
 
+    def test_cancel_picking_done_date(self):
```
kdwyer: It's make the method name more descriptive, perhaps
`test_cancel_picking_sets_done_date_on_entire_chain`?

And s/chaing/chain/ in the docstring

addons/udes_stock/models/stock_picking.py
```diff
@@ -954,6 +953,14 @@ def add_unexpected_parts(self, product_quantities, merge_on_confirm):
...
+        super()._compute_state()
+        for picking in self:
+            if picking.state == 'cancel' and not picking.date_done:
+                picking.date_done = fields.Datetime.now()
```
kdwyer: Do we want to hoist `now` outside of the loop to guarantee that all the pickings have exactly the same done date?

```python
        now = fields.Datetime.now()
        for picking in self:
            if picking.state == 'cancel' and not picking.date_done:
                picking.date_done = now
```

addons/udes_stock/tests/test_stock_picking.py
ws-mps: Moved inside the loop within a subtest.

scott-redshaw (approved):

kdwyer (approved):

PR #1016, opened 2026-05-27
[IMP] udes_security: Allow all default user groups to be added

scott-redshaw:
This is to allow all base.default_user groups irrelevant of  group.u_required_group_id_to_change when a user is being created for the first time

SE-2589

### DIFF
--- a/addons/udes_security/models/res_users.py
+++ b/addons/udes_security/models/res_users.py
@@ -90,8 +90,8 @@ class ResUsers(models.Model):
                     group = Group.browse(group_id)
 
                     if group.u_required_group_id_to_change:
-                        # Allow all groups to be added that are assigned to the default users groups
-                        if new_user and operation == 4:
+                        # Allow all groups to be added that are assigned to the default user
+                        if new_user and operation == RelFieldOps.Add:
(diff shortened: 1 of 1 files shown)


addons/udes_security/models/res_users.py
```diff
@@ -88,6 +90,11 @@ def _check_user_group_modify(self, vals, new_user=False):
                     group = Group.browse(group_id)
 
                     if group.u_required_group_id_to_change:
+                        # Allow all groups to be added that are assigned to the default users groups
+                        if new_user and operation == 4:
```
kdwyer: We've defined an enum to give these magic numbers symbolic names: https://github.com/unipartdigital/udes-open/blob/a3d09d4d8187d8fa60d77bbddcad482363e6c24c/addons/udes_common/tools/__init__.py#L6 (follow the import to see the enum). So we could have this, so the reader doesn't have to remember what "4" is supposed to mean:

```suggestion
                        if new_user and operation == RelFieldOps.Add:
```

ws-mps (approved):

PeterAlabaster (approved):

PR #1020, opened 2026-06-04
[FIX] udes_stock: Inventory at date bug

DarpanBhatt-UP:
SE-2760

### DIFF
--- a/addons/udes_stock/wizard/stock_quant_history.py
+++ b/addons/udes_stock/wizard/stock_quant_history.py
@@ -1,0 +1,11 @@
+from odoo import models, fields
+
+
+class StockQuantityHistory(models.TransientModel):
+    _inherit = "stock.quantity.history"
+
+    def open_at_date(self):
+        res = super().open_at_date()
+        if (ctx := res.get("context")) and "to_date" in ctx:
+            ctx.update({"to_date": fields.Datetime.to_string(ctx.get("to_date"))})
+        return res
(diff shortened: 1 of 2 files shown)


addons/udes_stock/wizard/stock_quant_history.py
```diff
@@ -0,0 +1,11 @@
+from odoo import models, fields
+
+
+class StockQuantityHistory(models.TransientModel):
+    _inherit = "stock.quantity.history"
+
+    def open_at_date(self):
+        res = super().open_at_date()
+        if (ctx := res.get("context")) and "to_date" in ctx:
```
PeterAlabaster, line 9: I think it would be useful to describe why we have added this

addons/udes_stock/wizard/stock_quant_history.py
```diff
@@ -0,0 +1,11 @@
...
+            ctx.update({"to_date": fields.Datetime.to_string(ctx.get("to_date"))})
```
armandcela, line 10:
```suggestion
        if  to_date := res.get("context", {}).get("to_date"):
            res["context"]["to_date"] = fields.Datetime.to_string(to_date)
```
Smth like this maybe,

DarpanBhatt-UP: thought about `to_date := res.get("context", {}).get("to_date")`

then show another super call with inventory valuation report in which to_date is not in context hence did like this.

armandcela: It wont break if is not in the context i think, anyway in the CR chat i have added another comment, i think that we might close this MR

armandcela (changes requested):

PR #1023, opened 2026-07-08
[IMP] udes_common: Context manager for new cursor environments

kdwyer:
branch-description

Add a new context manager to do the book-keeping required to prevent Odoo from writing updates from the original env to a copy using a different cursor.

See also https://github.com/unipartdigital/odoo-edi/commit/7bb82cce86a3855bddebb7f3f6d840d1f089b90b

### DIFF
--- a/addons/udes_common/tools/contextmanagers.py
+++ b/addons/udes_common/tools/contextmanagers.py
@@ -1,0 +1,40 @@
+"""Context managers for UDES."""
+
+import contextlib
+
+
+@contextlib.contextmanager
+def temp_env(env, cursor_factory):
+    """
+    Preserve pending ORM writes and restore afterwards.
+
+    This is a helper for an Odoo bug where it commits all pending writes even
+    when the cursor is not the original env cursor.
+
+    Example usage:
+
+    ```
+    from odoo import http
+    with temp_env(
+        http.request.env, http.request.registry
+    ) as tmp_env, tmp_env.cr as temp_cr:
+        ...
+    ```
+
+    Params:
+        env: Environment - the original env that we are "forking"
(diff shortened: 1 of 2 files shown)


addons/udes_common/tools/contextmanagers.py
```diff
@@ -0,0 +1,30 @@
...
+import contextlib
+
+
+@contextlib.contextmanager
+def temp_env(env, cursor_factory):
+    """
+    Preserve pending ORM writes and restore afterwards.
+
+    This is a helper for an Odoo bug where it commits all pending writes even
+    the cursor is not the original env cursor.
```
PeterAlabaster: nit; 

```suggestion
    when the cursor is not the original env cursor.
```

addons/udes_common/tools/contextmanagers.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Params:
```
PeterAlabaster, line 24: It might be worth giving a simple implementation example here
```python
from odoo import http
with temp_env(
    http.request.env, http.request.registry
) as tmp_env, tmp_env.cr as temp_cr:
    ...
```

addons/udes_common/tools/contextmanagers.py
kdwyer: Addressed by 7ac8924b, please review.

addons/udes_common/tools/contextmanagers.py
kdwyer: Addressed by 7ac8924b, please review.

PeterAlabaster (approved):

